PR #47, opened 2024-03-08
feat(new-parachain): add network config

### LINKED ISSUE #39: `pop new parachain`: add zombienet config
Add a default zombienet config file when generating a new parachain from a template.

evilrobot-01:
Adds simple network config file, so that parachain can easily be launched with the following after build:

```shell
```

closes #39

### DIFF
--- a/src/engines/generator.rs
+++ b/src/engines/generator.rs
@@ -36,6 +36,12 @@ pub(crate) struct PalletTests {
 	pub(crate) module: String,
 }
 
+#[derive(Template)]
+#[template(path = "base/network.templ", escape = "none")]
+pub(crate) struct Network {
+	pub(crate) node: String,
+}
+
 // todo : generate directory structure
 // todo : This is only for development
 #[allow(unused)]
--- a/src/engines/parachain_engine.rs
+++ b/src/engines/parachain_engine.rs
@@ -1,6 +1,6 @@
 use crate::{
 	commands::new::parachain::Template,
-	engines::generator::ChainSpec,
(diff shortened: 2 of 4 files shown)


Daanvdplas:
I had to checkout to the remote branch like so: `git checkout "remotes/origin/frank/feat(new-parachain)-add-network-config"` (using quotes), I think because of the parentheses.

AlexD10S (approved):

AlexD10S:
> FYI: I had to checkout to the remote branch like so: `git checkout "remotes/origin/frank/feat(new-parachain)-add-network-config"` (using quotes), I think because of the parentheses.

Took a while to figure out how to move to the branch too

Daanvdplas:
Only comment I have is that we might want to put it in a folder. Some ideas:
- `networks/basic.toml`
- `integration-tests/networks/basic.toml`
- `test-networks/basic.toml`
- `zombienet/network.toml`

evilrobot-01:
A folder with one file seems redundant to me. Fine if there are more files in the future. It's the same as creating a folder with a single mod.rs file, when a name.rs works perfectly fine.

Also prefer using plain language to keep things obvious for end user. We abstract zombienet with `pop up` and therefore using a simple name like 'network' for a network definition file seems the most obvious.

evilrobot-01:
> FYI: I had to checkout to the remote branch like so: `git checkout "remotes/origin/frank/feat(new-parachain)-add-network-config"` (using quotes), I think because of the parentheses.

Thanks for letting me know, will omit in future.

PR #62, opened 2024-03-13
Al3mart/fix create on add

al3mart:
This PR adds the following fix:

When using `pop add pallet template -r <runtime/path>` the structure `pallets/<name-of-pallet>` will be created within the workspace of the runtime, if the runtime is specified, if it doesn't exist. Meaning it will also create the right folders even if the runtime is run from other directory not part of the targeted workspace.

Before this change the command will add the pallet and inject in the runtime a local dependency that didn't exist in the project tree.

Unfortunately, the current state of `pop add pallet` introduces some breaking modifications on the Cargo files of the projects, so this can only be run once successfully.  Afterwards `cargo locate-project --workspace` will return errors that are not handled right now, leading to pallet folders being created in the directory from where the command was run.

### DIFF
--- a/src/helpers.rs
+++ b/src/helpers.rs
@@ -59,7 +59,7 @@ pub(crate) fn resolve_pallet_path(path: Option<String>) -> PathBuf {
 	}
 	// Check if inside a project
 	let cwd = current_dir().expect("current dir is inaccessible");
-	match workspace_dir() {
+	match workspace_dir(None) {
 		// No workspace available, return cwd.
 		None => cwd,
 		Some(dir) => {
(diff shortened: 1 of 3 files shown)


weezy20:
I found one issue while testing this out. I am in `pop-cli` repo which is a cargo workspace, and I run `cargo r -- add pallet`, this will make the tool think that I'm in a cargo workspace already and create `pallets` right there, on the pop-cli repo, rather than the root of the parachain repo. The fix would be to use relative paths to the provided argument `-r <runtime-path>`, going up 1 level from the runtime crate, and then adding a check to see if we are in a cargo workspace. We can additionally add checks to check if this is the global manifest by some sort of check that can tell us that `node` and `runtime` are members of this workspace.

src/helpers.rs
```diff
@@ -54,29 +54,57 @@ pub(crate) fn clone_and_degit(url: &str, target: &Path) -> Result<()> {
...
 	if let Some(path) = path {
 		return Path::new(&path).to_path_buf();
 	}
-	// Check if inside a template
+	// Check if inside a project
 	let cwd = current_dir().expect("current dir is inaccessible");
+	match workspace_dir(None) {
+		// No workspace available, return cwd.
+		None => cwd,
+		Some(dir) => {
```
weezy20, line 65: Suggest adding checks here to ensure that the cargo workspace belongs to a parachain project.

weezy20 (changes requested):
I found one issue while testing this out. I am in `pop-cli` repo which is a cargo workspace, and I run `cargo r -- add pallet`, this will make the tool think that I'm in a cargo workspace already and create `pallets` right there, on the pop-cli repo, rather than the root of the parachain repo. The fix would be to use relative paths to the provided argument `-r <runtime-path>`, going up 1 level from the runtime crate, and then adding a check to see if we are in a cargo workspace. We can additionally add checks to check if this is the global manifest by some sort of check that can tell us that `node` and `runtime` are members of this workspace.

src/helpers.rs
al3mart: This sounds like the logical next step :+1: . As the description says it just naively looks for the workspace root and creates any folder there.

al3mart:
> Looks like there are some cases where this panics or hangs. What is the command that I need to run to test this feature?

Right, @brunopgalvao. This PR only takes care of creating the folder structure `pallets/<my-addeed-pallet>` on the root of the workspace in case executing the command in one.

The changes don't address how `pop` identifies the runtime. At the time I pushed this the correct way of passing the location of the runtime to the tool was something like `pop add pallet template -r runtime/src/lib.rs`

al3mart:
I believe this one can be closed.

Issue #61, opened 2024-03-12
`pop new parachain` should initialize a git repo


weezy20:
`clone_and_degit` is used to instantiate a new parachain. But having a first commit or even `git init` run is desirable.

PR #65, opened 2024-03-14
feat: init git repo

CinematicCow:
This PR solves for #61. It initializes a new git repo with `initialized parachain` as the commit message. It takes the default global user config from git and uses it as the signature. 

Currently the test passes even if there is no `.git` dir in the new parachain. I need help with that.

### DIFF
--- a/src/helpers.rs
+++ b/src/helpers.rs
@@ -1,6 +1,6 @@
 use anyhow::Result;
 use cliclack::{log, outro_cancel};
-use git2::Repository;
+use git2::{IndexAddOption, Repository, ResetType};
 use std::{
 	env::current_dir,
 	fs::{self, OpenOptions},
@@ -47,6 +47,24 @@ pub(crate) fn clone_and_degit(url: &str, target: &Path) -> Result<()> {
 	let repo = Repository::clone(url, target)?;
 	let git_dir = repo.path();
 	fs::remove_dir_all(&git_dir)?;
+	Ok(())
+}
+
+/// Init a new git repo on creation of a parachain
+pub(crate) fn git_init(target: &Path, message: &str) -> Result<(), git2::Error> {
+	let repo = Repository::init(target)?;
+	let signature = repo.signature()?;
+
+	let mut index = repo.index()?;
+	index.add_all(["*"].iter(), IndexAddOption::DEFAULT, None)?;
+	let tree_id = index.write_tree()?;
+
+	let tree = repo.find_tree(tree_id)?;
+	let commit_id = repo.commit(Some("HEAD"), &signature, &signature, message, &tree, &[])?;
+
+	let commit_object = repo.find_object(commit_id, Some(git2::ObjectType::Commit))?;
+	repo.reset(&commit_object, ResetType::Hard, None)?;
+
 	Ok(())
 }
 
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -8,6 +8,7 @@ on:
 
 env:
   CARGO_TERM_COLOR: always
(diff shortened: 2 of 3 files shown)


src/helpers.rs
```diff
@@ -50,6 +50,24 @@ pub(crate) fn clone_and_degit(url: &str, target: &Path) -> Result<()> {
...
+	let signature = repo.signature()?;
+
+	let mut index = repo.index()?;
+	index.add_all(["*"].iter(), IndexAddOption::DEFAULT, None)?;
+	let tree_id = index.write_tree()?;
+
+	let tree = repo.find_tree(tree_id)?;
+	let commit_id = repo.commit(Some("HEAD"), &signature, &signature, message, &tree, &[])?;
+
+	let commit_object = repo.find_object(commit_id, Some(git2::ObjectType::Commit))?;
```
weezy20, line 65: Can you briefly elaborate on what this piece of code achieves?

weezy20 (approved):

src/helpers.rs
CinematicCow: - gets the index file of the repo and add all the files to it (staging area)
- creates a new git tree and write the index to it
- gets the tree from the `tree_id`
- creates a new commit, where, `signature` is the default author, commit message, the tree that was created and an empty vec of parent commit ids.
- resets the repo's working dir and index to the latest commit, to ensure repo's files match the state captured by the initial commit

weezy20: In the case of no git config for `name` and `email` does this approach work?

CinematicCow: if I'm not mistaken, git takes in the default user and system name if not specified explicitly. So yes, it should work, but I have to try and let you know.

weezy20:
@AlexD10S can you review?

weezy20:
@CinematicCow please push your test.

Daanvdplas:
Hello @CinematicCow! Amazing that you are contributing to our project.

>Currently the test passes even if there is no .git dir in the new parachain. I need help with that.

That is because it is initialised at [line 34 ](https://github.com/r0gue-io/pop-cli/blob/7dc2e5ddf7765f3e44730d438e6cb9931864db42/src/engines/parachain_engine.rs#L34 )and [68](https://github.com/r0gue-io/pop-cli/blob/7dc2e5ddf7765f3e44730d438e6cb9931864db42/src/engines/parachain_engine.rs#L68).

This is repetitive code and we should use your helper function. If you could remove the ones mentioned above and expand your test by including the situation of no git config and other test cases it would be amazing!

AlexD10S:
Thanks for contributing into the project. As Daan said it looks like is initialize in another function in the `create new parachain ` process.

Regarding to @weezy20 question, if no name or email the following line will throw an error:
```rs
let signature = repo.signature()?;
```

We want to display an error message to the user  in that case?

The rest of the code looks good to me, my only question is why we need the `commit`?
```rs
let commit_id = repo.commit(Some("HEAD"), &signature, &signature, message, &tree, &[])?;
```

weezy20:
@AlexD10S The commit is required so that the parachain that's created by pop new is ready for manipulation by pop add

CinematicCow:
- Fixed the test. The test now asserts on the number of commits.
- Signature missing error is now handled and displayed to the user.

Any ideas on how to handle the test on CI?

AlexD10S (approved):

weezy20 (approved):

AlexD10S (approved):

weezy20 (approved):

weezy20:
Thank you for your contributions @CinematicCow. This is much needed to make `add-pallet` work seamlessly. Enjoy using `pop`!

PR #73, opened 2024-03-20
chore: add rust-toolchain.toml

Daanvdplas:


rust-toolchain.toml
```diff
@@ -0,0 +1,14 @@
+[toolchain]
+channel = "stable"
+components = [
+	"cargo",
+	"clippy",
+	"rust-analyzer",
+	"rust-src",
+	"rust-std",
+	"rustc",
+	"rustc-dev",
```
weezy20, line 10: why is `rustc-dev` required?

Daanvdplas: Got it from here: https://github.com/paritytech/polkadot-sdk/blob/master/templates/solochain/env-setup/rust-toolchain.toml

Removed it because couldn't find a reason as well

weezy20 (approved):
LGTM!

PR #84, opened 2024-03-25
test: add unit test for `pop new pallet`

brunopgalvao:
This PR adds a unit test for `create_pallet_template`.

### DIFF
--- a/src/engines/pallet_engine.rs
+++ b/src/engines/pallet_engine.rs
@@ -96,4 +96,22 @@ mod tests {
 		let lib_rs_content = fs::read_to_string(pallet_path.join("src").join("lib.rs")).expect("Failed to read lib.rs");
         assert!(lib_rs_content.contains("pub mod pallet"), "lib.rs should contain pub mod pallet");
     }
+
+	#[test]
+	fn test_pallet_create_template_invalid_path() {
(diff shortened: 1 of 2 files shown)


AlexD10S (commented):
The test is good, but I feel we only testing the "happy path", can we test with wrong path, weird names or other alternatives.

Also some functions used by this method in helpers like `resolve_pallet_path` might need specific unit test to it.

brunopgalvao:
> The test is good, but I feel we only testing the "happy path", can we have a test with wrong path or other alternatives that make it fail.
> 
> Also some functions used by this method in helpers like `resolve_pallet_path` might need specific unit test to it.

Okay. Added some additional tests. Please approve :)

al3mart (approved):
The function tested uses `resolve_pallet_path` function that is feature-gated with the feature `parachains`.

This one is set as default, but I just noticed that this is the only helper function behind a feature, maybe it is worth removing that ?

We are also including helpers  in main if `contracts` or `parachains` (both default features) are active. 

Aside from that nit, I guess there could be more cases covered by tests like creating a pallet within a workspace, also from different folders in it, or creating a pallet in a different path than current working directory.

Approving, though would be cool getting more cases covered.

PR #86, opened 2024-03-26
test: pop up parachain

AlexD10S:
- Unit tests for `pop up parachain`. 
A couple of tests are under `unit_parachain` feature because takes a lot of time (Ideally slow process should be mocked for unit tests). To run the ignored tests: `cargo test  --features unit_parachain` 
To run all the tests:` cargo test --all-features`.

<del>- Integration test.
To have an integration test of this functionality is a bit tricky, as the command doesn't stop after the execution. 
I took as a reference a test in cumulus that let the command run for a X amount of seconds and then stop it. ([Example here](https://github.com/paritytech/polkadot-sdk/blob/master/cumulus/polkadot-parachain/tests/common.rs#L58)).</del>


evilrobot-01:
https://stackoverflow.com/questions/67859926/how-to-run-cargo-fmt-on-save-in-vscode

src/commands/new/contract.rs
```diff
@@ -65,8 +65,10 @@ mod tests {
 	#[test]
 	fn test_new_contract_command_execute_success() -> Result<()> {
 		let temp_contract_dir = tempfile::tempdir().expect("Could not create temp dir");
-		let command =
-			NewContractCommand { name: "test_contract".to_string(), path: Some(PathBuf::from(temp_contract_dir.path())) };
+		let command = NewContractCommand {
```
evilrobot-01, line 68: All these formatting changes shouldnt be included in this PR, it adds noise to the review. I assume that they were added due to the CI change, but then they should be fixed in a separate PR.

src/parachains/zombienet.rs
```diff
@@ -515,3 +515,336 @@ impl Source {
...
+	#[tokio::test]
+	async fn test_new_success() -> Result<()> {
+		//cache
+		let temp_dir = tempfile::tempdir().expect("Could not create temp dir");
+		let cache = PathBuf::from(temp_dir.path());
+
+		let zombienet = Zombienet::new(
+			cache.clone(),
+			"./tests/zombienet.toml",
+			Some(&"v1.7.0".to_string()),
```
evilrobot-01, line 534: Might it make sense to use constants to reduce the repetition?

evilrobot-01 (commented):
Great work and thank you!! Biggest distraction for me was the contract-related formatting changes which should not have been included in this PR. I understand why, but it should be out of scope of the changes included.

Apart from that, made some suggestions and comments. Also I ran a code coverage tool and there was only 70% test coverage on `zombienet.rs`, admittedly without the ignored tests which are now still re-running, so need to add more tests which we can also do in a follow up as we look to further improve test coverage across the codebase.

src/commands/new/contract.rs
AlexD10S: Yes, you are right. I run a `cargo fmt` in the last commit that added a lot of noise

tests/wrong_config_no_relay.toml
```diff
@@ -0,0 +1,13 @@
+[[parachains]]
```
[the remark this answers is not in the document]
AlexD10S: Good idea. Done!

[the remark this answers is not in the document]
AlexD10S: File generated just for testing now!

tests/up_parachain.rs
```diff
@@ -0,0 +1,78 @@
...
+	//         println!("All the lines: {:?}", output);
+	//     }
+	// });
+
+	// If after 15 secs is still running probably excution is ok
+	sleep(Duration::from_secs(15)).await;
+	assert!(cmd.try_wait().unwrap().is_none(), "the process should still be running");
+
+	// Stop the process
+	kill(Pid::from_raw(cmd.id().try_into().unwrap()), Signal::SIGINT).unwrap();
```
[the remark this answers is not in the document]
AlexD10S: Good idea. Remove dependency an use this command to kill the process

src/parachains/zombienet.rs
AlexD10S: yes, cleaner

src/parachains/zombienet.rs
```diff
@@ -515,3 +515,419 @@ impl Source {
...
+		let error_message = result_error.err().unwrap();
+		assert_eq!(error_message.root_cause().to_string(), "expected `parachain` to have `id`");
+
+		temp_dir.close()?;
+		Ok(())
+	}
+
+	#[tokio::test]
+	async fn test_relay_chain() -> Result<()> {
+		//cache
```
brunopgalvao, line 602: Remove.

src/parachains/zombienet.rs
```diff
@@ -515,3 +515,419 @@ impl Source {
...
+
+		let config = zombienet.configure();
+		assert!(config.is_ok());
+
+		temp_dir.close()?;
+		Ok(())
+	}
+
+	#[tokio::test]
+	#[ignore] // It takes long time to build
```
brunopgalvao, line 784: So in order to run this test, one would have to specify this test manually?

src/parachains/zombienet.rs
```diff
@@ -515,3 +515,419 @@ impl Source {
...
+		.await?;
+
+		let config = zombienet.configure();
+		assert!(config.is_ok());
+
+		temp_dir.close()?;
+		Ok(())
+	}
+
+	#[tokio::test]
```
brunopgalvao, line 783: Would be nice to add a comment on what this test is spawning or rename the function to indicate that

src/parachains/zombienet.rs
```diff
@@ -515,3 +515,419 @@ impl Source {
...
+
+		let missing_binaries = zombienet.missing_binaries();
+		assert_eq!(missing_binaries.len(), 0);
+
+		temp_dir.close()?;
+		Ok(())
+	}
+
+	#[tokio::test]
+	async fn test_configure() -> Result<()> {
```
brunopgalvao, line 762: Configure what? Add a comment or rename test function.

src/parachains/zombienet.rs
```diff
@@ -515,3 +515,419 @@ impl Source {
...
+		.await?;
+
+		let spawn = zombienet.spawn().await;
+		assert!(spawn.is_err());
+
+		temp_dir.close()?;
+		Ok(())
+	}
+
+	// Source tests
```
brunopgalvao, line 832: What does "Source tests" mean?

src/parachains/zombienet.rs
```diff
@@ -515,3 +515,419 @@ impl Source {
...
+		let result = source.process(&cache).await;
+		assert!(result.is_ok());
+		assert!(temp_dir.path().join(POLKADOT_BINARY).exists());
+
+		temp_dir.close()?;
+		Ok(())
+	}
+
+	#[tokio::test]
+	#[ignore] // It takes long time to build
```
brunopgalvao, line 852: When would we want to run this test? In our CI? If so, let's make sure we add `cargo test -- --ignored` to the CI job.

src/parachains/zombienet.rs
```diff
@@ -515,3 +515,419 @@ impl Source {
...
+		};
+
+		let result = source.process(&cache).await;
+		assert!(result.is_ok());
+		assert!(temp_dir.path().join(POLKADOT_BINARY).exists());
+
+		temp_dir.close()?;
+		Ok(())
+	}
+	#[test]
```
brunopgalvao, line 877: Add newline above this line.

tests/up_parachain.rs
```diff
@@ -0,0 +1,77 @@
...
+
+	//let stdout = cmd.stdout.take().unwrap();
+
+	// thread::spawn(move || {
+	//     let reader = BufReader::new(stdout);
+	//     for line in reader.lines() {
+	//         let output = line.unwrap();
+	//         println!("All the lines: {:?}", output);
+	//     }
+	// });
```
brunopgalvao, line 64: Remove.

src/parachains/zombienet.rs
```diff
@@ -515,3 +515,419 @@ impl Source {
...
+	use super::*;
+	use anyhow::Result;
+
+	const CONFIG_FILE_PATH: &str = "./tests/zombienet.toml";
+	const TESTING_POLKADOT_VERSION: &str = "v1.7.0";
+	const POLKADOT_BINARY: &str = "polkadot-v1.7.0";
+	const POLKADOT_PARACHAIN_BINARY: &str = "polkadot-parachain-v1.7.0";
+
+	#[tokio::test]
+	async fn test_new_success() -> Result<()> {
```
brunopgalvao, line 531: Are we running zombienet tests in our CI?
e.g. `cargo test --package pop-cli --bin pop -- parachains::zombienet::tests --nocapture `

brunopgalvao (changes requested):
Looks like the CI is failing.

src/parachains/zombienet.rs
AlexD10S: Does make sense to run this in the CI? The process build contracts and parachains.

Abhishek opened an issue about that here: https://github.com/r0gue-io/pop-cli/issues/88

src/parachains/zombienet.rs
AlexD10S: Same as a comment below, an Issue is opened here: https://github.com/r0gue-io/pop-cli/issues/88 to discuss the need of running everything or not in the CI

weezy20: The ignore should give a reason within the attribute rather than a comment:
`#[ignore = "<reason>"]`

src/parachains/zombienet.rs
weezy20: Same as above. Ignore reasons must be included within the attribute.

src/commands/new/contract.rs
```diff
@@ -72,6 +72,8 @@ mod tests {
 		let result = command.execute();
 		assert!(result.is_ok());
 
+		temp_contract_dir.close()?;
```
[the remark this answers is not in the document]
AlexD10S: yes, was just in case to don't left anything created after run all tests

src/parachains/zombienet.rs
AlexD10S: Is not running all tests in the CI? `cargo test --verbose`

src/commands/new/contract.rs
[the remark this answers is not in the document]
AlexD10S: But you are right, no need to add this function

AlexD10S:
> Looks like the CI is failing.

Removed the test that was failing. It was failing because is expecting user confirmation to download the binaries. Locally works because this binaries exist in cache.

The removed tests was more than 1h, because was trying all the process: create, build and run parachain (And relay chain). 

Recover the test when we start our e2e battery of tests: https://github.com/r0gue-io/pop-cli/pull/86/commits/50669f496b6cec2fcb8d96b85d41f349345611ae

brunopgalvao (approved):

Issue #37, opened 2024-03-08
Feature: Show the version of Polkadot when generating parachain


evilrobot-01:
It may be useful to know what version of Polkadot to expect.

AlexD10S:
Closed by https://github.com/r0gue-io/pop-cli/pull/89

PR #89, opened 2024-03-29
feat(new parachain): show polkadot version

CinematicCow:
This PR solves for #37. It fetches the latest tag from the repo that is cloned based on the template flag. The release tags are as follows:

| Template | Tag |
|--------|--------|
| base | polkadot-v1.9.0 |
| contract | v0.9.0 |
| frontier | polkadot-v0.9.37 | 

As each repo has it's own naming schema, how do we standardize the naming convention for the end user?

Also I saw there is already a function to get the release tag `Github::get_latest_release` using github's api, but that function was async and I am not familiar with async rust. Hope this PR helps :)


src/commands/new/parachain.rs
```diff
@@ -83,7 +83,7 @@ impl NewParachainCommand {
 				outro_cancel("git signature could not be found. Please configure your git config with your name and email")?;
 			}
 		}
-		spinner.stop("Generation complete");
+		spinner.stop(format!("Generation complete using {}", tag));
```
weezy20, line 86: Suggestion: can the `polkadot` version tag be logged as a separate line? 
`Generation Complete` - on spinner stop
then 
`Version: polkadot-v*`

CinematicCow: Version is now logged clearly on a separate line.

Daanvdplas (commented):
Hi @CinematicCow, amazing to see you again!

src/helpers.rs
```diff
@@ -43,11 +44,32 @@ pub(crate) fn write_to_file<'a>(path: &Path, contents: &'a str) {
...
+	Ok(release)
+}
+
+/// Fetch the latest release from a repository
+fn fetch_tag(repo: &Repository) -> String {
+	let mut release = String::new();
+
+	let version_reg = Regex::new(r"v\d+\.\d+\.\d+").unwrap();
+
+	for tag in repo.tag_names(None).unwrap().iter() {
```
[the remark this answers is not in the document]
weezy20: So I'd like to add here, that while this works, `fetch_tag` needs to be Optional (`Option<String>`) and that should be enough of a change to remove all instances of `unwrap` here. Template generation should not fail because `fetch_tag` failed. I believe this should've been in the issue description rather than here, but the problem is, if in future, we decide to support a template that doesn't follow the regex `"v\d+\.\d+\.\d+"`, then this code will also be modified so it's like a pending todo.

weezy20, line 64: Why not iterate in reverse order to start from latest tag/release?

src/engines/parachain_engine.rs
```diff
@@ -30,15 +34,15 @@ pub fn instantiate_template_dir(template: &Template, target: &Path, config: Conf
...
-	Ok(())
+	Ok(tag)
 }
 
-pub fn instantiate_base_template(target: &Path, config: Config) -> Result<()> {
+pub fn instantiate_base_template(target: &Path, config: Config) -> Result<String> {
 	let temp_dir = ::tempfile::TempDir::new_in(std::env::temp_dir())?;
 	let source = temp_dir.path();
-	clone_and_degit("https://github.com/r0gue-io/base-parachain", source)?;
+	let tag = clone_and_degit("https://github.com/r0gue-io/base-parachain", source)?;
```
weezy20, line 45: Same reason as down below. `tag` needs to be an Option i.e. `clone_and_degit` returns a `Result<Option<String>>`

weezy20 (changes requested):
Approach looks good. 
1. Iterator should be in reverse order to search starting from latest release
2. `fetch_tag` and all consumers should return and expect an Option, in case the regex fails or template source doesn't use tags.

src/helpers.rs
```diff
@@ -43,11 +44,30 @@ pub(crate) fn write_to_file<'a>(path: &Path, contents: &'a str) {
...
+	let release = fetch_latest_tag(&repo);
+
 	let git_dir = repo.path();
 	fs::remove_dir_all(&git_dir)?;
-	Ok(())
+	Ok(release)
+}
+
+/// Fetch the latest release from a repository
+fn fetch_latest_tag(repo: &Repository) -> Option<String> {
```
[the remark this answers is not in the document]
CinematicCow: yes it does, but as you were already cloning the repo locally, the required info was there, so I did not see the need for another network call. Also I am not familiar with async rust as much.

AlexD10S (approved):

src/helpers.rs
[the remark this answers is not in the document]
AlexD10S: I see you are right with that. However, there is one thing to consider here, you are fetching the tags which might be different from releases:

https://stackoverflow.com/questions/18506508/whats-the-difference-between-tag-and-release

weezy20 (approved):

Issue #88, opened 2024-03-29
Cargo test takes too long to test


weezy20:
Somewhat related to #81 and others, but `cargo test` is supposed to check and provide quick feedback for core features of `pop` for developers of pop rather than end-users. Since we are now building and checking a contract and a parachain when running `cargo test` it's extremely inefficient to check for coding errors when working on any feature.

Either feature gate or `#[ignore]` these long running tests so that they can be manually run by interested people, rather than having it run by default on `cargo test`. 

Additionally, modify CI to include `#[ignore]`'d tests if so desired. 

Also something worth considering is why are these tests not a part of the parachain/contract templates themselves?

brunopgalvao:
For e2e tests, I would not be against having a separate repo that has all the e2e test code which has a CI which runs independently from pop-cli repo.

AlexD10S:
Close with https://github.com/r0gue-io/pop-cli/pull/90

PR #90, opened 2024-04-01
fix: cargo test takes too long to test

AlexD10S:
Fixing Issue https://github.com/r0gue-io/pop-cli/issues/88

- Run only unit tests when `cargo test`
- unit tests that are a bit slower, build a contract move it into feature `unit_contract`. To run `cargo test --features unit_contract `
- Run e2e tests using a feature flag `cargo test --features e2e_contract` for contracts and `cargo test --features e2e_parachain` for parachains
- Moved the test that builds a parachain to e2e test

- In the CI run everything `cargo test --all-features -- --include-ignored`

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -65,4 +65,4 @@ jobs:
       run: cargo build
 
     - name: Run unit tests
-      run: cargo test --bins
+      run: cargo test --all-features -- --include-ignored
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -65,4 +65,4 @@ jobs:
       run: cargo build
 
     - name: Run tests
-      run: cargo test --all-features -- --include-ignored
+      run: cargo test --all-features
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -64,5 +64,5 @@ jobs:
     - name: Build default features
       run: cargo build
 
-    - name: Run unit tests
+    - name: Run tests
       run: cargo test --all-features -- --include-ignored
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -64,5 +64,5 @@ jobs:
     - name: Build default features
       run: cargo build
 
-    - name: Run tests
-      run: cargo test --verbose
+    - name: Run unit tests
+      run: cargo test --bins
--- a/README.md
+++ b/README.md
@@ -204,3 +204,32 @@ pop up parachain -f ./tests/zombienet.toml -p https://github.com/r0gue-io/pop-no
 
 > :information_source: Pop CLI will automatically source the necessary polkadot binaries. Currently, these will be built
 > if on a non-linux system.
+
+
+## Testing Pop CLI 
+
+To test the tool locally.
+
+Run the unit tests:
+
+```sh
+cargo test
+```
+
+Run the contracts e2e tests:
+
+```sh
+cargo test --features e2e_contract
+```
+
+Run the parachain e2e tests:
+
+```sh
+cargo test --features e2e_parachain
+```
+
+Run all tests, including the ignored ones:
+
+```sh
+cargo test --all-features -- --include-ignored
+```
--- a/README.md
+++ b/README.md
@@ -216,6 +216,12 @@ Run the unit tests:
 cargo test
 ```
 
+Run only contracts unit tests:
+
+```sh
+cargo test --features unit_contract
+```
+
 Run the contracts e2e tests:
 
 ```sh
@@ -228,8 +234,8 @@ Run the parachain e2e tests:
 cargo test --features e2e_parachain
 ```
 
-Run all tests, including the ignored ones:
+Run all tests:
 
 ```sh
-cargo test --all-features -- --include-ignored
+cargo test --all-features
 ```
--- a/tests/build_contract.rs
+++ b/tests/build_contract.rs
@@ -1,3 +1,4 @@
+#![cfg(feature = "e2e_contract")]
 use anyhow::{Error, Result};
 use assert_cmd::Command;
 use predicates::prelude::*;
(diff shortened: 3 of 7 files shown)


weezy20:
`cargo test --bins` runs fast with 3 ignored test. If I run `cargo test`, I get the same followed by a somewhat time consuming:
```shell
running 3 tests
test test_contract_build_fails_if_no_contract_exists ... ok
test test_contract_build_specify_path ... ok
test test_contract_build ... ok
```
Why is `test_contract_build` running even after being `#[ignore]`d?

.github/workflows/build.yml
```diff
@@ -64,5 +64,5 @@ jobs:
     - name: Build default features
       run: cargo build
 
-    - name: Run tests
-      run: cargo test --verbose
\ No newline at end of file
+    - name: Run unit tests
+      run: cargo test --bins
```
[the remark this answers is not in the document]
AlexD10S: test all binaries, in our cases will test everything except the integration tests located under folder /tests:
https://github.com/rust-lang/cargo/issues/10491#issuecomment-1074078912

AlexD10S:
> `cargo test --bins` runs fast with 3 ignored test. However, if I run `cargo test`, I get the same followed by a somewhat time consuming:
> 
> ```shell
> running 3 tests
> test test_contract_build_fails_if_no_contract_exists ... ok
> test test_contract_build_specify_path ... ok
> test test_contract_build ... ok
> ```
> 
> Why is `test_contract_build` running even after being `#[ignore]`d?

The ignored one was not being run, there were 2 tests called `test_contract_build`. I changed that. To run only unit tests now `cargo test`, to run e2e related to contracts: `cargo test --features e2e_contract` and to run everything ignored too: `cargo test --all-features -- --include-ignored`.

My doubt is still what to run in the CI, once we have the up tests run everything `cargo test --all-features -- --include-ignored` can take more than 2h but we will be sure the product is stable.

tests/build_contract.rs
```diff
@@ -16,7 +16,8 @@ fn setup_test_environment() -> Result<tempfile::TempDir, Error> {
 }
 
 #[test]
-fn test_contract_build() -> Result<(), Error> {
+#[cfg_attr(not(feature = "e2e_contract"), ignore)]
```
weezy20, line 19: Instead of `#[cfg_attr(not(feature = "e2e_contract"), ignore)]` on top of every test, one module `cfg` : `#![cfg(feature = "e2e_contract")]` suffices.

README.md
```diff
@@ -204,3 +204,32 @@ pop up parachain -f ./tests/zombienet.toml -p https://github.com/r0gue-io/pop-no
...
+Run the contracts e2e tests:
+
+```sh
+cargo test --features e2e_contract
+```
+
+Run the parachain e2e tests:
+
+```sh
+cargo test --features e2e_parachain
```
weezy20, line 228: No `e2e_parachain` gated test code found.

.github/workflows/build.yml
```diff
@@ -64,5 +64,5 @@ jobs:
     - name: Build default features
       run: cargo build
 
-    - name: Run tests
-      run: cargo test --verbose
\ No newline at end of file
+    - name: Run unit tests
```
weezy20, line 67: `name: Run tests` as this will run both unit & e2e tests.

weezy20 (changes requested):

README.md
weezy20: Suggest moving out this unit test:
```rs
        #[test]
	#[ignore = "Build parachain takes long time to build"]
	fn test_parachain_build_after_instantiating_template() -> Result<()> {
		let temp_dir =
			setup_template_and_instantiate().expect("Failed to setup template and instantiate");
		let build = build_parachain(&Some(temp_dir.path().to_path_buf()));
		assert!(build.is_ok(), "Result should be Ok");
		Ok(())
	}
```
 into it's own module `tests/build_parachain.rs` and gating it with `e2e_parachain`

weezy20:
>My doubt is still what to run in the CI, once we have the up tests run everything cargo test --all-features -- --include-ignored can take more than 2h but we will be sure the product is stable.

So here's my suggestion. For product stability, i.e. sanity of generated template builds, it doesn't make sense to have them run whilst a PR is open for review or constantly being worked on. This slows us down. 

At bare minimum unit tests must be run on PRs. 
For e2e tests, specifically `build_*` tests, they can be merged only on merge to main. See https://stackoverflow.com/a/61565445/6591533

brunopgalvao (commented):
I am not a fan of using `ignore`. I would rather use a tiered approach.

```
cargo test
cargo test --features all_tests
cargo test --features all_unit_tests
cargo test --features all_contract_unit_tests
cargo test --features all_parachain_unit_tests
cargo test --features all_e2e_tests
cargo test --features all_e2e_contract_tests
cargo test --features all_e2e_parachain_tests
```

OR

```
cargo test
cargo test --features tier1_tests
cargo test --features tier2_tests
cargo test --features tier3_tests
```
where tier 1 would be least amount of tests and tier 3 would be full testing.

AlexD10S:
> I am not a fan of using `ignore`. I would rather use a tiered approach.
> 
> ```
> cargo test
> cargo test --features all_tests
> cargo test --features all_unit_tests
> cargo test --features all_contract_unit_tests
> cargo test --features all_parachain_unit_tests
> cargo test --features all_e2e_tests
> cargo test --features all_e2e_contract_tests
> cargo test --features all_e2e_parachain_tests
> ```
> 
> OR
> 
> ```
> cargo test
> cargo test --features tier1_tests
> cargo test --features tier2_tests
> cargo test --features tier3_tests
> ```
> 
> where tier 1 would be least amount of tests and tier 3 would be full testing.

Done! No more ignored tests

README.md
```diff
@@ -206,9 +206,43 @@ pop up parachain -f ./tests/zombienet.toml -p https://github.com/r0gue-io/pop-no
...
+
+To test the tool locally.
+
+Run the unit tests:
+
+```sh
+cargo test
+```
+
+Run only contracts unit tests:
```
weezy20, line 220: nit: Some more information here can be useful. For eg: "Runs the unit tests within the ink smart contract"

weezy20 (approved):
Lgtm. Thanks @AlexD10S

PR #92, opened 2024-04-01
Feature: PopDb - a caching system for pop-cli

### LINKED ISSUE #9: Cache to improve UX
Internally to cache the path where the node was generated after execute `pop new parachain`. Then, on build, if the cache is non empty, we use that path.

Useful for `pop build parachain` https://github.com/r0gue-io/pop-cli/issues/2 to don't have to specify the path when build the parachain recently generated.

And for `pop add pallet` https://github.com/r0gue-io/pop-cli/issues/7 because for the runtime manipulation which will be part of add-pallet, either the runtime needs to be automatically inferred (depending on if pop new was called), or it needs to take a path arg.

### LINKED ISSUE #91: Pop Build - Parachain/Contract cargo project detection on current working dir
A discussion rather than an issue.

Currently, if `pop build` is invoked without a `path`, it builds any cargo project in the current working dir: 
```rs
pub fn build_parachain(path: Option<PathBuf>) -> anyhow::Result<()> {
	cmd("cargo", vec!["build", "--release"])
		.dir(path.clone().unwrap_or("./".into()))
		.run()?;
	Ok(())
}
```
This is fine if the current working dir is a parachain project, but is unexpected or unintended behaviour when invoked in some other cargo project. `pop build`, for the sake of correctness, should be ideally scoped only to build `contract` and `parachain` cargo projects. 
I understand that this is not really a problem, but this issue is meant for discussion on if it's worth the additional code or not.

weezy20:
`PopDb` is a caching system to cache recently used file paths with `new` commands like `pop new parachain` that are saved and used later when invoked with action commands like `build`, `test`, `add` etc. without explicit path args. Since `PopDb` is a generic key-value database, it maybe used for things beyond this, however this PR is just creating step 1 for `PopDb`. 

Usage : 

```rs
let popdb = PopDb::open_or_init();
// Set a key-value pair 
popdb.set_*(value);
// Fetch the same
popdb.get_*(key);
```

The preference for a file path is always in this order:
1. User cli args
4. Current working directory
5. PopDb 

Note: This preference is not possible because step `2` is always going to execute in any cargo project (see issue #91). ~~Currently, the preference is 1 -> 3 -> 2 until a `helper` function is created to close #91.~~

Tasks (requires #91): 
Implement PopDb cache look up in the following actions after `is_parachain(current_dir())` or `is_contract(current_dir())` checks have failed. 
- [ ] Build
- [ ] Test
- [ ] Up
- [ ] Add ( unless `add-pallet` is merged this cannot be done )

Implement PopDb set in the following actions
- [x] `new parachain`
- [x] `new contract`
- [x] `new pallet`

Additional notes:

Error handling for `PopDb` is invisible to the user unless a `LOG_TARGET` of `Debug` or above is chosen. This is because errors involved with this system should have little to no effect on the user experience and are solely meant for developers of `pop-cli`. Fundamentally, there are two classes of errors that could happen: 
1. Errors opening the filesystem path `/tmp/pop-cli-db` 
2. cache misses or garbage values from cache. 
The first kind of errors are unlikely to happen, as usually, users are granted read-write permissions on `/tmp`. However the second kind of errors are a bit more subtle, and they require careful consideration and invalidation in case certain criteria are not met. One instance of this is invalid UTF-8 in `--path`. Storing `Path` as `bytes` involve lossy compression, and for invalid characters, `String::from_utf8_lossy` may yield unusable values.

Closes #9


weezy20:
Note for reviewers: This cache has two sets of methods `set_*` and `get_*`. We `set` values when creating a new parachain, contract or a pallet, and the idea is to use these values on `build | test | check etc` actions if and only if `current_dir()` doesn't meet the criteria (#91). As a consequence, all the `get_*` methods are unused until #91 is closed.

AlexD10S:
Looking good, but logic is not completed, should I move it into Draft again?

And just for curiosity, what other cases you have in mind for PopDb beyond keeping the file paths for `build` and `test`?

PR #98, opened 2024-04-03
feat: guide user for parachain creation

AlexD10S:
The main reason of this PR is to create an interactive user guide for parachain generation and have an structure to manage different Template Providers.

-  Interactive user guide for parachain generation when no attributes has been input by the user: `pop new parachain`.

Prompt the user to pick customizable flags too if he has selected the Pop Base template parachain, and display the user latest releases to pick which one want to generate.
- Structure to manage different Template Providers: Pop (By default), and Parity now
           `pop new parachain my-parachain parity -t fpt`

Ready to integrate in the future templates from OpenZeppelin, Tanssi..... for example:
`pop new parachain my-parachain openzeppelin -t template`

(For the future, the work for OZ is done in this commit: https://github.com/r0gue-io/pop-cli/pull/98/commits/f2ecb4364f1887eb61c6c328b3441ab9e41a3046)

Rest of Pop templates without needing to indicate `pop` as a provider
Simple specifying `pop new parachain my-parachain -t contracts`  will generate parachain using pop contracts once it exists. 

- Warning notes:
   - Message when adding customizable flags in the input for provider templates, saying is not possible:
      `pop new parachain my-app openzeppelin template --symbol DOT --decimals 6 --endowment 1_000_000_000` right now just generates the template provider parachain but doesn't show any warning message.
   - At the end if is a external provider template: 

> NOTE: the resulting parachain has not been guaranteed to be audited or reviewed for security vulnerabilities. 
> Please consult the source repository at https://github.com/org-name/repo to assess production suitability.


evilrobot-01:
There is a lot of repetition in the main readme and the pop-cli crate readme. We need to think on how to resolve this - perhaps a build script / commit hook which copies it from cli to root?

crates/pop-parachains/src/utils/git.rs
```diff
@@ -46,16 +46,39 @@ impl Git {
...
+	) -> Result<Option<String>> {
 		let repo = match Repository::clone(&["https://github.com/", url].concat(), target) {
 			Ok(repo) => repo,
 			Err(_e) => Self::ssh_clone_and_degit(url, target)?,
 		};
 		// fetch tags from remote
 		let release = Self::fetch_latest_tag(&repo);
 
+		if tag_version.is_some() {
+			let tag = tag_version.clone().unwrap();
```
weezy20, line 62: Instead of 
```rs
if tag_version.is_some() {
  let tag = tag_version.clone().unwrap();
  ```
  why not use `if let`? 
  ```rs
  if let Some(ref tag) = tag_version {
  
  ```

File: crates/pop-parachains/src/templates.rs [hunk 2]
```diff
@@ -0,0 +1,114 @@
...
+pub struct Config {
+	pub symbol: String,
+	pub decimals: u8,
+	pub initial_endowment: String,
+}
+
+#[derive(Clone, Debug, PartialEq)]
+pub enum Template {
+	// Pop
+	Base,
```
weezy20, line 12: `BasePop` maybe consistent with the naming scheme of the other two variants `Partiy*`  ?

crates/pop-parachains/src/templates.rs
```diff
@@ -0,0 +1,114 @@
...
+		}
+	}
+	pub fn repository_url(&self) -> &str {
+		match &self {
+			Template::Base => "r0gue-io/base-parachain",
+			Template::ParityContracts => "paritytech/substrate-contracts-node",
+			Template::ParityFPT => "paritytech/frontier-parachain-template",
+		}
+	}
+}
```
weezy20, line 39: Since we are making this code to be future compatible with different providers and templates, I suggest a more refined approach rather than relying on manual matches. Look at `strum` or string enums an already imported library which I have used in the past to make string associations to enum variants. 

One enum variant can be made to match to a `url`, a `provider` and ofc any other name as desired. This functionality need not  be handwritten.

crates/pop-parachains/src/templates.rs
```diff
@@ -0,0 +1,114 @@
...
+		}
+	}
+	pub fn from(provider_name: &str) -> Self {
+		match provider_name {
+			"Pop" => Provider::Pop,
+			"Parity" => Provider::Parity,
+			_ => Provider::Pop,
+		}
+	}
+}
```
weezy20, line 61: Same as above. `strum` can help out here.

weezy20 (changes requested):

weezy20:
Agree with Frank on the Readme duplication.

README.md
```diff
@@ -18,32 +18,32 @@ cargo install --locked --git https://github.com/r0gue-io/pop-cli
...
-cd my-app
-pop build parachain
+# Get Parity's pallet-contracts enabled parachain template
+pop new parachain my-app parity -t cpt
+# Get Parity's evm compatible parachain template
+pop new parachain my-app parity -t fpt
 ```
 
-You can also customize your parachain by providing config options for token symbol (as it appears in chain metadata),
+For POP templates you can also customize your parachain by providing config options for token symbol (as it appears in chain metadata),
```
brunopgalvao, line 46: "For Pop templates you"

crates/pop-cli/README.md
```diff
@@ -18,32 +18,32 @@ cargo install --locked --git https://github.com/r0gue-io/pop-cli
...
-cd my-app
-pop build parachain
+# Get Parity's pallet-contracts enabled parachain template
+pop new parachain my-app parity -t cpt
+# Get Parity's evm compatible parachain template
+pop new parachain my-app parity -t fpt
 ```
 
-You can also customize your parachain by providing config options for token symbol (as it appears in chain metadata),
+For POP templates you can also customize your parachain by providing config options for token symbol (as it appears in chain metadata),
```
brunopgalvao, line 46: "For Pop templates you"

brunopgalvao (approved):
Excellent work!
```
┌   Pop CLI : Generate a parachain
│
◆  Select a template provider: 
│  ● Pop (An all-in-one tool for Polkadot development. 1 available options)
│  ○ Parity 
└
```
From the above prompt, the user will not know what the available options are.
If the select Pop in order to see what the available option (template) is:
```
◆  Select the type of parachain:
│  ● Standard Template (A standard parachain)
└
```
They can now see the Pop templates but there is no option to go back.

Is there a way to let them know what templates are available for each provider before they commit to it? Or perhaps add a "Go Back" option

README.md
```diff
@@ -18,32 +18,32 @@ cargo install --locked --git https://github.com/r0gue-io/pop-cli
...
-or
-
+Some examples are: 
 ```sh
-cd my-app
-pop build parachain
+# Get Parity's pallet-contracts enabled parachain template
+pop new parachain my-app parity -t cpt
+# Get Parity's evm compatible parachain template
+pop new parachain my-app parity -t fpt
```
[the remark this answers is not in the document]
AlexD10S: For me no, is more clear to pick the provider and then the template associate to it,  but maybe would be good to open a discussion and see users feedback.

Ideally would be to don't even have to add the flag -t:
`pop new p my-app parity cpt`

crates/pop-cli/src/commands/new/contract.rs
```diff
@@ -18,7 +18,7 @@ pub struct NewContractCommand {
 }
 
 impl NewContractCommand {
-	pub(crate) fn execute(self) -> anyhow::Result<()> {
+	pub(crate) async fn execute(&self) -> anyhow::Result<()> {
```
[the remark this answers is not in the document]
AlexD10S: We have  `fn execute(&self)` in all no? we should keep consistency and use the same all around the code

crates/pop-parachains/src/templates.rs
AlexD10S: No, Pop are the default ones of the tool: pop base, pop contracts, pop evm and pop nfts. 
ParityContracts is called like this only because Contracts will exist with Pop as a provider

crates/pop-cli/src/commands/new/contract.rs
```diff
@@ -49,7 +49,7 @@ impl NewContractCommand {
 		fs::create_dir_all(contract_path.as_path())?;
 		let mut spinner = cliclack::spinner();
 		spinner.start("Generating contract...");
-		create_smart_contract(self.name, contract_path.as_path())?;
+		create_smart_contract(self.name.clone(), contract_path.as_path())?;
```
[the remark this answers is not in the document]
AlexD10S: Fixed in Frank's PR

crates/pop-cli/README.md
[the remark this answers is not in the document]
AlexD10S: In my opinion no need to add extra information. Above we are talking about other providers, is clear to me Pop templates are the native templates of the tool

crates/pop-parachains/src/utils/git.rs
AlexD10S: Fixed in Frank's PR

crates/pop-parachains/src/templates.rs
AlexD10S: Good point. Same as above, solved after Frank's refactor

crates/pop-parachains/src/templates.rs
```diff
@@ -0,0 +1,114 @@
...
+		provider = Provider::Parity;
+		assert_eq!(provider.default_template(), Template::ParityContracts);
+	}
+
+	#[test]
+	fn test_convert_string_to_provider() {
+		assert_eq!(Provider::from("Pop"), Provider::Pop);
+		assert_eq!(Provider::from(""), Provider::Pop);
+		assert_eq!(Provider::from("Parity"), Provider::Parity);
+	}
```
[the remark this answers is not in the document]
AlexD10S: Are you suggesting remove the tests?

AlexD10S:
> There is a lot of repetition in the main readme and the pop-cli crate readme. We need to think on how to resolve this - perhaps a build script / commit hook which copies it from cli to root?

The README was exactly the same for root and pop-cli. I created now a symbolic link to have the same as the root without having the repetition, like cargo-contract has https://github.com/use-ink/cargo-contract/blob/master/crates/cargo-contract/README.md?plain=1

crates/pop-cli/src/commands/new/contract.rs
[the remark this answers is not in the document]
weezy20: The point is for build commands it should've been `self` from the start as after that match branch there's no need to keep it around. There's no benefit in taking references here, rather it creates the need for `clone()`s when owned values are required further down the pipeline.

README.md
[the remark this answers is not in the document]
weezy20: For a new user, they might not feel this as intuitive as it could've been. If I am new, I would want to start a new parachain without knowing the additional concept of a provider (which in this case seems mandatory). In such a case a default should've been selected for me.

crates/pop-parachains/src/templates.rs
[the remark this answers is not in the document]
weezy20: Maybe not all, but a few right off the bat don't make sense to me when a string is already associated with an enum variant. I will have to review the new additions before I give my definitive opinion.

README.md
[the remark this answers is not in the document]
AlexD10S: There is a default provider already, `pop`.
If you are new and don't want to use a specific provider, you don't have to know about that, no problem.
`pop new parachain my-app ` will generate the base parachain and `pop new parachain my-app -t contracts` the contracts parachain from Pop (When it exists)

crates/pop-cli/src/commands/new/contract.rs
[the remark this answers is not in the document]
AlexD10S: Let's make this change in a new PR as it involves refactor it in all the commands and this is out of scope in this changes.

crates/pop-cli/src/commands/new/parachain.rs
```diff
@@ -55,58 +45,125 @@ pub struct NewParachainCommand {
...
+				"{} {} available option(s)",
+				provider.description(),
+				provider.templates().len()
+			),
+		);
 	}
+	let provider = prompt.interact()?;
+	let template = display_select_options(provider)?;
+
+	let url = url::Url::parse(&["https://github.com/", template.repository_url()?].concat())
```
evilrobot-01, line 135: Can probably just move the `https://github.com/` prefix into the repository string to eliminate this complexity.

crates/pop-cli/src/helpers.rs
```diff
@@ -0,0 +1,138 @@
+// SPDX-License-Identifier: GPL-3.0
```
evilrobot-01, line 1: Most of these 'helpers' are specific to `new parachain` and should therefore exist in the `commands::new::parachain` module instead. They are not general purpose to the whole crate and I fear this general approach is only going to result in future technical debt.

evilrobot-01 (commented):
Biggest concern for me is the fact that functions are being dumped into `helpers.rs` rather than being modularised and how this will affect future development productivity.

crates/pop-parachains/src/templates.rs
```diff
@@ -0,0 +1,174 @@
...
+		props(Provider = "Parity", Repository = "paritytech/substrate-contracts-node")
+	)]
+	ParityContracts,
+	#[strum(
+		serialize = "fpt",
+		message = "EVM",
+		detailed_message = "Template node for a Frontier (EVM) based parachain.",
+		props(Provider = "Parity", Repository = "paritytech/frontier-parachain-template")
+	)]
+	ParityFPT,
```
weezy20, line 96: Like discussed before, We included `ParityFPT` here despite it being unmaintained. Does that motivate to also include EPT? 
```suggestion
	ParityFPT,
    ParityEPT,
```

crates/pop-cli/src/commands/new/parachain.rs
AlexD10S: Yes, good idea! Added in the Template enum the two methods to return the url needed, so will make mantaining it easier

crates/pop-cli/src/helpers.rs
AlexD10S: You are right, all methods are used only in `commands::new::parachain`, so moved there

crates/pop-parachains/src/templates.rs
AlexD10S: Good point. But people is using this templates, I'd suggest to keep them at least until we have the Pop contracts and evm templates ready.

evilrobot-01: Maybe we can update the description with a [deprecated] suffix.

crates/pop-parachains/src/templates.rs
```diff
@@ -108,8 +108,13 @@ impl Template {
 		self.get_str("Provider") == Some(provider.to_string().as_str())
 	}
 
-	pub fn repository_url(&self) -> Result<&str, Error> {
-		self.get_str("Repository").ok_or(Error::RepositoryMissing)
+	pub fn repository_url(&self) -> Result<String, Error> {
+		Ok(["https://github.com/", self.get_str("Repository").ok_or(Error::RepositoryMissing)?]
```
evilrobot-01, line 112: Still dont think this is necessary when you can just define `Repository = "https://github.com/paritytech/frontier-parachain-template"` on the enum. There is a minor additional size but it simplifies the code, reduces the concat at runtime and also allows any non-GH repo url to be defined.

evilrobot-01: Aaah, I now see it is probably due to the ssh url

crates/pop-parachains/src/utils/git.rs
```diff
@@ -12,6 +13,8 @@ use url::Url;
 
 pub struct Git;
 impl Git {
+	const GIT_SSH: &'static str = "git@github.com:";
```
evilrobot-01, line 16: This is a GH specific string rather than git, so suggest renaming to GITHUB_SSH and then reusing it from the ssh repo function.

evilrobot-01: There is a GitHub type somewhere, so maybe its better to be defined there.

crates/pop-parachains/src/utils/git.rs
```diff
@@ -27,10 +30,9 @@ impl Git {
 		}
 		Ok(())
 	}
-
 	pub(crate) fn ssh_clone(url: &Url, working_dir: &Path, branch: Option<&str>) -> Result<()> {
 		// Change the url to the ssh url with git@github.com: prefix, remove / from path and adding .git as suffix
-		let ssh_url = ["git@github.com:", &url.path()[1..], ".git"].concat();
+		let ssh_url = [Self::GIT_SSH, &url.path()[1..], ".git"].concat();
```
evilrobot-01, line 35: Strictly speaking this is currently a github url rather than git, so the function should probably take it as a parameter so that the function can be used by any git provider. Probably doesnt matter due to the function being non-public.

crates/pop-parachains/src/utils/git.rs
```diff
@@ -47,13 +49,13 @@ impl Git {
 	}
 	/// Clone `url` into `target` and degit it
 	pub fn clone_and_degit(
-		url: &str,
+		template: &Template,
```
evilrobot-01, line 52: Not a fan of this. Git should not need to know about a `Template` type, only a url.

crates/pop-parachains/src/templates.rs
AlexD10S: Still, your points are very valid. Does make sense to have something like this to avoid the concat:
```rs
#[default]
	#[strum(
		serialize = "base",
		message = "Standard",
		detailed_message = "A standard parachain",
		props(
			Provider = "Pop",
			Repository = "https://github.com/r0gue-io/base-parachain",
			SSHRepository = "git@github.com:r0gue-io/base-parachain.git"
		)
	)]
	Base,
```

evilrobot-01 (commented):
A few additional improvements can still be made imo. 

The git functionality shouldnt be tightly coupled to templates when it only needs a url. It shouldnt also know about GitHub, which is a platform built over git. These are somewhat nitpicky, but oversharing types / responsibilities leads to less reusability and harder to maintain code.

crates/pop-parachains/src/templates.rs
AlexD10S: But still in the Parity repo, and they haven't mark it as deprecated

evilrobot-01: Not been updated in ages tho, so providing that as an option could perhaps be seen as an endorsement.

crates/pop-parachains/src/templates.rs
evilrobot-01: Looking at the usage, the string is being parsed into a `Url` type. That can easily be used to create a new url in the case of ssh - something like `format!("git@{}:{}", url.host_str(), url.path())`

https://docs.rs/url/latest/url/

crates/pop-parachains/src/utils/git.rs
AlexD10S: You are right, should only pass a URL

crates/pop-parachains/src/templates.rs
AlexD10S: This is a nice approach

crates/pop-parachains/src/templates.rs
AlexD10S: We are currently showing a warning message after generating a parachain:
```
NOTE: the resulting parachain is not guaranteed to be audited or reviewed for security vulnerabilities. ──────────────────────────────────────────────────────╮
│                                                                                                                                                                │
│  Please consult the source repository at https://github.com/paritytech/substrate-contracts-node to assess production suitability and licensing restrictions.
```

I think with this is enough, but I can add `deprecated` in the help description of the Parity templates if you both think is needed

evilrobot-01: Only downside is that this is only shown after the parachain is generated rather than before.

I'm happy to leave as is and revisit separately.

crates/pop-parachains/src/utils/git.rs
```diff
@@ -219,7 +220,7 @@ impl GitHub {
 		format!("{}/releases/download/{tag}/{artifact}", repo.as_str())
 	}
 	pub(crate) fn convert_to_shh_url(url: &Url) -> String {
```
evilrobot-01, line 222: typo, should be ssh

AlexD10S: Sorry for that! Fixed

crates/pop-parachains/src/templates.rs
AlexD10S: Added a small label in the --help:
` [PROVIDER]  Template provider. Options are pop or parity (deprecated). [default: pop]`
And in user prompt:
```
 Select a template provider: 
│  ○ Pop 
│  ● Parity (Solutions for a trust-free world. 2 available option(s) [deprecated])
```

weezy20 (approved):
LGTM!

PR #99, opened 2024-04-03
refactor: various improvements

evilrobot-01:
An additional observation is that the OZ release does not include a zombienet config file as it appears in their main branch, so we should probably just download a copy manually to show the full flow.


AlexD10S (approved):
Looks great!

PR #105, opened 2024-04-10
Helpers: is_contract & is_parachain

### LINKED ISSUE #91: Pop Build - Parachain/Contract cargo project detection on current working dir

weezy20:
Close #91 
Unblock #92

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5450,7 +5450,7 @@ dependencies = [
  "symlink",
  "tempfile",
  "tokio",
- "toml_edit 0.22.6",
+ "toml_edit 0.22.9",
  "tracing-subscriber 0.3.18",
  "url",
  "walkdir",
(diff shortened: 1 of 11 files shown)


AlexD10S:
I am not sure about adding a method that is not used, it can add dead code. 
I prefer to add the code in the PR where the functionality is going to be used, my suggestion is to merge this in https://github.com/r0gue-io/pop-cli/pull/92 instead of main.

weezy20:
@AlexD10S Okay will do that.

Issue #8, opened 2024-02-29
Review Project Error Handling


AlexD10S:
Check error handling in all the code, for example the usage of `expect`.  For example If something fails in the middle of an action shouldn't do a panic leaving files.

brunopgalvao:
Example:
- https://github.com/r0gue-io/pop-cli-archive/blob/frank/feat-pop-up/src/parachain/zombienet.rs

PR #111, opened 2024-04-17
chore: use `thiserror` within crates

brunopgalvao:
`thiserror` is good for libraries:
- https://google.github.io/comprehensive-rust/error-handling/thiserror-and-anyhow.html

closes: https://github.com/r0gue-io/pop-cli/issues/8


crates/pop-contracts/src/new.rs
```diff
@@ -1,30 +1,51 @@
...
-/// Create a new smart contract at `target`
-pub fn create_smart_contract(name: String, target: &Path) -> anyhow::Result<()> {
-	// In this code, out_dir will automatically join `name` to `target`,
-	// which is created prior to the call to this function
-	// So we must pass `target.parent()`
-	new_contract_project(&name, target.canonicalize()?.parent())
+#[derive(Error, Debug)]
+pub enum Error {
+	#[error("Failed to create new contract project: {0}")]
+	NewContractError(String),
```
weezy20, line 8: I suggest to omit the `Error` suffix. When you're matching onn the error:
```rs
match fallible_operation() {
  Ok(_) => {}
  Err(e) => match e {
    Error::New,
    Error::Io,
    }
 ```
 helps eliminate some of the redundant visual noise.

File: crates/pop-contracts/src/utils/helpers.rs [hunk 2]
```diff
@@ -1,33 +1,40 @@
...
 
-// If the user specifies a path (which is not the current directory), it will have to manually
-// add a Cargo.toml file. If not provided, pop-cli will ask the user for a specific path. or ask
-// to the user the specific path (Like cargo-contract does)
-pub fn get_manifest_path(path: &Option<PathBuf>) -> anyhow::Result<ManifestPath> {
-	if path.is_some() {
-		let full_path: PathBuf =
-			PathBuf::from(path.as_ref().unwrap().to_string_lossy().to_string() + "/Cargo.toml");
+#[derive(Error, Debug)]
+pub enum Error {
```
weezy20, line 9: Why do individual modules have their own definition of an `Error` enum? For one crate, the idiomatic design is to have a single `error.rs` file, and collate all errors into that one enum. You can use struct variants if you feel one variant of `Error` can have multiple applications but fall under the same module.

brunopgalvao: Done.

AlexD10S:
Nice work! 
I see here the tests passes in the CI but for me locally the test `test_new_parachain_command_execute` is failing, can you double check is passing for you?

AlexD10S (changes requested):

crates/pop-contracts/src/test.rs
```diff
@@ -1,46 +1,65 @@
...
-	cmd("cargo", vec!["test"]).dir(path.clone().unwrap_or("./".into())).run()?;
+pub fn test_smart_contract(path: &Option<PathBuf>) -> Result<(), Error> {
+	// Execute `cargo test` command in the specified directory.
+	let result = cmd("cargo", vec!["test"])
+		.dir(path.clone().unwrap_or_else(|| PathBuf::from("./")))
+		.run()
+		.map_err(|e| Error::TestCommand(format!("Cargo test command failed: {}", e)))?;
 
-	Ok(())
+	if result.status.success() {
```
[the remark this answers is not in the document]
brunopgalvao: Good point. Removed.

brunopgalvao:
> Nice work! I see here the tests passes in the CI but for me locally the test `test_new_parachain_command_execute` is failing, can you double check is passing for you?

I just ran `cargo test --all-features` and all tests pass.

AlexD10S (approved):

PR #107, opened 2024-04-10
refactor: separate cli frontend with backend logic

AlexD10S:
The goal of this PR is to separate the cli frontend logic with the backend logic, making it easy for external projects to use Pop Functionality for generating, building, deploying and calling using `pop_contracts` crate or generating, building and running parachains and pallets  using `pop_parachains`. 

**Notes for the Reviewer:**
- Only a refactor to separate frontend functionality with backend functionality, no naming refactor has been done unless was needed.
- Commands in the Pop CLI UI that were using some types from external libraries like:
```rs
value: BalanceVariant<<DefaultEnvironment as Environment>::Balance>,
```
Has been replaced with:
```rs
value: String
```
So only crates know what is being used under the hood.

- Same the other way around, the function `into_parachain_template` might seems an absurd method but makes the type Parachain in crates not depended of `clap` library, which is only a frontend one:
```rs
#[derive(Clone, Parser, Debug, Display, EnumString, PartialEq)]
pub enum Template {
	@@ -18,6 +15,15 @@ pub enum Template {
	#[strum(serialize = "Base Parachain Template", serialize = "base")]
	Base,
}
impl Template {
	pub fn into_parachain_template(&self) -> ParachainTemplate {
		match self {
			Template::Base => ParachainTemplate::Base,
			Template::Contracts => ParachainTemplate::Contracts,
			Template::FPT => ParachainTemplate::FPT,
		}
	}
}
```
**How to perform a quick testing:**
To run all the tests: `cargo test --all-features `

To quickly manually test all flows:
_Contracts_
```
cargo run new contract testc -p ../
cargo run build contract -p ../testc  
```
Run substrate cargo contracts:
```
cargo install contracts-node
substrate-contracts-node
```

```
pop up contract -p ../testc  --constructor new --args "false" --suri //Alice
pop call contract -p ../testc --contract 5GpeoZzRq3MumHeEDvZ32oTJQLJiHnSDRWRkggKabZzkUY15 --message get --suri //Alice
pop call contract -p ../testc --contract 5GpeoZzRq3MumHeEDvZ32oTJQLJiHnSDRWRkggKabZzkUY15 --message flip --suri //Alice -x
pop call contract -p ../testc --contract 5GpeoZzRq3MumHeEDvZ32oTJQLJiHnSDRWRkggKabZzkUY15 --message get --suri //Alice
```

_Parachain_:
```
cargo run new parachain my-app -p ../
cargo run build parachain -p ../my-app
[...]

PR #112, opened 2024-04-17
tests: some unit tests for call contracts and up contracts

AlexD10S:
Just some unit tests for the functions call contracts and up contracts.

Will need some e2e tests to test the whole functionality, need running a network first.


crates/pop-contracts/README.md
```diff
@@ -115,4 +115,4 @@ let contract_address = call_smart_contract(call_exec,  weight_limit, url)
 ```
 
 ## Acknowledgements
-`pop-contracts` would not be possible without the awesome crate: [`cargo-contract`](https://github.com/paritytech/cargo-contract).
\ No newline at end of file
+`pop-contracts` would not be possible without the awesome crate: [`cargo-contract`](https://github.com/paritytech/cargo-contract).
```
weezy20, line 118: maybe "`this` awesome crate" as it's a singular mention?

crates/pop-contracts/src/call.rs
```diff
@@ -130,3 +130,98 @@ pub async fn call_smart_contract(
 		display_events.display_events::<DefaultEnvironment>(Verbosity::Default, &token_metadata)?;
 	Ok(output)
 }
+
+#[cfg(feature = "unit_contract")]
```
weezy20, line 134: Feature not defined in crate manifest.

crates/pop-contracts/README.md
AlexD10S: Not sure why it appears in this PR to be honest, it was a fix in another PR (A merge I guess).
Suggested by Daan here: https://github.com/r0gue-io/pop-cli/pull/107#discussion_r1562808571, let's keep it with `the`

crates/pop-contracts/src/call.rs
AlexD10S: True! Well spotted

weezy20: Btw, do we really need this extra feature? If only for the test and not anything else?

AlexD10S: That's a good point, first approach was to just ignore the very slow "unit tests".
In this discussion https://github.com/r0gue-io/pop-cli/pull/90#pullrequestreview-1975791633 Bruno suggested to use this approach instead of ignoring

weezy20: I understand that. But my concern is now these are libaries. We should select and define `default-features` for these crates &exclude these test-only-features. The end user does not have to know about / run the entire test suite.

AlexD10S: Are not excluded by default? if they are not indicated in with default = ["unit_contract", "unit_parachain"] ?

weezy20: Yes you're correct. The `default-features` would rather be an indicator that you don't need these additional features.

brunopgalvao:
Looks like you have some errors in the CI other than that LGTM.

AlexD10S:
> Looks like you have some errors in the CI other than that LGTM.

Test passed now!

brunopgalvao (approved):

PR #113, opened 2024-04-19
fix: clone when user use ssh 

AlexD10S:
A user reported an issue when generating the parachain:
```
~/github/pop
❯ pop new parachain Demo
┌   Pop CLI : Generating "Demo" using Base Parachain Template!
│
Error: authentication required but no callback set; class=Ssh (23); code=Auth (-16)
```

It seem our current code for cloning a repository was throwing an error, fixed using `RepoBuilder`. Based on this example: https://docs.rs/git2/0.13.1/git2/build/struct.RepoBuilder.html#example 

After that fix user was having another issue:
```
~/github/pop
❯ pop new parachain my-app
┌   Pop CLI : Generating "my-app" using Base Parachain Template!

Error: remote rejected authentication: Failed getting response; class=Ssh (23); code=Auth (-16)
```
With seems is an issue authenticating for the SSH clone. Fixed using the library: https://crates.io/crates/git2_credentials

Also I removed `Repository::init(target)?;` from `instantiate_base_template`. After this PR https://github.com/r0gue-io/pop-cli/pull/65 where we where the git init repo was introduced it gets duplicated.

**_How to replicate the issue:_**
This is tricky, I wasn't able to get same issue as the user, but if you change your way to clone, to use SSH you will see an error too in current main:
`git config --global url."git@github.com:".insteadOf "https://github.com/"`
A guide on how to set up SSH cloning: 
https://www.theserverside.com/blog/Coffee-Talk-Java-News-Stories-and-Opinions/github-clone-with-ssh-keys

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,6 +31,7 @@ serde_json = { version = "1.0"}
 serde = { version = "1.0", features = ["derive"] }
 zombienet-sdk = { git = "https://github.com/r0gue-io/zombienet-sdk", branch = "pop", version = "0.1.0-alpha.1"}
 zombienet-support = { git = "https://github.com/r0gue-io/zombienet-sdk", branch = "pop", version = "0.1.0-alpha.1" }
+git2_credentials = "0.13.0"
 
 # pop-cli
 clap = { version = "4.4", features = ["derive"] }
--- a/crates/pop-parachains/src/utils/git.rs
+++ b/crates/pop-parachains/src/utils/git.rs
@@ -1,6 +1,8 @@
 use anyhow::{anyhow, Result};
-use git2::{build::RepoBuilder, FetchOptions, IndexAddOption, Repository, ResetType};
-use git2::{Cred, RemoteCallbacks};
+use git2::{
+	build::RepoBuilder, FetchOptions, IndexAddOption, RemoteCallbacks, Repository, ResetType,
+};
+use git2_credentials::CredentialHandler;
 use regex::Regex;
 use std::path::Path;
 use std::{env, fs};
@@ -30,11 +32,13 @@ impl Git {
 			Template::Contracts => "https://github.com/paritytech/substrate-contracts-node.git",
 			Template::Base => "https://github.com/r0gue-io/base-parachain",
 		};
-		println!("before");
+		println!("Going to clone the repo");
 		let repo = Repository::clone(url, target).unwrap_or(Self::ssh_clone(template, target)?);
 
+		println!("Cloned");
 		// fetch tags from remote
 		let release = Self::fetch_latest_tag(&repo);
+		println!("Release: {:?}", release);
 
 		let git_dir = repo.path();
 		fs::remove_dir_all(&git_dir)?;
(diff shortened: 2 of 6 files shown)


Cargo.toml
```diff
@@ -31,6 +31,7 @@ serde_json = { version = "1.0"}
 serde = { version = "1.0", features = ["derive"] }
 zombienet-sdk = { git = "https://github.com/r0gue-io/zombienet-sdk", branch = "pop", version = "0.1.0-alpha.1"}
 zombienet-support = { git = "https://github.com/r0gue-io/zombienet-sdk", branch = "pop", version = "0.1.0-alpha.1" }
+git2_credentials = "0.13.0"
```
weezy20, line 34: IIRC, git2 itself has methods that expose functionality to SSH clone repositories.

crates/pop-parachains/src/utils/git.rs
```diff
@@ -32,6 +42,30 @@ impl Git {
...
+	/// For users that have ssh configuration for cloning repositories
+	fn ssh_clone(template: &Template, target: &Path) -> Result<Repository> {
+		let ssh_url = match template {
+			Template::FPT => "git@github.com:paritytech/frontier-parachain-template.git",
+			Template::Contracts => "git@github.com:paritytech/substrate-contracts-node.git",
+			Template::Base => "git@github.com:r0gue-io/base-parachain.git",
+		};
+		// Prepare callback and fetch options.
+		let mut callbacks = RemoteCallbacks::new();
+		let git_config = git2::Config::open_default().unwrap();
```
weezy20, line 54: Please handle this error properly. `.context("reason for failure")?`

crates/pop-parachains/src/utils/git.rs
```diff
@@ -32,6 +42,30 @@ impl Git {
 		Ok(release)
 	}
 
+	/// For users that have ssh configuration for cloning repositories
+	fn ssh_clone(template: &Template, target: &Path) -> Result<Repository> {
+		let ssh_url = match template {
+			Template::FPT => "git@github.com:paritytech/frontier-parachain-template.git",
```
weezy20, line 48: nit: Feels like we can codify these URLS into some const.

```rs
const TEMPLATE_FPT: &'static str = "paritytech/frontier-parachain-template.git"
// Then during usage with either HTTPS or SSH
let _ = "git@github.com".concat(TEMPLATE_FPT)
```

weezy20 (changes requested):
Great work @AlexD10S. Only mandating changes in the error handler for `git2:Config`. Rest looks good.

weezy20:
TBF, I still do not understand how to replicate this issue because we are straight up asking `git2` in pop to use `https`. I don't understand the point of this configuration:
```sh
git config --global url."git@github.com:".insteadOf "https://github.com/"
```

AlexD10S:
> TBF, I still do not understand how to replicate this issue because we are straight up asking `git2` in pop to use `https`. I don't understand the point of this configuration:

Some people set up their git configuration to clone the repositories using SSH using `insteadOf`

Before this fixes you can try:
```shell
git config --global url."git@github.com:".insteadOf "https://github.com/"
```

And this will throw an error when try to clone, because even pop has https:: your git will go with the new configuration. See this issue: https://github.com/rust-lang/git2-rs/issues/901

Cargo.toml
weezy20: @AlexD10S thoughts?

brunopgalvao: Is `git2_credentials` needed or could we use this: https://docs.rs/git2/latest/git2/build/struct.RepoBuilder.html#example

weezy20 (approved):

PR #120, opened 2024-04-24
refactor: maximise use of enums

evilrobot-01:
Maximises the use of the `Provider` and `Template` enums so that new templates can be added easily and in a single place.

Note: I think the Parity templates should be marked as deprecated owing to how old the available releases are. We could also look to provide the default branch as an additional option to the available releases, perhaps via an opt-in flag/property by provider or template.


AlexD10S (approved):
It looks very good! Thanks for improving it

PR #121, opened 2024-04-26
Add and export const strs for HTTPS/SSH github url prefixes

weezy20:


AlexD10S:
Oh.. I was working in another approach to remove the hardcoded Github prefix having the methods returning the proper url in the Templates enum

PR #122, opened 2024-04-26
build: enable vendored-openssl via feature to ease cross compilation

evilrobot-01:
Fixes the cross compilation issues. Results of a run at https://github.com/r0gue-io/pop-cli/actions/runs/8853101362, tested briefly on all platforms via Docker/Linux/Rosetta using Mac.

Also removed skylake as dont think we need to make that available.


AlexD10S (approved):

brunopgalvao:
Tested and worked:
- https://github.com/brunopgalvao/pop-cli/releases

PR #123, opened 2024-04-28
feat(up parachain): improve build ux

evilrobot-01:
Improves the ux when building binaries for `pop up parachain`, by updating `cliclack` and by using a callback to pass underlying `cargo build` status updates back up to cli ui for output.

Also adds a retry attempt should the first attempt fail. This is primarily due to an interrupted process, but could potentially be due to a connectivity issue.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1291,9 +1291,9 @@ checksum = "98cc8fbded0c607b7ba9dd60cd98df59af97e84d24e49c8557331cfc26d301ce"
 
 [[package]]
 name = "cliclack"
-version = "0.1.13"
+version = "0.2.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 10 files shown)


AlexD10S (commented):
The UX looks much nicer now. 
Can we add a comment explaining why we this callbacks, this code might be confusing:
`binary.source(&cache, &|_| {}).await?;`

Can we add the same approach in `pop build parachain ` and `pop build contract` ?

evilrobot-01:
> Can we add a comment explaining why we this callbacks, this code might be confusing: `binary.source(&cache, &|_| {}).await?;`

Have improved the types used to better convey usage and intent, along with doc comments for public functions and types.

> Can we add the same approach in `pop build parachain ` and `pop build contract` ?

Can be done fairly easily for others as a separate PR after latest refactor, but would only undertake if the approach is approved.

AlexD10S (approved):

brunopgalvao (approved):

PR #127, opened 2024-05-01
feat: pop install

### LINKED ISSUE #110: pop install
Add a command` pop install <options>` that set up the environment for parachain and smart contract development.

This command will help new developers, detecting their current environment for Rust, Protobuf, Zombienet, and other installing essential tools that are required for development.

weezy20:
Setup substrate development environment. 
Closes #110


crates/pop-cli/src/commands/install/mod.rs
```diff
@@ -0,0 +1,33 @@
...
+		let client = reqwest::Client::new();
+		let script = client
+			.get("https://raw.githubusercontent.com/r0gue-io/pop-cli/pop-install/scripts/get_substrate.sh")
+			.send()
+			.await
+			.context("Network Error: Failed to fetch script from Github")?
+			.text()
+			.await?;
+		fs::write(scripts_path.as_path(), script).await?;
+		Command::new("bash").arg(scripts_path).status().await?;
```
AlexD10S, line 29: We used `duct::cmd` to run cargo test in the test smart contract functionality.

crates/pop-cli/src/commands/install/mod.rs
```diff
@@ -0,0 +1,33 @@
...
+#[derive(Args)]
+#[command(args_conflicts_with_subcommands = true)]
+/// Setup user environment for substrate development
+/// Runs script `scripts/get_substrate.sh`
+pub(crate) struct InstallArgs;
+
+impl InstallArgs {
+	pub(crate) async fn execute(self) -> anyhow::Result<()> {
+		if cfg!(target_os = "windows") {
+			return Ok(cliclack::log::error("Windows is supported only with WSL2")?);
```
AlexD10S, line 16: Should we add a URL for more info? https://docs.substrate.io/install/windows/

crates/pop-cli/src/main.rs
```diff
@@ -36,40 +36,44 @@ enum Commands {
 	#[clap(alias = "t")]
 	#[cfg(feature = "contract")]
 	Test(commands::test::TestArgs),
+	/// Install prerequisites for Substrate development
```
AlexD10S, line 39: I like _Set up the environment for Substrate development_ more. But up to you to decide

AlexD10S (commented):
Code looks good (just few comments), but the script needs more user testing, in a Mac OS (Darwin) I get an error:
`error: failed to set permissions for '/Users/alexbean/.cargo/bin/rustup-init': No such file or directory (os error 2)`

After this PR, next steps has to be have this functionality in Rust code to improve UX: We can have better customization on the user messages (Instead of the echo, see this PR: https://github.com/r0gue-io/pop-cli/pull/123), also prompt the user to pick what they want/don't want to install....

brunopgalvao:
If we are going to make edits to the `get_substrate` script, I advocate we replace it with Rust code (as we were already planning on doing so).

scripts/get_substrate.sh
```diff
@@ -0,0 +1,75 @@
...
+elif [[ "$OSTYPE" == "darwin"* ]]; then
+	set -e
+	echo "Mac OS (Darwin) detected."
+
+	if ! which brew >/dev/null 2>&1; then
+		/bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install.sh)"
+	fi
+
+	brew update
+	brew install openssl cmake llvm
```
peterwht, line 50: needs to include `protobuf`

peterwht (changes requested):
Looks mostly good! A few changes I would like to see for this first version:
- protobuf needs to be included for mac
- there should be a question at the start saying "this will attempt to install the following components: <components>. do you want to proceed?"

Long term, it would be nice to not use a .sh script and have it entirely in code. But let's see how useful this command is first before furhter improving on it :)

PR #131, opened 2024-05-01
feat(cli): integrate assets parachain template

al3mart:
Include Assets parachain template to the collection of templates and its tests.

### DIFF
--- a/crates/pop-parachains/src/templates.rs
+++ b/crates/pop-parachains/src/templates.rs
@@ -79,11 +79,21 @@ pub enum Template {
 		props(Provider = "Pop", Repository = "https://github.com/r0gue-io/base-parachain")
 	)]
 	Base,
+	#[strum(
+		serialize = "nfts",
+		message = "Nfts",
+		detailed_message = "Parachain configured with assets and nfts functionalities.",
+		props(
+			Provider = "Pop",
+			Repository = "https://github.com/r0gue-io/base-parachain/tree/nfts-parachain"
+		)
+	)]
+	Nfts,
 	// Parity
 	#[strum(
 		serialize = "cpt",
 		message = "Contracts",
-		detailed_message = "Minimal Substrate node configured for smart contracts via pallet-contracts.",
+		detailed_message = "Parachain configured with wasm smart contracts (pallet-contracts) support.",
 		props(
 			Provider = "Parity",
 			Repository = "https://github.com/paritytech/substrate-contracts-node"
@@ -93,7 +103,7 @@ pub enum Template {
 	#[strum(
 		serialize = "fpt",
 		message = "EVM",
-		detailed_message = "Template node for a Frontier (EVM) based parachain.",
+		detailed_message = "Parachain configured with EVM (Frontier) support.",
 		props(
 			Provider = "Parity",
 			Repository = "https://github.com/paritytech/frontier-parachain-template"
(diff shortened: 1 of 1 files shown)


crates/pop-parachains/src/templates.rs
```diff
@@ -79,11 +79,21 @@ pub enum Template {
 		props(Provider = "Pop", Repository = "https://github.com/r0gue-io/base-parachain")
 	)]
 	Base,
+	#[strum(
+		serialize = "nfts",
+		message = "Nfts",
+		detailed_message = "Parachain configured with assets and nfts functionalities.",
+		props(
+			Provider = "Pop",
+			Repository = "https://github.com/r0gue-io/base-parachain/tree/nfts-parachain"
```
evilrobot-01, line 88: Suggest a rename to `assets-parachain`

crates/pop-parachains/src/templates.rs
```diff
@@ -79,11 +79,21 @@ pub enum Template {
 		props(Provider = "Pop", Repository = "https://github.com/r0gue-io/base-parachain")
 	)]
 	Base,
+	#[strum(
+		serialize = "nfts",
+		message = "Nfts",
+		detailed_message = "Parachain configured with assets and nfts functionalities.",
```
evilrobot-01, line 85: "Parachain configured with fungible and non-fungilble asset functionalities."

crates/pop-parachains/src/templates.rs
```diff
@@ -79,11 +79,21 @@ pub enum Template {
...
+		props(
+			Provider = "Pop",
+			Repository = "https://github.com/r0gue-io/base-parachain/tree/nfts-parachain"
+		)
+	)]
+	Nfts,
```
evilrobot-01, line 91: `Assets`

evilrobot-01 (changes requested):

crates/pop-parachains/src/templates.rs
```diff
@@ -79,11 +79,21 @@ pub enum Template {
...
 	// Parity
 	#[strum(
 		serialize = "cpt",
 		message = "Contracts",
-		detailed_message = "Minimal Substrate node configured for smart contracts via pallet-contracts.",
+		detailed_message = "Parachain configured with wasm smart contracts (pallet-contracts) support.",
```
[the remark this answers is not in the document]
al3mart: Good, didn't know

crates/pop-parachains/src/templates.rs
```diff
@@ -93,7 +103,7 @@ pub enum Template {
 	#[strum(
 		serialize = "fpt",
 		message = "EVM",
-		detailed_message = "Template node for a Frontier (EVM) based parachain.",
+		detailed_message = "Parachain configured with EVM (Frontier) support.",
```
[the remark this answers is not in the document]
al3mart: Nice, good to know.

crates/pop-parachains/src/templates.rs
evilrobot-01: Also, as we are presenting releases to install via the CLI, it _might_ be better to separate these into different repos, otherwise we will need to add filtering of releases by tag - think about x templates all needing to be released at the same time based on a Polkadot version update. We will also need to set up additional branch protection rules for each of the templated branches.

Using separate repos means we can use the same release tag versioning, so think we should consider future maintenance requirements before merging this.

PR #132, opened 2024-05-01
feat(cli): integrate assets parachain template.

al3mart:
This PR substitutes #131.

Asset parachain template moved to: https://github.com/r0gue-io/assets-parachain

### DIFF
--- a/crates/pop-parachains/src/templates.rs
+++ b/crates/pop-parachains/src/templates.rs
@@ -79,11 +79,21 @@ pub enum Template {
 		props(Provider = "Pop", Repository = "https://github.com/r0gue-io/base-parachain")
 	)]
 	Base,
+	#[strum(
+		serialize = "nfts",
+		message = "Nfts",
+		detailed_message = "Parachain configured with assets and nfts functionalities.",
+		props(
+			Provider = "Pop",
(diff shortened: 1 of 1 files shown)


brunopgalvao (approved):

AlexD10S:
Just a small thing. This is the info when running `pop new parachain --help`:
```
Options:
  -t, --template <TEMPLATE>            Template to use: 'base' for Pop and 'cpt' and 'fpt' for Parity templates
```
Can you add there assets? The code is here: https://github.com/r0gue-io/pop-cli/blob/main/crates/pop-cli/src/commands/new/parachain.rs#L25

evilrobot-01:
> Can you add there assets?

We shouldn't have to update this description with each change like this, it is prone to error. It should be general and should either call a function to generate the list of templates based on available enum variants, or refer to another subcommand to show all accepted values.

Seems that clap has something called `PossibleValue`, so seems that might be the right approach.

evilrobot-01:
Addressed in https://github.com/r0gue-io/pop-cli/pull/133

evilrobot-01 (approved):
Super clean PR, shows how easy it will be to add templates moving forward.

I would wait until https://github.com/r0gue-io/pop-cli/pull/133 is approved and merged before merging this one.

AlexD10S (approved):

PR #136, opened 2024-05-02
feat: add telemetry support

peterwht:
This PR introduces ANONYMOUS telemetry to get metrics on the following things:
- Is Pop CLI being used. This reporting only increases a counter. No personal information reported.
- What commands are used. This, once again, increases a counter with no user inputted information reported. For example, if `pop new parachain my_personal_filename` is inputted, only `pop new parachain` will be reported. 

A new arg has been added to `new parachain`: `release_tag`. This was added due to a refactor that enables calling `generate_parachain_from_template` once, instead of twice in separate flows. The latest release version is used if release_tag is not set. This argument will likely be most useful in a CI.

- [x] Telemetry handler
- [x] Report if CLI is used and what commands are used
- [x] Report generic errors (no user inputted info included)
- [x] Manage opt-in via config
- [x] Provide a way to disable telemetry will be in a new PR. Can be disabled by compiling with the `no-telemetry` feature flag. 
- ~~[ ] Provide explicit explanation of how the telemetry works, what it's for, and what it tracks~~ will be in a new PR https://github.com/r0gue-io/pop-cli/pull/157
- [x] Unit tests


Cargo.toml
```diff
@@ -19,28 +19,32 @@ duct = "0.13"
 git2 = "0.18"
 tempfile = "3.8"
 tokio = { version = "1.0", features = ["macros", "rt-multi-thread"] }
-url = { version = "2.5"}
+url = { version = "2.5" }
```
evilrobot-01, line 22: Can we sort these alphabetically please? Makes it easier to visually scan to find something, especially when on mobile and ctrl-f not as easy.

Rest of the file should be done, but out of scope of this PR

crates/pop-cli/Cargo.toml
```diff
@@ -14,6 +14,8 @@ duct.workspace = true
 tempfile.workspace = true
 url.workspace = true
 tokio.workspace = true
+serde.workspace = true
```
evilrobot-01, line 17: As above, please insert alphabetically.

crates/pop-cli/src/commands/build/contract.rs
```diff
@@ -19,6 +19,12 @@ impl BuildContractCommand {
 	pub(crate) fn execute(&self) -> anyhow::Result<()> {
 		clear_screen()?;
 		intro(format!("{}: Building a contract", style(" Pop CLI ").black().on_magenta()))?;
+
+		tokio::spawn(pop_telemetry::record_cli_command(
```
evilrobot-01, line 23: Could you add imports for `tokio::spawn`, `pop_telemetry::record_cli_command` and `serde_json::json` in main.rs so that all usages can be simplified to a one liner such as `spawn(record_cli_command("build", json!({"contract": ""})));`

Actually, would it not be much easier to push this up the stack to https://github.com/r0gue-io/pop-cli/blob/353161d8b03936f1ca1e36fbb41cd9d942b6f649/crates/pop-cli/src/main.rs#L47 so that all the telemetry is defined in a single place - i.e. in the match arm before call command.execute()?

crates/pop-cli/src/commands/test/contract.rs
```diff
@@ -19,15 +19,27 @@ pub(crate) struct TestContractCommand {
...
 	pub(crate) fn execute(&self) -> anyhow::Result<()> {
 		clear_screen()?;
+
 		if self.features.is_some() && self.features.clone().unwrap().contains("e2e-tests") {
 			intro(format!(
 				"{}: Starting end-to-end tests",
 				style(" Pop CLI ").black().on_magenta()
 			))?;
+
+			tokio::spawn(pop_telemetry::record_cli_command(
```
evilrobot-01, line 29: Guess this might be an exception to the above comment, but think its better to have one exception that needing to touch all the above files.

crates/pop-cli/src/main.rs
```diff
@@ -38,41 +40,126 @@ enum Commands {
 	Test(commands::test::TestArgs),
 }
 
+fn init_config() -> Result<()> {
```
evilrobot-01, line 43: Minor, but would move this after main. One would expect to see the main function first for a binary, just so its quick to see what it is doing and to branch out from there.

crates/pop-cli/src/main.rs
```diff
@@ -38,41 +40,126 @@ enum Commands {
...
+	let mut tel_error_handle: Option<JoinHandle<pop_telemetry::Result<()>>> = None;
+
 	let cli = Cli::parse();
-	match cli.command {
+	let res = match cli.command {
 		Commands::New(args) => Ok(match &args.command {
 			#[cfg(feature = "parachain")]
-			commands::new::NewCommands::Parachain(cmd) => cmd.execute().await?,
+			commands::new::NewCommands::Parachain(cmd) => cmd.execute().await.map_err(|err| {
+				tel_error_handle = Some(tokio::spawn(pop_telemetry::record_cli_command(
```
evilrobot-01, line 76: As above, a lot of repetition twith tokio::, pop_telemetry:: and serde_json::

crates/pop-cli/src/main.rs
```diff
@@ -38,41 +40,126 @@ enum Commands {
...
+	// This await handle is used as the error will not have sufficient time to report before destruction.
+	let mut tel_error_handle: Option<JoinHandle<pop_telemetry::Result<()>>> = None;
+
 	let cli = Cli::parse();
-	match cli.command {
+	let res = match cli.command {
 		Commands::New(args) => Ok(match &args.command {
 			#[cfg(feature = "parachain")]
-			commands::new::NewCommands::Parachain(cmd) => cmd.execute().await?,
+			commands::new::NewCommands::Parachain(cmd) => cmd.execute().await.map_err(|err| {
```
evilrobot-01, line 75: Might be a good opportunity to clean this up by importing the commands to eliminate the commands::* prefixes.

crates/pop-cli/src/main.rs
```diff
@@ -38,41 +40,126 @@ enum Commands {
...
+					serde_json::json!({"new": "parachain"}),
+				)));
+				err
+			})?,
 			#[cfg(feature = "parachain")]
-			commands::new::NewCommands::Pallet(cmd) => cmd.execute().await?,
+			commands::new::NewCommands::Pallet(cmd) => cmd.execute().await.map_err(|err| {
+				tel_error_handle = Some(tokio::spawn(pop_telemetry::record_cli_command(
+					"error",
+					serde_json::json!({"new": "pallet"}),
```
evilrobot-01, line 86: As mentioned elsewhere, moving the command telemetry from the individual files to above cmd.execute() here will allow this json payload to be defined once per match arm, used for both recording the command and the error.

crates/pop-telemetry/Cargo.toml
```diff
@@ -0,0 +1,15 @@
+[package]
+name = "pop-telemetry"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+thiserror.workspace = true
```
evilrobot-01, line 7: Please sort alphabetically, as noted above.

crates/pop-telemetry/src/lib.rs
```diff
@@ -0,0 +1,179 @@
+use reqwest::Client;
+use serde::{Deserialize, Serialize};
+use serde_json::{json, Value};
+use std::fs::{create_dir_all, File};
```
evilrobot-01, line 4: nest usings

crates/pop-telemetry/src/lib.rs
```diff
@@ -0,0 +1,179 @@
...
+			.json(&payload)
+			.send()
+			.await
+			.map_err(TelemetryError::NetworkError)?;
+
+		Ok(())
+	}
+}
+
+pub async fn record_cli_used() -> Result<()> {
```
evilrobot-01, line 91: Public function needs doc comments.

crates/pop-telemetry/src/lib.rs
```diff
@@ -0,0 +1,179 @@
...
+
+	let payload = generate_payload("cli", CARGO_PKG_VERSION, "/", WEBSITE_ID, "", json!({}));
+
+	let res = tel.send_json(payload).await;
+	log::debug!("send_cli_used result: {:?}", res);
+
+	Ok(())
+}
+
+pub async fn record_cli_command(command_name: &str, data: Value) -> Result<()> {
```
evilrobot-01, line 106: Public function needs doc comments.

crates/pop-telemetry/src/lib.rs
```diff
@@ -0,0 +1,179 @@
...
+#[derive(Serialize, Deserialize, Debug)]
+struct OptIn {
+	allow: bool,
+	version: String,
+}
+#[derive(Serialize, Deserialize, Debug)]
+pub struct Config {
+	opt_in: OptIn,
+}
+pub fn config_file_path() -> Result<PathBuf> {
```
evilrobot-01, line 130: Public function needs doc comments.

crates/pop-telemetry/src/lib.rs
```diff
@@ -0,0 +1,179 @@
...
+	let config_path = dirs::config_dir().ok_or(TelemetryError::ConfigFileNotFound)?.join("pop");
+	// Creates pop dir if needed
+	create_dir_all(config_path.as_path()).map_err(|err| TelemetryError::IO(err))?;
+	Ok(config_path.join("config.json"))
+}
+
+pub fn write_default_config() -> Result<Option<PathBuf>> {
```
evilrobot-01, line 137: Public function needs doc comments.

crates/pop-telemetry/src/lib.rs
```diff
@@ -0,0 +1,179 @@
...
+	} else {
+		// if the file already existed, return None
+		return Ok(None);
+	}
+
+	Ok(Some(config_path))
+}
+
+fn generate_payload(
+	hostname: &str,
```
evilrobot-01, line 158: If this is always "cli", why take it as a parameter, just hardcode in the payload as with language

crates/pop-telemetry/src/lib.rs
```diff
@@ -0,0 +1,179 @@
...
+	title: &str,
+	url: &str,
```
evilrobot-01, line 160: Same, just hardcode "/" below

crates/pop-telemetry/src/lib.rs
```diff
@@ -0,0 +1,179 @@
...
+) -> Value {
+	json!({
+		"payload": {
+			"hostname": hostname,
+			"language": "en-US",
+			"referrer": "",
+			"screen": "1920x1080",
+			"title": title,
+			"url": url,
+			"website": website_id,
```
evilrobot-01, line 173: Just use the WEBSITE_ID const and remove the parameter. Same with title and CARGO_PKG_VERSION

evilrobot-01 (commented):
Know it is still a WIP, but left some comments.

peterwht:
Error reporting currently does not work (in most cases) as the `?` operator is causing early exit. Noting here, and will fix in a later commit.

peterwht:
Do not merge until we have proper endpoint setup. This should be done by hardcoding the endpoint in the code instead of using the `env!` variable.

crates/pop-cli/src/main.rs
```diff
@@ -40,39 +45,115 @@ enum Commands {
...
+				tel_data = ("new", "contract", json!("default"));
+
+				cmd.execute().await
+			},
+		},
 		Commands::Build(args) => match &args.command {
 			#[cfg(feature = "parachain")]
-			commands::build::BuildCommands::Parachain(cmd) => cmd.execute(),
+			build::BuildCommands::Parachain(cmd) => {
+				tel_data = ("build", "parachain", json!(""));
```
evilrobot-01, line 100: `Value::Null` for all these?

crates/pop-cli/src/main.rs
```diff
@@ -40,39 +45,115 @@ enum Commands {
...
+					tel_data = ("test", "contract", json!(feature));
+					Ok(())
+				},
+				Err(e) => Err(e),
+			},
 		},
+	};
+
+	// Best effort to send on first try, no action if failure
+	let _ =
```
evilrobot-01, line 146: As above, wrap these two records in `if let Some(t) = maybe_tel` to only spawn if telemetry enabled.

crates/pop-cli/src/main.rs
```diff
@@ -40,39 +45,115 @@ enum Commands {
...
+		record_cli_command(maybe_tel.clone(), tel_data.0, json!({tel_data.1: tel_data.2})).await;
```
evilrobot-01, line 147: What happens if tel_data remains uninitialised? We are then reporting empty data. Easy solution is to wrap the tuple in an Option and only report when some.

crates/pop-cli/src/main.rs
```diff
@@ -83,6 +164,24 @@ fn cache() -> Result<PathBuf> {
 	Ok(cache_path)
 }
 
+fn init_config() -> Result<()> {
+	let path = config_file_path()?;
+	match pop_telemetry::write_default_config(&path) {
+		Ok(written) => {
+			if written {
+				log::info(format!("Initialized config file at {}", &path.to_str().unwrap()))?;
```
evilrobot-01, line 172: Why info? Seems more like a debug level.

crates/pop-cli/src/main.rs
```diff
@@ -40,39 +45,115 @@ enum Commands {
...
+					Ok(())
+				},
+				Err(e) => Err(e),
+			},
 			#[cfg(feature = "parachain")]
-			commands::new::NewCommands::Pallet(cmd) => cmd.execute().await?,
+			new::NewCommands::Pallet(cmd) => {
+				// when there are more pallet selections, this will likely have to move deeper into the stack
+				tel_data = ("new", "pallet", json!("template"));
+
```
evilrobot-01, line 86: Spaces unnecessary imo. Like one sentence paragraphs.

crates/pop-cli/src/main.rs
```diff
@@ -40,39 +45,115 @@ enum Commands {
...
+	// if config file errors set telemetry to None, otherwise Some(tel)
+	let maybe_tel = maybe_config_file.ok().map(|path| Telemetry::new(endpoint.to_string(), path));
+
+	// handle for await not used here as telemetry should complete before any of the commands do.
+	// Sends a generic ping saying the CLI was used
+	spawn(record_cli_used(maybe_tel.clone()));
+
+	// type to represent static telemetry data. I.e., does not contain data dynamically chosen by user
+	// like in pop new parachain.
+	let mut tel_data: (&str, &str, Value) = ("", "", json!(""));
```
evilrobot-01, line 64: As the telemetry data is only being written after command execution, it would be much simpler to just return tel_data from each of the match arms. The success type of the result is currently (), therefore unused and can therefore be easily replaced with tel data. Using a return type ensures that every condition returns telemetry, whereas this current approach could result in some condition simply not updating the telemetry data.

crates/pop-telemetry/src/lib.rs
```diff
@@ -0,0 +1,365 @@
+use reqwest::Client;
+use serde::de::DeserializeOwned;
+use serde::{Deserialize, Serialize};
```
evilrobot-01, line 3: Nest usings

crates/pop-telemetry/src/lib.rs
```diff
@@ -0,0 +1,365 @@
...
+	// This should include the domain and api path (e.g. localhost:3000/api/send)
+	endpoint: String,
+	// Has the user opted-in to anonymous telemetry
+	opt_in: bool,
+	// Reqwest client
+	client: Client,
+}
+
+impl Telemetry {
+	pub fn new(endpoint: String, config_path: PathBuf) -> Self {
```
evilrobot-01, line 46: Undocumented public function

crates/pop-telemetry/src/lib.rs
```diff
@@ -0,0 +1,365 @@
...
+			.map_err(TelemetryError::NetworkError)?;
+
+		Ok(())
+	}
+}
+
+/// Generically reports that the CLI was used to the telemetry endpoint.
+/// There is explicitly no reqwest retries on failure to ensure overhead
+/// stays to a minimum.
+pub async fn record_cli_used(maybe_tel: Option<Telemetry>) -> Result<()> {
```
evilrobot-01, line 89: As above, the paramater should not take an option but only an actual value. Calling functions should be forced to only call this when they have valid telemetry data to record.

crates/pop-telemetry/src/lib.rs
```diff
@@ -0,0 +1,365 @@
...
+}
+
+/// Reports what CLI command was called to telemetry.
+///
+/// parameters:
+/// `command_name`: the name of the command entered (new, up, build, etc)
+/// `data`: the JSON representation of subcommands. This should never include any user inputted
+/// data like a file name.
+pub async fn record_cli_command(
+	maybe_tel: Option<Telemetry>,
```
evilrobot-01, line 107: As above, remove the option, and probably the TelemetryNotSet error as expect it will become unused.

evilrobot-01 (changes requested):
I think there are some improvements which can still be made. The biggest issue is the lack of a feature flag, which I think is a requirement to opt-out when running tests during CI. We dont want to have GH runs reporting telemetry data.

crates/pop-cli/src/main.rs
peterwht: This is a cliclack log. Made more explicit

crates/pop-cli/src/commands/new/parachain.rs
```diff
@@ -32,6 +32,8 @@ pub struct NewParachainCommand {
 		value_parser = crate::enum_variants!(Template)
 	)]
 	pub(crate) template: Option<Template>,
+	#[arg(short = 'v', long, help = "Tag version to use for template")]
+	pub(crate) tag_version: Option<String>,
```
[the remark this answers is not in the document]
peterwht: renamed, justification added.

crates/pop-cli/src/main.rs
evilrobot-01: Should it not then read `"Initialized telemetry config file at ..."` instead?

crates/pop-cli/src/commands/new/parachain.rs
```diff
@@ -43,12 +49,6 @@ pub struct NewParachainCommand {
 		default_value = "1u64 << 60"
 	)]
 	pub(crate) initial_endowment: Option<String>,
-	#[arg(
-		short = 'p',
-		long,
-		help = "Path for the parachain project, [default: current directory]"
-	)]
-	pub(crate) path: Option<PathBuf>,
```
AlexD10S, line 51: Why remove it?
This was introduced by @weezy20  here: https://github.com/r0gue-io/pop-cli/pull/100  to keep consistency with `pop new contracts` where we have the flag path:

crates/pop-telemetry/src/lib.rs
```diff
@@ -0,0 +1,322 @@
...
+use serde_json::{json, Value};
+use std::{
+	fs::{create_dir_all, File},
+	io,
+	io::{Read, Write},
+	path::PathBuf,
+};
+use thiserror::Error;
+
+const WEBSITE_ID: &str = "3da3a7d3-0d51-4f23-a4e0-5e3f7f9442c8";
```
AlexD10S, line 12: Should this be a .env variable too?

AlexD10S (commented):
Looks great!  I just left a couple of comments.
Can't wait to see some metrics!

crates/pop-cli/src/commands/new/parachain.rs
peterwht: As Frank mentioned, the path arg was never used. I recommend we only add it back in if it becomes needed.

brunopgalvao (commented):
Looks like the Github Action check has not completed. The PR LGTM. Let's get the checks to pass.

.github/workflows/build.yml
```diff
@@ -62,7 +62,7 @@ jobs:
           git-user: ${{ env.GITHUB_ACTOR }}
 
       - name: Run integration tests
-        run: cargo test --no-default-features --features contract --test contract
+        run: cargo test --test contract
```
evilrobot-01, line 65: Why was this removed? I explicitly set it this way so that we wouldnt send telemetry from CI runs?

.github/workflows/build.yml
```diff
@@ -75,4 +75,4 @@ jobs:
           git-user: ${{ env.GITHUB_ACTOR }}
 
       - name: Run integration tests
-        run: cargo test --no-default-features --features parachain --test parachain
+        run: cargo test --test parachain
```
evilrobot-01, line 78: As above

crates/pop-cli/Cargo.toml
```diff
@@ -32,12 +35,16 @@ pop-parachains = { path = "../pop-parachains", optional = true }
...
+pop-telemetry = { path = "../pop-telemetry", optional = true }
+
 [dev-dependencies]
 assert_cmd = "2.0.14"
 predicates = "3.1.0"
 
 [features]
-default = ["contract", "parachain"]
+default = ["contract", "parachain", "telemetry"]
+static-ssl = ["git2/vendored-openssl"]
```
evilrobot-01, line 47: Why is this here? It was removed in main.

evilrobot-01 (changes requested):
Seems to be regressions, presumably due to resolvign merged conflicts.

.github/workflows/build.yml
evilrobot-01: Perhaps it needs a comment above each so it doesnt get restored again in the future.

evilrobot-01 (approved):
LGTM!

PR #137, opened 2024-05-02
feat(cli): integrate contracts and evm parachain template.

AlexD10S:


crates/pop-parachains/src/templates.rs
```diff
@@ -170,6 +206,18 @@ mod tests {
 			template.repository_url().unwrap(),
 			"https://github.com/paritytech/frontier-parachain-template"
 		);
+		template = Template::Assets;
```
evilrobot-01, line 209: As above, use a map to reduce the maintenance burden.

crates/pop-parachains/src/new_parachain.rs
```diff
@@ -21,7 +21,7 @@ pub fn instantiate_template_dir(
 ) -> Result<Option<String>> {
 	sanitize(target)?;
 
-	if matches!(template, &Template::Base) {
+	if template.matches(&Provider::Pop) {
```
evilrobot-01, line 24: This seems incorrect?

crates/pop-parachains/src/templates.rs
```diff
@@ -79,10 +79,31 @@ pub enum Template {
 		props(Provider = "Pop", Repository = "https://github.com/r0gue-io/base-parachain")
 	)]
 	Base,
```
evilrobot-01, line 81: I think we should consider renaming `Base` here to `Standard`. Creating a base parachain sounds weird. Creating a standard parachain sounds rational. As we are still alpha, it is an opportunity to correct this.

It can still pull from the base-parachain repo, its just the UX that could be better.

crates/pop-parachains/src/templates.rs
```diff
@@ -79,10 +79,31 @@ pub enum Template {
...
+	#[strum(
+		serialize = "contracts",
+		message = "Contracts",
+		detailed_message = "Parachain configured to supports Wasm-based contracts.",
+		props(Provider = "Pop", Repository = "https://github.com/r0gue-io/contracts-parachain")
+	)]
+	Contracts,
```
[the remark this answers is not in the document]
AlexD10S: Yes sorted. My logic was to have Contracts the second as will be probably be the most used, but better don't make this assumptions

crates/pop-parachains/src/templates.rs
weezy20: Vanilla Parachain or Default Parachain are ideal candidates as well. But Standard is something I think we can all agree on.

crates/pop-parachains/src/templates.rs
```diff
@@ -143,12 +164,27 @@ mod tests {
 		template = Template::ParityFPT;
 		assert_eq!(template.matches(&Provider::Pop), false);
 		assert_eq!(template.matches(&Provider::Parity), true);
+
```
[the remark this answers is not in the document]
AlexD10S: Nice!

crates/pop-parachains/src/new_parachain.rs
AlexD10S: No, for external parachains we only pull the repo from Github. For Pop ones we instantiate the parachain (customization flags, generate zombienet config...).
Before was hardcoded only for Base parachain, with this change will take all Pop templates

crates/pop-parachains/src/templates.rs
AlexD10S: Refactored to Standard as the three of us seems to agree. Default sound good to me too.

crates/pop-parachains/src/new_parachain.rs
```diff
@@ -21,21 +21,21 @@ pub fn instantiate_template_dir(
 ) -> Result<Option<String>> {
 	sanitize(target)?;
 
-	if matches!(template, &Template::Base) {
-		return instantiate_base_template(target, config, tag_version);
+	if template.matches(&Provider::Pop) {
+		return instantiate_base_template(template, target, config, tag_version);
```
weezy20, line 25: Can be renamed to `instantiate_standard_template` in line with prior changes.

crates/pop-parachains/src/templates.rs
```diff
@@ -141,64 +155,75 @@ pub enum Error {
...
+			("assets".to_string(), Template::Assets),
+			("contracts".to_string(), Template::Contracts),
+			("evm".to_string(), Template::EVM),
+			("cpt".to_string(), Template::ParityContracts),
+			("fpt".to_string(), Template::ParityFPT),
+		])
+	}
+	fn templates_urls() -> HashMap<String, &'static str> {
+		HashMap::from([
+			("standard".to_string(), "https://github.com/r0gue-io/base-parachain"),
```
weezy20, line 172: Feels like `base-parachain` should be renamed as well; Not related to the PR but should be addressed on that repo.

weezy20 (approved):

crates/pop-parachains/src/new_parachain.rs
```diff
@@ -81,7 +81,7 @@ mod tests {
 			initial_endowment: "1000000".to_string(),
 		};
 		let result: anyhow::Result<Option<String>> =
-			instantiate_base_template(temp_dir.path(), config, None);
+			instantiate_base_template(&Template::Standard, temp_dir.path(), config, None);
```
brunopgalvao, line 84: `instantiate_standard_template`

crates/pop-parachains/src/templates.rs
AlexD10S: For this suggestion is better you open an issue in the base-parachain repo.

README.md
```diff
@@ -30,6 +30,15 @@ If no guidance is needed, proceed with:
 pop new parachain my-app
 ```
 
+`pop-cli` supports diverse project templates, to use a specific one use the flag `--template`:
+```sh
+# Create a contracts parachain
```
evilrobot-01, line 35: Can we retain the same order (alphabetically) please?

brunopgalvao (approved):

brunopgalvao (approved):

Issue #138, opened 2024-05-03
Error running `pop up parachain -f ./network.toml`:  error decoding response body: invalid type: map


chungquantin:
My project is created using the command `pop new parachain my-app`, after running the command to deploy the local relaychain `pop up parachain -f ./network.toml`, I got the below error:
```
┌   Pop CLI : Deploy a parachain
│
Error: error decoding response body: invalid type: map, expected a sequence at line 1 column 0

Caused by:
    invalid type: map, expected a sequence at line 1 column 0
```

AlexD10S:
Hi! First of all thanks for reporting issues and helping us to improve pop-cli.

I am trying to replicate your issue, but with the flow you describe is working for me.

- I assume your `network.toml` is the one generated by pop-cli, but can you share your `network.toml`? so I can see if there is something wrong in the generation of it.

- Also I'd appreciate more information with about system and in what commit or branch you installed `pop-cli`

chungquantin:
@AlexD10S Thank you for our response. Below is my `network.toml`.

<img width="568" alt="Screenshot 2024-05-04 at 20 14 43" src="https://github.com/r0gue-io/pop-cli/assets/56880684/7e99fabd-6ca5-42d5-8db4-b3a75b24c2a1">

Actually, I added a few other pallets to my runtime and it builds successfully. I expect the error with `pop up parachain` won't do anything with the runtime. 

My pop cli version is `pop-cli 0.1.0`

AlexD10S:
The `network.toml` is ok, so if you added some pallets and builds successfully `pop up parachain` should work fine.
I have been researching the error `invalid type: map, expected a sequence at line 1 column 0` and it looks it might be an error when parsing some external data ([StackOverflow answer](https://stackoverflow.com/questions/48595735/invalid-type-map-expected-a-sequence-when-deserializing-a-nested-json-struct )).

I have started a new branch in the repo to try to replicate your issue and fix the problem. Can you test the app using the branch `fix/error-parsing`?
You can install it with:
```
cargo install --locked --git https://github.com/r0gue-io/pop-cli --branch fix/error-parsing --force
```
It might be an issue when fetching the Polkadot releases from Github:
https://api.github.com/repos/paritytech/polkadot-sdk/releases

chungquantin:
@AlexD10S Thank you. It is fixed!

AlexD10S:
[...]

PR #142, opened 2024-05-07
fix: error parsing polkadot version from github API

AlexD10S:
Close: https://github.com/r0gue-io/pop-cli/issues/138

In `pop up parachain` we are getting the last Polkadot release from the GitHub API. If there is something wrong with this call (for example:  IP blocked by Github due to rate limit), return at least a default version

### DIFF
--- a/crates/pop-parachains/src/up.rs
+++ b/crates/pop-parachains/src/up.rs
@@ -368,12 +368,10 @@ impl Zombienet {
 
 	async fn latest_polkadot_release() -> Result<String, Error> {
 		let repo = Url::parse(POLKADOT_SDK).expect("repository url valid");
-		println!("{:?}", repo);
 		match GitHub::get_latest_releases(&repo).await {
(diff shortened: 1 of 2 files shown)


evilrobot-01:
What happens if user already has latest version installed, but is then dropped down to an older default version? It will then attempt to clone and build, potentially failing on the clone for the same reason.

AlexD10S:
> What happens if user already has latest version installed, but is then dropped down to an older default version? It will then attempt to clone and build, potentially failing on the clone for the same reason.

If user has latest version and the default version is an older one it will download and build the older one. Even if not installing the latest one, at least we are preventing to crash like in: https://github.com/r0gue-io/pop-cli/issues/138
And for an advanced user that knows what version is the last one, can still specify the version with the flag `--relay_chain`

About failing the clone the same limit as the API?

brunopgalvao (approved):

PR #149, opened 2024-05-08
feat: pop install

### LINKED ISSUE #110: pop install

AlexD10S:
`pop install` command to set up the environment for development by installing required packages.

Closes #110

Next steps can be:
- Improve the UX installation process with less messaging like in here: https://github.com/r0gue-io/pop-cli/pull/123
- Ideal to only install those required based on feature: contracts/parachain.


crates/pop-cli/src/commands/install/mod.rs
```diff
@@ -2,32 +2,228 @@
...
+	cmd("brew", vec!["install", "protobuf", "openssl", "cmake"]).run()?;
+	install_rustup().await?;
+
+	log::success("✅ Installation complete.");
+	Ok(())
+}
+
+async fn install_arch() -> anyhow::Result<()> {
+	log::info("More information about the packages to be installed here: https://docs.substrate.io/install/linux/")?;
+	if !confirm("📦 Do you want to proceed with the installation of the following packages:cmake gcc openssl-1.0 pkgconf git clang?")
```
brunopgalvao, line 68: I like how you prompt the user to confirm.

.github/workflows/install.yml
```diff
@@ -20,8 +20,33 @@ jobs:
         run: apt-get update && apt-get -y install build-essential cmake curl git
       - name: Install Rust
         run: curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
-      - name: Install Pop
+      - name: Pop Install
```
evilrobot-01, line 23: I think it would be better to separate this into two steps, so that it is easier to parse the output of the run. Suggestion is "Install Pop" as before, without the 'pop --version' and then a separate step for "Pop Install".

See the current workflow output as an example.

.github/workflows/install.yml
```diff
@@ -20,8 +20,33 @@ jobs:
...
         run: |
           . "$HOME/.cargo/env"
           cargo install --locked --path ./crates/pop-cli
-          pop --version
\ No newline at end of file
+          pop install -y
+  debian:
```
evilrobot-01, line 28: Please sort the jobs alphabetically so it easier to maintain long term

.github/workflows/install.yml
AlexD10S: Done!

.github/workflows/install.yml
```diff
@@ -11,6 +11,36 @@ defaults:
...
+          cargo install --locked --path ./crates/pop-cli
+      - name: Run Pop Install
+        run: |
+          . "$HOME/.cargo/env"
+          pop install -y
+  macos:
+    runs-on: macos-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Install Rust
```
evilrobot-01, line 35: I think we should `brew update && brew install cmake openssl protobuf` as per https://learn.onpop.io/pop/v/pop-cli/welcome/installing-pop-cli/macos so all consistent and so that we have checks in case this stops working.

crates/pop-cli/src/commands/install/mod.rs
```diff
@@ -2,32 +2,234 @@
...
 use tokio::{fs, process::Command};
 
 #[derive(Args)]
 #[command(args_conflicts_with_subcommands = true)]
-/// Setup user environment for substrate development
-/// Runs script `scripts/get_substrate.sh`
-pub(crate) struct InstallArgs;
+/// Setup user environment for development
+pub(crate) struct InstallArgs {
+	/// Before install all the dependencies needed, do not ask the user for confirmation.
```
evilrobot-01, line 14: "Automatically install all dependencies required without prompting for confirmation."

crates/pop-cli/src/commands/install/mod.rs
```diff
@@ -2,32 +2,234 @@
...
+	#[clap(short('y'), long)]
+	skip_confirm: bool,
+}
 
 impl InstallArgs {
 	pub(crate) async fn execute(self) -> anyhow::Result<()> {
-		if cfg!(target_os = "windows") {
-			return Ok(cliclack::log::error("Windows is supported only with WSL2")?);
+		if cfg!(target_os = "macos") {
```
evilrobot-01, line 21: Missing https://github.com/r0gue-io/pop-cli/blob/ccfc12322d8d7da61c81d87568854517ddcedcf7/crates/pop-cli/src/commands/new/parachain.rs#L71-L72 and https://github.com/r0gue-io/pop-cli/blob/ccfc12322d8d7da61c81d87568854517ddcedcf7/crates/pop-cli/src/commands/new/parachain.rs#L99

.github/workflows/install.yml
```diff
@@ -11,6 +11,36 @@ defaults:
     shell: bash
 
 jobs:
+  debian:
```
evilrobot-01, line 14: Missing Arch using https://hub.docker.com/_/archlinux image as the container tag

crates/pop-cli/src/commands/install/mod.rs
```diff
@@ -2,32 +2,234 @@
...
+		install_rustup().await?;
+		log::success("✅ Installation complete.")?;
 		Ok(())
 	}
 }
+
+async fn install_mac(skip_confirm: bool) -> anyhow::Result<()> {
+	log::info("More information about the packages to be installed here: https://docs.substrate.io/install/macos/")?;
+	if !skip_confirm {
+		prompt_for_confirmation("Homebrew, protobuf, openssl, rustup, and cmake")?
```
evilrobot-01, line 56: Fine for now, but these static strings are error prone when updating n the future, easy to forget something when all scattered about and then things get out of sync and quality degrades.

.github/workflows/install.yml
```diff
@@ -11,6 +11,36 @@ defaults:
...
+      - name: Install Rust
+        run: curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
+      - name: Install Pop
+        run: |
+          . "$HOME/.cargo/env"
+          cargo install --locked --path ./crates/pop-cli
+      - name: Run Pop Install
+        run:
+          pop install -y
   ubuntu:
```
evilrobot-01, line 44: Missing redhat in line with supported types below

crates/pop-cli/src/commands/install/mod.rs
```diff
@@ -2,32 +2,234 @@
...
+				vec![
+					"~/.cargo/env
+",
+				],
+			)
+			.run()?;
+			cmd("rustup", vec!["default", "stable"]).run()?;
+		},
+	}
+	cmd("rustup", vec!["update", "nightly"]).run()?;
```
evilrobot-01, line 200: Note that the toolchain files should sort all this automatically on first build of a project, provided one exists. Fine to leave as is tho.

crates/pop-cli/src/main.rs
```diff
@@ -44,7 +44,7 @@ enum Commands {
 	#[clap(alias = "t")]
 	#[cfg(feature = "contract")]
 	Test(commands::test::TestArgs),
-	/// Install prerequisites for Substrate development
+	/// Set up the environment for development
```
evilrobot-01, line 47: "Set up the environment for development by installing required packages."

evilrobot-01 (commented):
Great work! I think there are some missing items before this can be merged tho, just so that it is a little more robust in terms of QA and also has a consistent UX with the other commands, as noted.

Note that I do not use brew on mac so have not been able to test it manually there myself, but I did do so in a local Ubuntu container.

crates/pop-cli/src/commands/install/mod.rs
AlexD10S: Good point

crates/pop-cli/src/commands/install/mod.rs
```diff
@@ -2,32 +2,234 @@
...
+
+async fn install_rustup() -> anyhow::Result<()> {
+	match cmd("which", vec!["rustup"]).read() {
+		Ok(output) => {
+			log::info(format!("ℹ️ rustup installed already at {}.", output))?;
+			cmd("rustup", vec!["update"]).run()?;
+			cmd("rustup", vec!["default", "stable"]).run()?;
+		},
+		Err(_) => {
+			let mut spinner = cliclack::spinner();
```
Daanvdplas, line 185:
```suggestion
			let spinner = cliclack::spinner();
```

Daanvdplas (commented):
Removed homebrew from machine and tried to build pop cli. Unfortunately I needed `cmake` which also required me to install `homebrew`. After installing those I could build `pop` and run `pop install`.

The error message regarding `cmake`:
Screenshot 2024-05-09 at 23 10 35

We should accommodate for homebrew and cmake in the readme

crates/pop-cli/src/commands/install/mod.rs
AlexD10S: Done, good to follow good practices since the beginning!

AlexD10S:
> 
> We should accommodate for homebrew and cmake in the readme

This is an error to build pop. I have added a link to the documentation in the README for more information: https://learn.onpop.io/pop/v/pop-cli/welcome/installing-pop-cli

crates/pop-cli/src/commands/install/mod.rs
```diff
@@ -1,33 +1,344 @@
...
 use anyhow::Context;
 use clap::Args;
+use cliclack::{clear_screen, confirm, intro, log, outro, set_theme};
+use duct::cmd;
+use os_info::Type;
+use strum::Display;
 use tokio::{fs, process::Command};
 
+#[derive(Display)]
+pub enum DEPENDENCIES {
```
evilrobot-01, line 12: Note, only constants should be in upper case.

AlexD10S: Mistake!

.github/workflows/install.yml
AlexD10S: Fixed. CI runs succesfully here: https://github.com/r0gue-io/pop-cli/actions/runs/9029075945

.github/workflows/install.yml
AlexD10S: Fixed, CI: https://github.com/r0gue-io/pop-cli/actions/runs/9029075945

crates/pop-cli/src/commands/install/mod.rs
```diff
@@ -141,13 +250,13 @@ async fn install_redhat(skip_confirm: bool) -> anyhow::Result<()> {
...
-			"protobuf-compiler",
-			"clang",
-			"clang-devel",
+			&DEPENDENCIES::Cmake.to_string(),
+			&DEPENDENCIES::OpenSslDevel.to_string(),
+			&DEPENDENCIES::Git.to_string(),
+			&DEPENDENCIES::Protobuf.to_string(),
+			&DEPENDENCIES::ProtobufCompiler.to_string(),
+			&DEPENDENCIES::Clang.to_string(),
+			&DEPENDENCIES::ClangDevel.to_string(),
```
Daanvdplas: Here and the above no rustup required?

AlexD10S: Yes. Rustup is installed in all environments. Is in the main function: https://github.com/r0gue-io/pop-cli/blob/alexd10s/pop-install/crates/pop-cli/src/commands/install/mod.rs#L92

crates/pop-cli/src/commands/install/mod.rs
```diff
@@ -0,0 +1,338 @@
...
+			"{}, {}, {}, {}, {} and {}",
+			Dependencies::Curl,
+			Dependencies::Git,
+			Dependencies::Clang,
+			Dependencies::Make,
+			Dependencies::Openssl,
+			Dependencies::Rustup,
+		))?
+	}
+	cmd("pacman", vec!["-Syu", "--needed", "--noconfirm", &Dependencies::Openssl.to_string()])
```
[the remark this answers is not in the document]
AlexD10S: Yes. It took me a while to figure out why the CI was failing in `arch` and is because it was missing it.
For the `openSSL` is need in `macos` and `arch`, but for example in `debian` ad `ubuntu` we are installing` libssl-dev`.

You can see the whole list here: https://stackoverflow.com/a/77978063/5964038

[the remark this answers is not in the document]
evilrobot-01: Can this not just be added with the dependencies on lines 147-150? If it needs to be explicitly installed first, I suggest adding it above Curl first (along with a comment as to why) and if it is explicitly needed to be run as a separate command then it should have a comment noting why.

crates/pop-cli/src/commands/install/mod.rs
```diff
@@ -0,0 +1,338 @@
...
+		Ok(())
+	}
+}
+
+async fn install_mac(skip_confirm: bool) -> anyhow::Result<()> {
+	log::info("More information about the packages to be installed here: https://docs.substrate.io/install/macos/")?;
+	if !skip_confirm {
+		prompt_for_confirmation(&format!(
+			"{}, {}, {}, {} and {}",
+			Dependencies::Homebrew,
```
evilrobot-01, line 103: Can you not use `use Dependencies::*` at top of file to avoid this prefix constantly?

crates/pop-cli/src/commands/install/mod.rs
```diff
@@ -0,0 +1,338 @@
...
+	.run()?;
+
+	Ok(())
+}
+async fn install_ubuntu(skip_confirm: bool) -> anyhow::Result<()> {
+	log::info("More information about the packages to be installed here: https://docs.substrate.io/install/linux/")?;
+	if !skip_confirm {
+		prompt_for_confirmation(&format!(
+			"{}, {}, {}, {}, {} and {}",
+			Dependencies::Git,
```
evilrobot-01, line 162: Fine for now, but in future you could define all the required items in a const array at the top of the function and then just use them as a set when required.

evilrobot-01 (commented):
A few last minor comments.

crates/pop-cli/src/commands/install/mod.rs
AlexD10S: I considered it, but the legibility of the code seemed worse to me, for example:
```
const MACOS_PACKAGES: [Dependencies; 5] = [Homebrew, Protobuf, Openssl, Rustup, Cmake];
cmd("brew",vec!["install",&MACOS_PACKAGES[1].to_string(),&MACOS_PACKAGES[2].to_string(),&MACOS_PACKAGES[4].to_string(),],).run()?;
```

evilrobot-01: I meant something simpler like (not tested): 
`cmd("brew",vec!["install"].concat(PACKAGES.iter().map(|d| d.to_string()).run()?;`

You also dont need to prefix it with `MACOS_` if the const is defined within the function.

AlexD10S: Doesn't look bad, However we will need to mantain a few constants. For example, for MAC:
```
const PACKAGES: [Dependencies; 5] = [Homebrew, Protobuf, Openssl, Rustup, Cmake];
prompt_for_confirmation(&format!("{}", PACKAGES.join(""))
```
but to install with `brew` only need 3 of them, so we will need: 
```
const PACKAGES_BREW_INSTALL: [Dependencies; 3] = [Protobuf, OpenssL,, Cmake];
cmd("brew",vec!["install"].concat(PACKAGES.iter().map(|d| d.to_string()).run()?;
```

evilrobot-01: Constants dont have to be defined at the top of the module, they can be defined within the relevant function. It is simply so that you define the list of packages once, then for the confirmation and the install you use that same set of packages. Then if you need to add/remove a package, you just add it to the const array and job done.

evilrobot-01: Just .iter().skip(2)

evilrobot-01: As mentioned, fine to ignore for now. Its just easiest to define things once and use them in many places to save having to update stuff in multiple places every time there might be a change.

AlexD10S: Thanks for the feedback. I was just trying to fully understand it. I'll leave things as they are for now, but I'll consider this refactor with proper testing.

crates/pop-cli/src/commands/install/mod.rs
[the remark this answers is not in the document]
AlexD10S: Yes, it can. Merged with the rest of the installs.

crates/pop-cli/src/commands/install/mod.rs
```diff
@@ -0,0 +1,313 @@
...
+	#[strum(serialize = "pkg-config")]
+	PkgConfig,
+	#[strum(serialize = "protobuf")]
+	Protobuf,
+	#[strum(serialize = "protobuf-compiler")]
+	ProtobufCompiler,
+	#[strum(serialize = "rustup")]
+	Rustup,
+}
+const MACOS_PACKAGES: [Dependencies; 5] = [Homebrew, Protobuf, Openssl, Rustup, Cmake];
```
evilrobot-01, line 49: I thought we werent using this?

AlexD10S: I apologize, this was not meant to be pushed.

crates/pop-cli/src/commands/install/mod.rs
```diff
@@ -0,0 +1,312 @@
...
+		Err(_) => {
+			let spinner = cliclack::spinner();
+			spinner.start("Installing rustup ...");
+			run_external_script("https://sh.rustup.rs").await?;
+			outro("rustup installed!")?;
+			cmd(
+				"source",
+				vec![
+					"~/.cargo/env
+",
```
evilrobot-01, line 270: Stray newline, weird formatting?

evilrobot-01 (approved):
LGTM! 🚀

PR #152, opened 2024-05-08
fix: update para id

evilrobot-01:
Should be outside of the system parachain range.

### DIFF
--- a/crates/pop-parachains/templates/base/network.templ
+++ b/crates/pop-parachains/templates/base/network.templ
@@ -10,7 +10,7 @@ name = "bob"
 validator = true
 
 [[parachains]]
-id = 1000
+id = 2000
 default_command = "./target/release/^^node^^"
 
 [[parachains.collators]]


AlexD10S:
I see in base-parachain we have the `para_id: 1000`: https://github.com/r0gue-io/base-parachain/blob/main/node/src/chain_spec.rs#L76
We should change it there too.

evilrobot-01:
> I see in base-parachain we have the `para_id: 1000`: https://github.com/r0gue-io/base-parachain/blob/main/node/src/chain_spec.rs#L76 We should change it there too.

Yeah, that comes from the cumulus template: https://github.com/paritytech/polkadot-sdk/blob/6fdb522ded3813f43a539964af78d5fc6d9f1e97/templates/parachain/node/src/chain_spec.rs#L76. 

Zombienet effectively replaces whatever value is set in chain_spec.rs with that specified in network config file. It only starts to matter when doing xcm stuff, just thought it would be worth correcting it in the config file prior to first release. I'll open a PR there to keep it consistent.

AlexD10S (approved):

PR #155, opened 2024-05-10
build(deps): bump zombienet-sdk

evilrobot-01:
Bumps zombienet-sdk dependency, ensure that `args` works when specified in network config.


AlexD10S (approved):
Works now!

PR #158, opened 2024-05-11
ci: add code coverage

evilrobot-01:
Adds code coverage checks.

Coverage report from this branch at https://app.codecov.io/gh/r0gue-io/pop-cli/tree/frank%2Fcode-coverage.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,4 +1,4 @@
-name: Build & Test Pop-CLI
+name: ci
 
 on:
   push:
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -1,4 +1,4 @@
-name: Build & Test Pop-CLI
+name: ci
```
AlexD10S, line 1: Why this change?

evilrobot-01: Better naming for all the CI checks, which are slowly growing to be more than just "build and test". It also keeps the name short, so when viewed in the checks list below it is easier to parse.

AlexD10S (approved):

PR #162, opened 2024-05-12
feat(telemetry): support DO_NOT_TRACK telemetry

peterwht:


evilrobot-01:
Can we do the same with the standardised CI environment variable? I believe it is simply something like CI=true but please kindly check.

peterwht:
> Can we do the same with the standardised CI environment variable? I believe it is simply something like CI=true but please kindly check.

Great idea. CI=true is correct. Will add in

peterwht:
Unit test CI is likely failing due to CI=true in GitHub which now disables telemetry.

PR #168, opened 2024-05-14
fix: licenses

AlexD10S:
- Added symbolik link in the crates to point into the root `LICENSE` file.
- Fixed a `Cargo.toml` license.


peterwht (approved):

evilrobot-01:
Note that I added deny checks to ci in #170 - debatable whether we want this added due to the noise however. Just mentioning it due to the `.gitignore`.

evilrobot-01 (approved):

PR #170, opened 2024-05-14
ci: add dependency and license checks

evilrobot-01:
Adds cargo-deny to CI workflow for dependency and license checks.

Note: the list of allowed licenses have not yet been validated in terms of compliance. This should be completed before this PR is accepted.

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -12,10 +12,15 @@ allow = [
     "ISC",
     "GPL-3.0",
     "MIT",
-    "OpenSSL",
-    "Unicode-DFS-2016"
+    "MPL-2.0",
+    "Unicode-DFS-2016",
+    "Unlicense"
 ]
 confidence-threshold = 0.93
+
+[[licenses.exceptions]]
+allow = ["OpenSSL"]
+name = "ring"
 
 [[licenses.clarify]]
 name = "ring"
--- a/deny.toml
+++ b/deny.toml
@@ -1,6 +1,14 @@
 [graph]
 all-features = true
 
+# This section is considered when running `cargo deny check advisories`
+[advisories]
+unmaintained = "warn"
+ignore = [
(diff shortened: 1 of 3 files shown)


deny.toml
```diff
@@ -0,0 +1,25 @@
+[graph]
+all-features = true
+
+[licenses]
+allow = [
```
AlexD10S, line 5: Comparing your list from [this resource](https://www.gnu.org/licenses/license-list.html#GPLCompatibleLicenses), it looks good to me.
Based on the list we can add more, for example:
` ` ` 
"MPL-2.0",
"Unlicense",
` ` ` 
but might be is not needed for now

deny.toml
```diff
@@ -0,0 +1,25 @@
...
+    "Apache-2.0",
+    "Apache-2.0 WITH LLVM-exception",
+    "BSL-1.0",
+    "BSD-2-Clause",
+    "BSD-3-Clause",
+    "CC0-1.0",
+    "ISC",
+    "GPL-3.0",
+    "MIT",
+    "OpenSSL",
```
AlexD10S, line 15: It looks better to me to to remove OpenSSL from here ([resource](https://www.gnu.org/licenses/license-list.html#OpenSSL))

And add these lines at the bottom:
```rs
[[licenses.exceptions]]
allow = ["OpenSSL"]
name = "ring"
```

evilrobot-01: Yup, much better, thanks!

AlexD10S (approved):

brunopgalvao:
Maybe:
```
[advisories]
unmaintained = "warn"
```

AlexD10S:
> Maybe:
> 
> ```
> [advisories]
> unmaintained = "warn"
> ```

Yes, good point.Adding this with this PR: https://github.com/r0gue-io/pop-cli/pull/174/files

AlexD10S (approved):

Issue #102, opened 2024-04-08
feature: pop build contract debug


Daanvdplas:
I had to build my contract in debug mode however there is no way to specify that

PR #177, opened 2024-05-16
feat: check if build exists before deploying contract with pop up

### LINKED ISSUE #34: `pop up contract`: build contract if it has not been built
Based on this PR comment: https://github.com/r0gue-io/pop-cli/pull/27#discussion_r1515973155

For `pop up contract`, we should do a check, if the contract has not been built, then build it.

bolajahmad:
Closes #34 

`pop up contract` tries to deploy contract even if there is not build found and will eventually throw error because of that. It is necessary to add a check for existing build (and generate one), before continuing with the deployment.


crates/pop-cli/src/commands/up/contract.rs
```diff
@@ -54,7 +54,23 @@ pub struct UpContractCommand {
...
 		clear_screen()?;
+
+		// Check if build exists in the specified "Contract build folder"
+		let build_path = PathBuf::from(
+			self.path.clone().unwrap_or("/.".into()).to_string_lossy().to_string()
+				+ "/target/release/ink",
+		);
+
+		if !build_path.exists() {
+			intro(format!("{}: Building a contract", style(" Pop CLI ").black().on_magenta()))?;
```
evilrobot-01, line 65: A `warning("NOTE: contract has not yet been built")` here before building would be nice. :)

bolajahmad: Oh, sounds good.

I don't understand the intent, should it be used alongside the current intro ("building a contract") or replace this altogether?

AlexD10S:
If I don't have the contract build and run b`pop up contract` works perfectly.
However if I manually build the contract and then run `pop up contract` is still building it, for some reason it doesn't find the repo `!build_path.exists()`.

crates/pop-cli/src/commands/up/contract.rs
bolajahmad: Alright, that makes sense!

AlexD10S:
> If I don't have the contract build and run b`pop up contract` works perfectly. However if I manually build the contract and then run `pop up contract` is still building it, for some reason it doesn't find the repo `!build_path.exists()`.

The issue here is with this line:
```
let build_path = PathBuf::from(
			self.path.clone().unwrap_or("/.".into()).to_string_lossy().to_string()
				+ "/target/release/ink",
		);
```
`release` is not needed, check only for `target/ink`.
Also once this PR is implemented: https://github.com/r0gue-io/pop-cli/issues/102 and the user can specify if build in release mode or debug mode we will have to change it too.

bolajahmad:
> > If I don't have the contract build and run b`pop up contract` works perfectly. However if I manually build the contract and then run `pop up contract` is still building it, for some reason it doesn't find the repo `!build_path.exists()`.
> 
> The issue here is with this line:
> 
> ```
> let build_path = PathBuf::from(
> 			self.path.clone().unwrap_or("/.".into()).to_string_lossy().to_string()
> 				+ "/target/release/ink",
> 		);
> ```
> 
> `release` is not needed, check only for `target/ink`. Also once this PR is implemented: #102 and the user can specify if build in release mode or debug mode we will have to change it too.

Alright, I'll try to remove the target/ink and see if that helps.

PR #180, opened 2024-05-16
feat: add ability to run a script once network is initialized

brunopgalvao:
Simple test:
```
cargo run up parachain -f ./tests/zombienet.toml -p https://github.com/r0gue-io/pop-node --cmd "echo Hello from Pop CLI"
```


crates/pop-cli/src/commands/up/parachain.rs
```diff
@@ -26,6 +27,12 @@ pub(crate) struct ZombienetCommand {
 	/// The url of the git repository of a parachain to be used, with branch/release tag specified as #fragment (e.g. 'https://github.com/org/repository#tag'). A specific binary name can also be optionally specified via query string parameter (e.g. 'https://github.com/org/repository?binaryname#tag'), defaulting to the name of the repository when not specified.
 	#[arg(short, long)]
 	parachain: Option<Vec<String>>,
+	/// The path to a script to be run after the network has been set up.
+	#[clap(name = "init", short = 'i', long)]
+	setup_script: Option<String>,
```
evilrobot-01, line 32: Something more generic like `command` would be better as it could be anything added here instead of a script - e.g. `echo "hi!"`

The separate args can then be removed so it becomes `-i "echo hi!"`

tests/zombienet.toml
```diff
@@ -3,6 +3,7 @@ chain = "rococo-local"
 
 [[relaychain.nodes]]
 name = "alice"
+rpc_port = 8833
```
evilrobot-01, line 6: Shouldnt be included in this PR (out of scope)

crates/pop-cli/src/commands/up/parachain.rs
```diff
@@ -138,6 +145,16 @@ impl ZombienetCommand {
 					}
 				}
 
+				if self.setup_script.is_some() {
+					spinner.set_message("Initializing your setup script...");
+					// sleep for 15 seconds to allow the network to start up
+					sleep(Duration::from_secs(15)).await;
```
brunopgalvao, line 151: @evilrobot-01 I am not comfortable with using `sleep` to make sure the network is up. Is there any other way of solving this? `sleep` could vary per machines performance and how complex the network that the user is spinning up e.g. 1 parachain vs 10 parachains

crates/pop-cli/src/commands/up/parachain.rs
brunopgalvao: Updated.

tests/zombienet.toml
brunopgalvao: Removed.

crates/pop-cli/src/commands/up/parachain.rs
```diff
@@ -151,6 +159,26 @@ impl ZombienetCommand {
...
 
+pub(crate) async fn run_custom_command(
+	spinner: &ProgressBar,
+	command: &str,
+) -> Result<(), anyhow::Error> {
+	spinner.set_message(format!("Running command: {}", command.to_string()));
+	sleep(Duration::from_secs(15)).await;
+
+	// Split the command into the base command and arguments
+	let mut parts = command.split_whitespace();
```
AlexD10S, line 170: Instead of parsing here, we can use `clap`, and has has the arguments like this:
```
/// The command to run after the network has been launched.
#[clap(name = "cmd", short = 'c', long)]
command: Option<String>,
/// The script arguments, encoded as strings.
#[clap(long, num_args = 0..)]
args: Vec<String>,
```

evilrobot-01: Does this not require the specification of an additional `--args`? We dont want this feature to make the usage more complex, hence my suggestion of having it as a single arg.

AlexD10S (commented):
Some comments:
- We need to explain this option in the `README` and if is with an useful example on how can be useful for devs much better.
- If I run the example `--cmd "echo Hello from Pop CLI"` as you suggest for a quick test it looks a bit weird, it gets stuck for a few seconds in Running command: `echo Hello from Pop CLI`... without really executing it and after it clear the screen to display the information of the `pop up`.  UX has to be improved, maybe showing the message that is running the command after spawn the networks?
- I left a comment about parsing the arguments of the script: pop up -cmd "echo" --args ...

crates/pop-cli/src/commands/up/parachain.rs
AlexD10S: Yes it needs an additional `--args`. At first it looked weird to me everything in one command, but I see your point.

Example of run a script `fund-dev-accounts` that has an argument for the balance to fund:
```
--cmd "./fund-dev-accounts 1000" 
```
or
```
--setup_script fund-dev-accounts --args 1000
```

We can start with `--cmd "./fund-dev-accounts 1000"` and iterate with user's feedback.

evilrobot-01: I agree that the args is probably easier for us, but I think it probably clutters things with noise. A single arg keeps the feature self-contained.

I like starting with simple and taking feedback to improve.

peterwht:
This would be good to test with a simple Subxt script that connects to the launched network.

brunopgalvao:
> This would be good to test with a simple Subxt script that connects to the launched network.

Created a simple script.

```
git clone https://github.com/brunopgalvao/set-balance
cd set-balance
cargo build
```

./tests/zombienet.toml
```
[relaychain]
chain = "rococo-local"

[[relaychain.nodes]]
name = "alice"
rpc_port = 8833
validator = true

[[relaychain.nodes]]
name = "bob"
validator = true
```
```
pop up parachain -f ./tests/zombienet.toml -r v1.8.0 -p --cmd /Users/bruno/src/set-balance/target/debug/set-balance
```

Output
Screenshot 2024-05-22 at 5 25 45 PM

Screenshot 2024-05-22 at 5 26 13 PM

AlexD10S:
> > This would be good to test with a simple Subxt script that connects to the launched network.
Thanks, LGTM now! 
This example makes it easier to test the function. I have added an issue in the documentation to convert this example into a tutorial: https://github.com/r0gue-io/pop-docs/issues/2

AlexD10S (approved):

PR #182, opened 2024-05-17
feat: add paseo support

evilrobot-01:
Adds `chain-spec-generator` support, so that `pop up` can be used to launch local `paseo`, `kusama` and `polkadot` networks.


AlexD10S (approved):
Great job. Support for other chains, such as `asset-hub` and `bridge-hub`... can be easily added if needed.

Issue #129, opened 2024-05-01
UX: `pop up parachain`


brunopgalvao:
There has been discussion around the `pop up parachain` command.

What if the user simply wants to spin up a Relay chain? 
e.g. `pop up parachain -f network.toml` - where the network.toml file only has the Relay chain

Perhaps the command should be: `pop up polkadot`

Anyways, I have never seen a parachain run without a Relay chain so better UX would be to rename the `pop up parachain` command to `pop up polkadot`. This would equate to spinning up a Polkadot network. Whether it is a Polkadot network with 10 parachains or simply just the Relay chain, this command suffices.

Ser, what about Substrate solo chains? We can accommodate for this by having an additional command:
```
pop up chain
```
What we really want is to be in the directory of the project and for the command to be: 
```
pop up
``` 
which would result in automatically either a) detect the network.toml file and spin up the network, or b) recognize that it is a Substrate solo chain and run it

## Deliverables

Rename `pop up parachain` to `pop up polkadot`

Make it so `pop up polkadot` by itself, with no network.toml file, will spin up a Relay chain in dev mode
- If the network.toml file exists in same directory, it overrrides, use that

Add an additional command to support Substrate solo chains: `pop up chain`

Make it so the user only has to run `pop up` and pop-cli will look in the directory for the network.toml file as well as detect whether it is a Substrate solo chain and deploy the network accordingly.

## Nice-to-haves

Configuration files within Pop CLI for networks e.g. Acala, Moonbeam, etc.
User can then say: `pop up acala` or `pop up acala, asset_hub, moonbeam`

AlexD10S:
I wouldn’t recommend changing`pop up parachain` to  `pop up polkadot`
To me pop up parachain makes sense,In my mind the typical flow for a new user is as follows
```
pop up new parachain
pop up build parachain
pop up parachain -f network.toml
```

However I like the idea of `pop up polkadot` that spin up a Relay chain in dev mode or other chains. 
So I'd vote in favour of having both: pop up parachain, pop up polkadot

brunopgalvao:
We will not be touching
```
pop ___ parachain
```
So:
```
pop new parachain
pop build parachain
```
will remain intact.

However, if one wants to spin up a network they should be able to:
```
pop up
```
within the directory of their network.toml file.

Therefore no need for the command:
```
pop up parachain
```
If they are outside the project directory then:
```
[...]

Issue #35, opened 2024-03-07
`pop up contract`: run substrate-contracts-node if it does not exist


brunopgalvao:
Based on this PR comment: https://github.com/r0gue-io/pop-cli/pull/27#discussion_r1515973155

For `pop up contract`, we should do a check, if there is no substrate node running on localhost:9944 then have Pop CLI run one.

AlexD10S:
Same as commented here: https://github.com/r0gue-io/pop-cli/issues/31#issuecomment-2124252867
When this PR is merged https://github.com/r0gue-io/pop-cli/pull/185 we can just call that functionality to run the local node

PR #185, opened 2024-05-20
Add the "pop up contracts-node" command

Moliholy:
# Description

This PR adds the following command that spawns a [substrate-contracts-node](https://github.com/paritytech/substrate-contracts-node) instance.

```sh
pop up contracts-node
```

It should be executed in a different terminal, and prior to its execution it will check whether the repository is downloaded and build the binary if it isn't.

It's definitely not rocket science, and could be improved, but IMO it's a decent step forward towards a better usage by integrating in the tool a component that was excluded, yet it was used for deploying smart contracts.

## Possible improvements

- Check whether the current version of the software is the latest, and if not force-download the latest.
- The above, but with a `--force`-like command.
- Include features when building the binary.
- Pass command-line arguments to the node for more sophisticated setup.

### DIFF
--- a/README.md
+++ b/README.md
@@ -196,6 +196,12 @@ Pop provides the latest version out-of-the-box by running:
 pop up contracts-node
 ```
 
+If you want to run any other node with `pallet-contracts` you need to change `CONTRACTS_NODE` environment variable:
+
+```sh
+export CONTRACTS_NODE="YOUR_CONTRACTS_NODE_PATH"
+```
+
 Run e2e testing on the Smart Contract:
 
 ```sh
--- a/README.md
+++ b/README.md
@@ -137,12 +137,10 @@ Build the Smart Contract:
 pop build contract -p ./my_contract
 ```
 
-To deploy a Smart Contract you need a chain running. For testing purposes one option is to
-run [substrate-contracts-node](https://github.com/paritytech/substrate-contracts-node):
+To deploy a Smart Contract you need a chain running. For testing purposes you can simply spawn a contract node:
 
 ```sh
-cargo install contracts-node
-substrate-contracts-node
+pop up contracts-node
 ```
 
 > :information_source: We plan to automate this in the future.
@@ -192,17 +190,10 @@ pop call contract -p ./my_contract --contract $INSTANTIATED_CONTRACT_ADDRESS --m
 ## E2E testing
 
 For end-to-end testing you will need to have a Substrate node with `pallet contracts`.
-You do not need to run it in the background since the node is started for each test independently.
-To install the latest version:
+Pop provides the latest version out-of-the-box by running:
 
-```
-cargo install contracts-node --git https://github.com/paritytech/substrate-contracts-node.git
-```
-
-If you want to run any other node with `pallet-contracts` you need to change `CONTRACTS_NODE` environment variable:
-
-```
-export CONTRACTS_NODE="YOUR_CONTRACTS_NODE_PATH"
+```sh
+pop up contracts-node
 ```
 
 Run e2e testing on the Smart Contract:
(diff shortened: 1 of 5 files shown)


AlexD10S:
Hola @Moliholy! Great to see you here contributing to pop-cli. Looks good to me!

We are currently having an open discussion about the best UX for `pop up`: https://github.com/r0gue-io/pop-cli/issues/129
I like `pop up contracts-node` , but want to see @brunopgalvao thoughts on it.


Also after this PR is merged, we can tackle this issue: https://github.com/r0gue-io/pop-cli/issues/35

Moliholy:
@AlexD10S pleased to be around :-)

Yes, I saw that issue before, but I supposed its decision was not taken since it was still open, so I thought the best approach would be, at least for now, to continue with the current project's structure.

As for the name itself, I definitely don't have a strong opinion nor I'm particularly good at naming, so if you find a better one feel free to change it.

README.md
```diff
@@ -192,17 +190,10 @@ pop call contract -p ./my_contract --contract $INSTANTIATED_CONTRACT_ADDRESS --m
...
+Pop provides the latest version out-of-the-box by running:
 
-```
-cargo install contracts-node --git https://github.com/paritytech/substrate-contracts-node.git
-```
-
-If you want to run any other node with `pallet-contracts` you need to change `CONTRACTS_NODE` environment variable:
-
-```
-export CONTRACTS_NODE="YOUR_CONTRACTS_NODE_PATH"
```
brunopgalvao, line 205: This is used to point to the `contracts-node` binary which is needed for running the e2e tests.

`e2e-tests` use the `contracts-node` binary to test against.

If you want to run e2e tests against a live contracts node you can look here:
- https://github.com/use-ink/ink-docs/pull/283/files#diff-b6664aae173c30e529697e60b09951bdcad954788b1452c97fc352514065e88cR267

crates/pop-cli/src/commands/up/contracts_node.rs
```diff
@@ -0,0 +1,31 @@
...
+use crate::style::{style, Theme};
+
+const BIN_NAME: &str = "substrate-contracts-node";
+
+#[derive(Args)]
+pub(crate) struct ContractsNodeCommand;
+impl ContractsNodeCommand {
+	pub(crate) async fn execute(&self) -> anyhow::Result<()> {
+		clear_screen()?;
+		intro(format!("{}: Deploy a contracts node", style(" Pop CLI ").black().on_magenta()))?;
```
AlexD10S, line 16: We currently use 'deploy a contracts node,' but it doesn't sound correct. In an ongoing PR, we are already changing it to `Launch`: (https://github.com/r0gue-io/pop-cli/pull/186/files#diff-676bc9da9a6a8b135164d15340c548a94dcad29b8cd784ebda6a9e3ef2313622R37).
Could you change the message to "Launch a local contracts node"?

AlexD10S (approved):
Just added a little change in a message but the rest looks good to me!

Thanks again for contributing!!

README.md
Moliholy: Yes... but if you're providing now a node inside the tool, why would a user want to do that?

crates/pop-cli/src/commands/up/contracts_node.rs
Moliholy: Done!

Moliholy:
Rebased to the latest `main`.

README.md
brunopgalvao: So how do you run the e2e tests?

Moliholy: I just restored that part.

AlexD10S (approved):

Issue #184, opened 2024-05-20
zombienet-sdk using 0.2 in Cargo.toml when 0.2.1 is available


ltfschoen:
https://github.com/r0gue-io/pop-cli/blob/main/Cargo.toml#L55
```
# zombienet-sdk = "0.2"
# zombienet-support = "0.2"
zombienet-sdk = { git = "https://github.com/r0gue-io/zombienet-sdk", branch = "pop", version = "0.2.1" }
zombienet-support = { git = "https://github.com/r0gue-io/zombienet-sdk", branch = "pop", version = "0.2.1" }
```

Daanvdplas:
Closed by #186

PR #186, opened 2024-05-21
fix: container ux

evilrobot-01:
- tests: refactors to improve testability, increasing coverage. There is still more work to be done though.
- ci: adds `docker` job to simply ensure the docker image builds. An additional issue should be opened up to ensure that all commands are tested on all platforms, including on the docker image. This may require publishing, but I dont think we are ready to publish a docker image yet.
- docs: needs to be addressed separately, once this is approved.

A general guide on manually testing as follows.

# Overview
The following guide provides an overview on launching a local network, including a parachain generated from a template, within a container. Note that port accessibility from outside of the container still needs to be addressed, with suggested workarounds provided at the end. It is suggested to review these before starting with anything else to assess suitability.

Build and then run `pop` within a container interactively, optionally mounting a local directory to pop's cache within the container for reuse.
```shell
docker build -t pop .
docker run -it -v ./.cache:/root/.cache/pop --entrypoint bash pop
```

Then generate a parachain using `pop`, changing into the relevant directory once done:
```shell
pop new parachain
cd ./my-parachain
```

Before building the parachain, source the required binary dependencies for launching the local network by simply specifying the relevant network configuration file (e.g. `network.toml`). These binaries are typically those required to run the relay chain or any system parachains, but could also include other parachains in your solution beyond those which you are developing.

```shell
```

You should be shown which binaries are missing and be prompted to source them automatically where possible, using the latest versions available when not explicitly specified using the command line arguments. See `pop up parachain --help` for more information.

> Note: We currently do this before building the parachain simply to conserve space within the container, as any build artifacts are automatically removed once these binaries are built. This is currently not the case when building a locally generated parachain, where retaining build artifacts is preferred for faster rebuilds of the parachain after local changes. The `pop up ...` command will be updated to automatically build the parachain as required in the near future.

Now build the parachain binary and then finally launch the local network.
```shell
pop build parachain
```

> Note: the ports used by each of the network nodes will remain inaccessible from outside of the container by default. Consider using the `--net=host` beta feature available on Docker Desktop (YMMV). More information is available at https://docs.docker.com/network/network-tutorial-host/#prerequisites. An alternative is expose a known port for each network node when launching the container and then specify those ports within the generated `network.toml` file - e.g. `rpc_port = 8833` before launching the local network within the container.


Cargo.toml
```diff
@@ -52,8 +52,8 @@ toml_edit = { version = "0.22", features = ["serde"] }
 symlink = "0.1"
 serde_json = "1.0"
 serde = { version = "1.0", features = ["derive"] }
-zombienet-sdk = "0.2"
-zombienet-support = "0.2"
+zombienet-sdk = "0.2.2"
```
AlexD10S, line 55: This closes https://github.com/r0gue-io/pop-cli/issues/184

AlexD10S (approved):
Great PR! 
We will use this refactor to increase test coverage as a reference for the `pop_contracts` crate.

PR #191, opened 2024-05-23
feat(new): ux improvements

peterwht:
This PR introduces UX improvements to the pop new parachain command.
- Warns user if template does not have any releases
- Introduces an optional "SupportedVersions" field to the Template variants.
- Introduces an optional "IsAudited" field to the Template variants. If set to true, will hide the audit warning.
- Displays the repo license for the template

SupportedVersions is introduced as a simple way to support OpenZeppelin templates. Right now, OpenZeppelin's default branch uses a monorepo style. However, the `v1.0.0` release does not use a monorepo. By specifying the supported version, we can control the flow in a consistent way. When OpenZeppelin introduces a release with the monorepo, we will be able to use the supported versions to not support v1.0.0 and we can introduce additional parameters to properly setup the sub-directory.

By default `SupportedVersion` (when not present) supports all versions and default branch. However, if something is set, then it will only support those versions.

### DIFF
--- a/crates/pop-cli/src/commands/new/parachain.rs
+++ b/crates/pop-cli/src/commands/new/parachain.rs
@@ -131,13 +131,8 @@ async fn guide_user_to_generate_parachain() -> Result<NewParachainCommand> {
 	let template = display_select_options(provider)?;
 
 	let url = url::Url::parse(&template.repository_url()?).expect("valid repository url");
-	// Get only the latest 3 releases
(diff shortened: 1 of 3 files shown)


AlexD10S:
Great job! 
Unit tests are failing because of the new test templates added. Is an easy fix, just add them in these functions: https://github.com/r0gue-io/pop-cli/blob/main/crates/pop-parachains/src/templates.rs#L219

PR #197, opened 2024-06-04
feat: add OpenZeppelin template

peterwht:
This PR adds support for the OpenZeppelin Generic Polkadot Runtime Template.

Non-audited OpenZeppelin templates are not supported. OpenZeppelin's only release is not in the monorepo format. So, any future releases will need additional changes to support the monorepo. 

Co-authored-by: Frank Bell <60948618+evilrobot-01@users.noreply.github.com>


al3mart (approved):
Awesome!

AlexD10S (approved):

Issue #124, opened 2024-04-29
Organization: `crates/common`


weezy20:
A common home for utilities that are not necessarily pop specific but can be also used by application code including our crates to build on top of. Some existing code may include aggregation for `Github`, `write_to_file`, `create_signer` and other helpful utilities that are present in `utils/` module. Can also serve as home for the upcoming caching layer  (pop_db) or project detection (`is_parachain` and `is_contract`).

Issue #31, opened 2024-03-07
Add `CONTRACTS_NODE_PATH` to Pop when running e2e tests for contracts


brunopgalvao:
Based on this PR comment: https://github.com/r0gue-io/pop-cli/pull/22#issuecomment-1980361632

The idea would be devs can run e2e tests like this:
```
pop test contract  -p ./my_contract --features e2e-tests -n /Users/bruno/src/substrate-contracts-node/target/debug/substrate-contracts-node
```

AlexD10S:
When this PR is merged https://github.com/r0gue-io/pop-cli/pull/185 can we just call that functionality to run the local node?

PR #206, opened 2024-06-12
refactor: run substrate-contracts-node in `pop up contract` if it does not exist

AlexD10S:
- For `pop up contract`, check if the specified chain is live. If not, prompt the user to run the Substrate contracts node in the background for testing using pop-cli.
- This PR refactors the previous PR https://github.com/r0gue-io/pop-cli/pull/185 by removing the  `pop up contracts-node` command and reusing the logic to run the `contracts-node` exclusively for testing purposes. This improves the `pop up contract` user experience and will be reused for automatic execution in e2e functionality (https://github.com/r0gue-io/pop-cli/issues/31).
- Improves the UX by downloading the binaries directly instead of building the node, following the approach Frank is implementing for `pop up parachain` in this PR https://github.com/r0gue-io/pop-cli/pull/199

Closes https://github.com/r0gue-io/pop-cli/issues/35

**Reviewer Considerations**
- The GitHub functionality is the same code as in the `pop_parachains` crate. I copied the necessary logic into the `pop_contracts` crate, but does it make more sense to just import the `pop_parachains` crate instead?
Ideally, we should have a common crate as suggested here: https://github.com/r0gue-io/pop-cli/issues/124, but that seems out of scope for this PR. 
- The contracts node runs in the background, so after the contract deployment is finished, it remains running. The user will need to manually stop it.


peterwht (approved):

al3mart (changes requested):
Nice!

Overall I think this flow is nicer.

Regarding that git duplicated code I agree we should move in the way of having a common crate, or maybe it makes sense importing it from the existing one just to be able to move ahead and then we address the common crate later on.

Given that we are launching a para for the user if no arguments are given, we could also use `//Alice` by default if no other suri is given.

When dry-running the changes I got stuck at:

```
│  
◇  The chain "ws://localhost:9944/" is not live. Would you like pop to start a local node in the background for testing?
│  Yes 
│
```

And didn't see any progress from there on. I just built a contract and run `pop up contract --suri //Alice`. Is this expected ?

crates/pop-cli/src/commands/up/contract.rs
```diff
@@ -68,6 +68,21 @@ impl UpContractCommand {
 			log::success(result.to_string())?;
 		}
 
+		if !is_chain_alive(self.url.clone()).await? {
+			if !confirm(format!(
```
Moliholy, line 72: I'd actually allow the user to pass a flag to skip this check. Thinking about CI.

Moliholy:
> Given that we are launching a para for the user if no arguments are given, we could also use //Alice by default if no other suri is given.

@al3mart just some heads up: when using `EthereumSignature` instead of `MultiSignature`, `//Alice` does not work. Instead, [Alith private key](https://github.com/polkadot-js/common/blob/e3b11360ab381a51bb05544cf2872646037d3579/packages/keyring/src/testing.ts#L80) should be used.

AlexD10S:
> Nice!
> 
> Overall I think this flow is nicer.
> 
> Regarding that git duplicated code I agree we should move in the way of having a common crate, or maybe it makes sense importing it from the existing one just to be able to move ahead and then we address the common crate later on.
> 
> Given that we are launching a para for the user if no arguments are given, we could also use `//Alice` by default if no other suri is given.
> 
> When dry-running the changes I got:
> 
> ```
> ◆  Local node started successfully in the background.
> │  
> ▲  NOTE: The contracts node is running in the background with process ID 188188. Please close it manually when done testing.
> │  
> ┌   Pop CLI : Deploy a smart contract
> │
> Error: Invalid number of input arguments: expected 1, 0 provided
> ```
> 
> I just built a contract and run `pop up contract --suri //Alice`. @AlexD10S did you see this ?

The error thrown is correct. To initialize the contract, you must call the constructor method. If not specified, pop-cli uses 'new' by default. For the smart contract you are trying to deploy, this constructor expects one argument:
```
        #[ink(constructor)]
        pub fn new(init_value: bool) -> Self {
```
The argument has to be specified with the `--args` flag. If you try `pop up contract --suri //Alice --args false` works. 
The same behaviour happens with `cargo-contracts`. If you try `cargo contract instantiate --suri //Alice` you will see the same error.

Also, I pushed a new commit implementing your suggestion to use the dev account by default (//Alice) if the account for deploying the contract is not specified (--suri).

crates/pop-cli/src/commands/up/contract.rs
Moliholy: I meant... users using pop are the ones that would execute this command in the CI. That's what I meant.

peterwht (approved):

al3mart (approved):

al3mart:
> The error thrown is correct. To initialize the contract, you must call the constructor method.

So right! That was a bad review from my side. For whatever reason I started passing no arguments on my testing. :pray:

AlexD10S:
> > The error thrown is correct. To initialize the contract, you must call the constructor method.
> 
> So right! That was a bad review from my side. For whatever reason I started passing no arguments on my testing. 🙏

You made a good point; the error message is not very clear. However, I'm unsure how we can improve it because the error is thrown by the contract_extrinsics. We could add a check in pop-cli to retrieve the constructor and compare the expected arguments.

crates/pop-cli/src/commands/up/contract.rs
AlexD10S: You are right, I have added a `skip_confirm` flag.

AlexD10S:
> 
> @al3mart just some heads up: when using `EthereumSignature` instead of `MultiSignature`, `//Alice` does not work. Instead, [Alith private key](https://github.com/polkadot-js/common/blob/e3b11360ab381a51bb05544cf2872646037d3579/packages/keyring/src/testing.ts#L80) should be used as the uri.
> 
> You can check whether the blockchain uses `EthereumSignature` by checking the chainspec includes the `isEthereum: true` flag.

In my opinion, this is out of scope for this PR. We are safe using `//Alice `if no `--suri` is indicated. This is just a default scenario. If the chain has `EthereumSignature`, the user will have to specify the keys they want to use. But thanks for sharing, please feel free to add it to the issues, and we will consider adding this feature..

al3mart (approved):
LGTM

Issue #176, opened 2024-05-16
pop up parachain: build parachain if it has not been built


AlexD10S:
Similar approach of the issue pop up contract: build contract if it has not been built https://github.com/r0gue-io/pop-cli/issues/34

In this case check if the pop up needs a binary of the local chain that has not been build yet, and if this is the case builds the parachain calling https://github.com/r0gue-io/pop-cli/blob/main/crates/pop-parachains/src/build.rs#L5 

This is a bit more tricky than the pop up contract because has to parse the zombienet.toml file and see we are in the parachain folder before run build

bolajahmad:
Please assign this to me, for some reasons I can't assign it to myself. 

[parachain label](https://github.com/r0gue-io/pop-cli/labels/parachain) is okay, but this also affects the cli. We should have a label for cli-affected issues too.

bolajahmad:
Closing this issue as it has been fixed automatically by an [approved PR](https://github.com/r0gue-io/pop-cli/pull/199).

@AlexD10S

PR #207, opened 2024-06-14
draft: build parachain on missing local binaries

bolajahmad:
While trying to spin up a substrate node using `pop up parachain`, the command will fail if the node has not been built/compiled already using `pop build parachain`.

This PR will automatically check if there is a build/binary existing for the node and then allow `pop up...` to progress.

### DIFF
--- a/crates/pop-cli/src/commands/up/parachain.rs
+++ b/crates/pop-cli/src/commands/up/parachain.rs
@@ -7,9 +7,10 @@ use cliclack::{
 };
 use console::{Emoji, Style};
 use duct::cmd;
-use pop_parachains::{Error, NetworkNode, Source, Status, Zombienet};
+use pop_parachains::{build_parachain, Error, NetworkNode, Source, Status, Zombienet};
 use std::{fs::remove_dir_all, time::Duration};
 use tokio::time::sleep;
+use std::path::PathBuf;
 
 #[derive(Args)]
 pub(crate) struct ZombienetCommand {
@@ -160,6 +161,9 @@ impl ZombienetCommand {
 			// Check for any local binaries which need to be built manually
 			let local: Vec<_> = missing.iter().filter(|(_, _, local)| *local).collect();
 			if local.len() > 0 {
+				for (name, binary, _) in local.iter() {
+					let _ = build_parachain(&Some(PathBuf::from(binary.name.clone())));
+				}
 				outro_cancel(
 					"🚫 Please manually build the missing binaries at the paths specified and try again.",
 				)?;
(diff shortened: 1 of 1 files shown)


crates/pop-cli/src/commands/up/parachain.rs
```diff
@@ -83,6 +84,23 @@ impl ZombienetCommand {
 			.to_string();
 			log::warning(format!("⚠️ The following binaries specified in the network configuration file cannot be found locally:\n   {list}"))?;
 
+			// Check for any local binaries which need to be built manually
+			let local: Vec<_> = missing.iter().filter(|(_, _, local)| *local).collect();
+			if local.len() > 0 {
+
+				for (name, binary, _) in local.iter() {
```
AlexD10S, line 91: There is a warning binary is not used.

crates/pop-cli/src/commands/up/parachain.rs
```diff
@@ -83,6 +84,23 @@ impl ZombienetCommand {
...
+					intro(format!("{}: Building your parachain", style(" Pop CLI ")
+						.black()
+						.on_magenta()
+					))?;
+					let _ = build_parachain(&Some(PathBuf::from(name.clone())))?;
```
AlexD10S, line 96: No need for `.clone()` here

AlexD10S, line 96: No need for `let _` if you use `?` after build_parachain

crates/pop-cli/src/commands/up/parachain.rs
```diff
@@ -83,6 +84,23 @@ impl ZombienetCommand {
 			.to_string();
 			log::warning(format!("⚠️ The following binaries specified in the network configuration file cannot be found locally:\n   {list}"))?;
 
+			// Check for any local binaries which need to be built manually
+			let local: Vec<_> = missing.iter().filter(|(_, _, local)| *local).collect();
+			if local.len() > 0 {
+
+				for (name, binary, _) in local.iter() {
+					intro(format!("{}: Building your parachain", style(" Pop CLI ")
```
AlexD10S, line 92: To keep consistency with the funcionality we have in Contracts, can you add a warning message too? like the one you added for contracts?
`log::warning(format!("NOTE: contract has not yet been built."))?;`

AlexD10S (changes requested):
When I try to run it without building it i am getting an error:
```
┌   Pop CLI : Building your parachain
│
Error: No such file or directory (os error 2)
```

I think is because of the path you are sending to the `build_parachain` function is incorrect.
The steps to replicate the error are:
`pop new parachain my-parachain`
`cd  my-parachain`
`pop up parachain -f network.toml`

AlexD10S:
Closing this PR as the issue https://github.com/r0gue-io/pop-cli/issues/176 has been closed but this other PR: https://github.com/r0gue-io/pop-cli/pull/199

Issue #183, opened 2024-05-20
refactor(readme): shorten


Daanvdplas:
The readme is very heavy in terms of text while we have the pop cli docs. I suggest we do it like hardhat: https://github.com/NomicFoundation/hardhat

Basically moving all info that explains how to use pop cli to the pop cli docs and in the README provide info about pop, how to contribute, install etc.

AlexD10S:
This was solved already in the PR: https://github.com/r0gue-io/pop-cli/pull/223

PR #209, opened 2024-06-17
chore: set `CONTRACTS_NODE_PATH` env variable for e2e tests

AlexD10S:
Closes https://github.com/r0gue-io/pop-cli/issues/31

As specified in the issue, implemented the flag `--contracts-node` with the path to the contracts-node binary to be used in the end-to-end (e2e) tests. This will allow `pop` to set the environment variable for the e2e tests as detailed here:  https://use.ink/basics/contract-testing/end-to-end-e2e-testing
```
pop test contract  -p ./my_contract --features e2e-tests -n /Users/bruno/src/substrate-contracts-node/target/debug/substrate-contracts-node
```


README.md
```diff
@@ -202,22 +202,26 @@ pop call contract -p ./my_contract --contract $INSTANTIATED_CONTRACT_ADDRESS --m
...
+ pop test contract  -p ./my_contract --features e2e-tests
 ```
 
-Run e2e testing on the Smart Contract:
+If you want to run any other node with `pallet-contracts` you need to change `CONTRACTS_NODE` environment variable:
 
+```sh
+export CONTRACTS_NODE="YOUR_CONTRACTS_NODE_PATH"
+```
+You can automate this process with the `--contracts-node` flag:
```
Daanvdplas, line 221:
```suggestion
Or, use the `--contracts-node` flag:
```

This should be the first choice TMO, second doing the export

crates/pop-contracts/src/test.rs
```diff
@@ -22,7 +22,14 @@ pub fn test_smart_contract(path: &Option<PathBuf>) -> Result<(), Error> {
 /// # Arguments
 ///
 /// * `path` - location of the smart contract.
```
Daanvdplas, line 24: missing parameter docs

README.md
```diff
@@ -202,22 +202,26 @@ pop call contract -p ./my_contract --contract $INSTANTIATED_CONTRACT_ADDRESS --m
...
+Run e2e testing on the Smart Contract:
 
 ```sh
-export CONTRACTS_NODE="YOUR_CONTRACTS_NODE_PATH"
+# Run e2e tests for an existing smart contract
+ pop test contract  -p ./my_contract --features e2e-tests
 ```
 
-Run e2e testing on the Smart Contract:
+If you want to run any other node with `pallet-contracts` you need to change `CONTRACTS_NODE` environment variable:
```
Daanvdplas, line 216:
```suggestion
If you want to run a different node with `pallet-contracts` you need to change `CONTRACTS_NODE` environment variable:
```

README.md
```diff
@@ -202,22 +202,26 @@ pop call contract -p ./my_contract --contract $INSTANTIATED_CONTRACT_ADDRESS --m
...
 ```
 cargo install contracts-node --git https://github.com/paritytech/substrate-contracts-node.git
 ```
 
-If you want to run any other node with `pallet-contracts` you need to change `CONTRACTS_NODE` environment variable:
+Run e2e testing on the Smart Contract:
 
 ```sh
-export CONTRACTS_NODE="YOUR_CONTRACTS_NODE_PATH"
+# Run e2e tests for an existing smart contract
```
Daanvdplas, line 212: Don't see the use for this comment, more cases like this in the README. Most of them have a description already above the code import. Should be fixed when fixing #183

crates/pop-cli/src/commands/test/contract.rs
```diff
@@ -14,6 +14,12 @@ pub(crate) struct TestContractCommand {
 	path: Option<PathBuf>,
 	#[arg(short = 'f', long = "features", help = "Features for the contract project")]
 	features: Option<String>,
+	#[arg(
+		short = 'n',
+		long = "contracts-node",
+		help = "Path to the contracts node to run the e2e tests [default: none]"
```
Daanvdplas, line 20:
```suggestion
		help = "Path to the contracts node to run e2e tests [default: none]"
```

crates/pop-cli/src/commands/test/contract.rs
```diff
@@ -26,7 +32,7 @@ impl TestContractCommand {
 				style(" Pop CLI ").black().on_magenta()
 			))?;
```
Daanvdplas, line 34: Unnecessary newlines in this function. Newlines should in my opinion only have to be used in big code blocks and separate processes that happen. An `if else` statement is in my opinion one process.

crates/pop-cli/src/commands/test/contract.rs
```diff
@@ -14,6 +14,12 @@ pub(crate) struct TestContractCommand {
...
 	features: Option<String>,
+	#[arg(
+		short = 'n',
+		long = "contracts-node",
+		help = "Path to the contracts node to run the e2e tests [default: none]"
+	)]
+	contracts_node_path: Option<PathBuf>,
 }
 
 impl TestContractCommand {
```
Daanvdplas, line 25: May I ask why we are taking a reference to `self` in `execute(&self)`? It requires to do a lot of clones while `self` is not used upstream. Not 100% sure but perhaps worth looking into.

crates/pop-contracts/src/test.rs
```diff
@@ -31,26 +38,39 @@ pub fn test_e2e_smart_contract(path: &Option<PathBuf>) -> Result<(), Error> {
...
-#[cfg(feature = "unit_contract")]
 #[cfg(test)]
 mod tests {
 	use super::*;
-	use std::fs;
 	use tempfile;
 
-	fn setup_test_environment() -> Result<tempfile::TempDir, Error> {
+	#[test]
+	fn test_contract_test() -> Result<(), Error> {
```
Daanvdplas, line 47: These tests only check whether it works correctly in the positive scenario. I'm missing tests where scenarios that should not work are tested as well. For example, I am now able to run `pop test contract` in my parachain crate.

Daanvdplas (changes requested):
- If I do `pop test contract --features --help` it doesn't display me which features there are. Would be nice it this could be provided.

README.md
AlexD10S: You are right, actually the export CONTRACTS_NODE is not needed anymore

crates/pop-cli/src/commands/test/contract.rs
AlexD10S: Fixed in this PR: https://github.com/r0gue-io/pop-cli/pull/217

AlexD10S:
> * If I do `pop test contract --features --help` it doesn't display me which features there are. Would be nice it this could be provided.

I noticed that too. I was unsure about including it in this PR, but based on your comment, it makes sense to fix it now.

Daanvdplas (approved):
LGTM!

crates/pop-cli/src/commands/test/contract.rs
```diff
@@ -10,27 +10,30 @@ use std::path::PathBuf;
 pub(crate) struct TestContractCommand {
 	#[arg(short = 'p', long, help = "Path for the contract project [default: current directory]")]
 	path: Option<PathBuf>,
-	#[arg(short = 'f', long = "features", help = "Features for the contract project")]
+	#[arg(short = 'f', long = "features", help = "Features for the contract project", value_parser=["e2e-tests"])]
 	features: Option<String>,
+	#[arg(
+		short = 'n',
+		long = "contracts-node",
```
evilrobot-01, line 17: Why does this have to be so wordy? The whole point of choosing pop (and all the shorter abbreviations for subcommands) was to make it as short to type as possible. Why cant it just be `--node`?

crates/pop-cli/src/commands/test/contract.rs
```diff
@@ -10,27 +10,30 @@ use std::path::PathBuf;
...
+		help = "Path to the contracts node to run e2e tests [default: none]"
+	)]
+	contracts_node_path: Option<PathBuf>,
```
evilrobot-01, line 20: Its obvious its a path based on the type. Suggest just `node`

crates/pop-contracts/src/test.rs
```diff
@@ -31,26 +41,60 @@ pub fn test_e2e_smart_contract(path: &Option<PathBuf>) -> Result<(), Error> {
...
```
evilrobot-01, line 50: Personal opinion, but prefixing the test names with `test_` when there is a `#[test]` attribute seems redundant. I know it makes sense here due to the name of the function being tested, but I have seen it being used elsewhere. But then there is a suffix of `_test` - perhaps consider `_works` instead.

evilrobot-01 (commented):
General observations.

crates/pop-cli/src/commands/test/contract.rs
AlexD10S: Makes sense to keep it easier. Renamed to `--node`

crates/pop-cli/src/commands/test/contract.rs
AlexD10S: Same as previous comment, renamed.

crates/pop-contracts/src/test.rs
AlexD10S: I removed the `_test` suffix and renamed the tests included in this PR. 
I agree it makes sense for the other tests to exclude the `test_` prefix, which will be addressed in a separate PR.

evilrobot-01 (approved):

README.md
```diff
@@ -225,22 +225,21 @@ pop call contract -p ./my_contract --contract $INSTANTIATED_CONTRACT_ADDRESS --m
...
+ pop test contract  -p ./my_contract --features e2e-tests
 ```
 
-Run e2e testing on the Smart Contract:
-
+If you want to run a different node with `pallet-contracts` you need to change `CONTRACTS_NODE` environment variable using the  `--contracts-node` flag:
```
evilrobot-01, line 239: Shouldnt this be `--node` flag?

evilrobot-01 (changes requested):
Just one item in the docs which I think was missed after the arg rename.

README.md
AlexD10S: Yes! Apologies for this mistake

evilrobot-01 (approved):
LGTM!

PR #216, opened 2024-06-22
feat: clean cache

### LINKED ISSUE #78: Feature: Cache management
`pop up parachain` currently sources polkadot/system parachain binaries and caches them for subsequent runs. The user should be able to easily see and clear what is in the cache.

Note that this was already implemented in a previous version of the tool, so I will just port it over. We just need to align on the command structure which can allow reuse for other related functionality - e.g. inspect: `pop inspect cache`, clear: `pop clear cache`

evilrobot-01:
Adds a simple command which allows a user to clean the cache by selecting those cached artifacts which are no longer required.

Note: uses 'clean' so that we can potentially add a default implementation of 'pop clean' to wrap 'cargo clean'. No real benefit apart from fewer characters to type.

Closes #78

### DIFF
--- a/crates/pop-cli/src/commands/clean.rs
+++ b/crates/pop-cli/src/commands/clean.rs
@@ -1,4 +1,5 @@
 // SPDX-License-Identifier: GPL-3.0
+
 use crate::cli::traits::*;
 use anyhow::Result;
 use clap::{Args, Subcommand};
--- a/crates/pop-cli/src/commands/clean.rs
+++ b/crates/pop-cli/src/commands/clean.rs
@@ -17,7 +17,6 @@ pub(crate) struct CleanArgs {
 /// Remove generated/cached artifacts.
 #[derive(Subcommand)]
 pub(crate) enum Command {
-	#[cfg(feature = "parachain")]
 	/// Remove cached artifacts.
 	#[clap(alias = "c")]
 	Cache,
--- a/crates/pop-cli/src/commands/clean.rs
+++ b/crates/pop-cli/src/commands/clean.rs
@@ -1,0 +1,126 @@
+// SPDX-License-Identifier: GPL-3.0
+
+use crate::{cache, style::Theme};
+use anyhow::Result;
+use clap::{Args, Subcommand};
+use cliclack::{clear_screen, confirm, intro, log, multiselect, outro, outro_cancel, set_theme};
+use console::style;
+use std::{
+	fs::{read_dir, remove_file},
+	path::PathBuf,
+};
+
+#[derive(Args)]
+#[command(args_conflicts_with_subcommands = true)]
+pub(crate) struct CleanArgs {
+	#[command(subcommand)]
+	pub(crate) command: Command,
+}
+
+/// Remove generated/cached artifacts.
+#[derive(Subcommand)]
+pub(crate) enum Command {
+	#[cfg(feature = "parachain")]
+	/// Remove cached artifacts.
+	#[clap(alias = "c")]
+	Cache(CleanCacheCommand),
+}
+
+#[derive(Args)]
+pub(crate) struct CleanCacheCommand;
+
+impl CleanCacheCommand {
+	/// Executes the command.
+	pub(crate) fn execute(self) -> Result<()> {
+		clear_screen()?;
+		set_theme(Theme);
+		intro(format!("{}: Remove cached artifacts", style(" Pop CLI ").black().on_magenta()))?;
+
+		// Get the cache contents
+		let cache = cache()?;
+		if !cache.exists() {
+			outro_cancel("🚫 The cache does not exist.")?;
+			return Ok(());
+		};
+		let contents = contents(&cache)?;
+		if contents.is_empty() {
+			outro(format!(
+				"ℹ️ The cache at {} is empty.",
+				cache.to_str().expect("expected local cache is invalid")
+			))?;
+			return Ok(());
+		}
+		log::info(format!(
+			"ℹ️ The cache is located at {}",
+			cache.to_str().expect("expected local cache is invalid")
+		))?;
+
+		// Prompt for selection of artifacts to be removed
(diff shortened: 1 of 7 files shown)


evilrobot-01:
Need https://github.com/r0gue-io/pop-cli/pull/217 merged so that I can complete the implementation of mocking for testing the implementation of the command.

crates/pop-cli/src/commands/mod.rs
```diff
@@ -33,15 +37,18 @@ pub(crate) enum Command {
 	#[clap(alias = "t")]
 	#[cfg(feature = "contract")]
 	Test(test::TestArgs),
-	/// Set up the environment for development by installing required packages.
-	#[clap(alias = "i")]
-	Install(install::InstallArgs),
+	/// Remove generated/cached artifacts.
+	#[clap(alias = "C")]
+	#[cfg(feature = "parachain")]
```
AlexD10S, line 42: It now affect contracts too. For e2e tests, we keep the `contracts-node` binary in cache as well.
https://github.com/r0gue-io/pop-cli/blob/main/crates/pop-contracts/src/utils/contracts_node.rs#L46

crates/pop-cli/src/commands/mod.rs
```diff
@@ -90,8 +97,10 @@ impl Command {
 					Err(e) => Err(e),
 				},
 			},
-			#[cfg(any(feature = "parachain", feature = "contract"))]
-			Self::Install(args) => install::Command.execute(args).await.map(|_| Value::Null),
+			#[cfg(feature = "parachain")]
```
AlexD10S, line 100: Same as above.

crates/pop-cli/src/commands/clean.rs
```diff
@@ -0,0 +1,126 @@
...
+			),
+		};
+		if !confirm(prompt).interact()? {
+			outro("ℹ️ No artifacts removed")?;
+			return Ok(());
+		}
+
+		// Finally remove selected artifacts
+		for file in &selected {
+			remove_file(file)?
```
AlexD10S, line 84: If it is a folder, it does not remove it.

AlexD10S (changes requested):
Remove `#[cfg(feature = "parachain")]` and support removing folders.

A nice-to-have feature would be the option to clean all with an `all` option in the multiselect or a `--all` flag. However, this feature is not essential for this PR. If it is not included, I will create a new issue with the `good first issue` tag for new contributors.

crates/pop-cli/src/commands/clean.rs
evilrobot-01: Do we need to remove folders? I dont expect there to be any at this moment in time, so perhaps the listing should only show files and exclude folders?

evilrobot-01:
Note: have wrapped the cliclack usage in a trait in an effort to enable testing of interactions and improve coverage in the pop-cli crate. Its a manual mock so isnt perfect, but does enable some testing of command logic. Note also that the cli should not be defined on the command as it is not state applicable to the command, but is fine for an initial iteration. This should be refactored into a command handler in the future which injects the required dependencies for a cleaner implementation.

evilrobot-01:
> Remove `#[cfg(feature = "parachain")]` and support removing folders.

Why do we need to remove folders? To my knowledge no folders are created here.

> A nice-to-have feature would be the option to clean all with an `all` option in the multiselect or a `--all` flag. 

Would be nice to have but afraid I dont have the bandwidth to add this now.

crates/pop-cli/src/commands/clean.rs
AlexD10S: When extracting the contents for the `substrate-contracts-node`, it includes the binary inside a folder named "artifacts": https://github.com/r0gue-io/pop-cli/blob/main/crates/pop-contracts/src/utils/contracts_node.rs#L59

evilrobot-01: Not saying this is the solution, but would it not be better to move it out of the archive folder on download? https://github.com/r0gue-io/pop-cli/blob/03aecfe8b684b8a0004158eb1d14d081e7253ee2/crates/pop-parachains/src/up/sourcing.rs#L230

Otherwise the implementation here is going to get more complicated. `artifacts` has no real meaning, requiring nested traversal to collect all files in all subdirectories, so we would need to prefix the name with the relative path.

evilrobot-01: Also, not sure if the contracts node checks for newer versions and prompts for updates but if not, might be worthwhile moving the sourcing implementations to a common crate to maximise reuse and consistency.

AlexD10S: It makes sense to have the same behaviour for contracts-node. At least the first point of removing the folder when downloading. Opened and issue: https://github.com/r0gue-io/pop-cli/issues/224

AlexD10S (approved):
Looks good to be merged now. 
I've created two new issues to address in separate PRs (https://github.com/r0gue-io/pop-cli/issues/224 and https://github.com/r0gue-io/pop-cli/issues/225).
Additionally, as per your comment, we'll need to refactor the `cliclack` wrapper and integrate it with the rest of the commands.

crates/pop-cli/src/commands/clean.rs
```diff
@@ -0,0 +1,278 @@
+// SPDX-License-Identifier: GPL-3.0
```
AlexD10S, line 1: One small thing before merge, add newline here

evilrobot-01:
> Additionally, as per your comment, we'll need to refactor the `cliclack` wrapper and integrate it with the rest of the commands.

I think we should address the cli parameter on the command as a separate PR first, as it is a dependency rather than command state, before moving all the commands to this type of implementation. I think it will be much easier that way, at least in terms of reviews.

crates/pop-cli/src/commands/clean.rs
evilrobot-01: Done

AlexD10S (approved):

Issue #29, opened 2024-03-06
`pop build parachain`: Export WASM and Genesis files


AlexD10S:
Add an option for `pop build parachain` to export WASM and genesis files (has to generate a chain spec file) after build it.

The steps are [Prepare the parachain collator](https://docs.substrate.io/tutorials/build-a-parachain/connect-a-local-parachain/#prepare-the-parachain-collator):

1. Export the WebAssembly runtime for the parachain: `./target/release/parachain-template-node export-genesis-wasm --chain raw-parachain-chainspec.json para-2000-wasm`
2. Generate a parachain genesis state:` ./target/release/parachain-template-node export-genesis-state --chain raw-parachain-chainspec.json para-2000-genesis-state`

Issue #28, opened 2024-03-06
`pop build parachain`: Generate a chain specification file


AlexD10S:
Add an option for `pop build parachain` to generate a chain spec file after build it.

The steps are [Modify the default chain specification](https://docs.substrate.io/tutorials/build-a-parachain/connect-a-local-parachain/#modify-the-default-chain-specification):

1. Generate the plain text chain specification: `./target/release/parachain-template-node build-spec --disable-default-bootnode > plain-parachain-chainspec.json`
2. Modify the file to change para_id indicated by the user (flag --para_id)
3. Generate a raw chain specification file from the modified chain specification file:` ./target/release/parachain-template-node build-spec --chain plain-parachain-chainspec.json --disable-default-bootnode --raw > raw-parachain-chainspec.json`

Issue #188, opened 2024-05-22
chore:  message showing parachain being build 


AlexD10S:
When building a parachain, the current message is:
```
Pop CLI : Building your parachain
```
Improve the message to display the name of the parachain being built.

Suggested by Bruno here: https://github.com/r0gue-io/pop-cli/issues/74#issuecomment-2048961689

bolajahmad:
How can the CLI tell what the name of the parachain is? The only argument to build is the path (which can be in various form). It's not suitable as a name. Maybe a new argument to `BuildParachainCommand` but what should be the default value?

AlexD10S:
Close with https://github.com/r0gue-io/pop-cli/pull/219

PR #219, opened 2024-06-24
feat(build parachain): generate specification, wasm and genesis state files

AlexD10S:
If the user specifies the `parachain_id` when building the parachain, the process will include the generation of a chain spec file (Closes https://github.com/r0gue-io/pop-cli/issues/28). Additionally, export the WebAssembly runtime for the parachain, and generate a parachain genesis state (Closes https://github.com/r0gue-io/pop-cli/issues/29).
```
pop build parachain -p ../my-parachain --para_id 2000
```
The process for generating the files follows the steps described in the [Substrate Tutorial](https://docs.substrate.io/tutorials/build-a-parachain/connect-a-local-parachain/#modify-the-default-chain-specification). More explanation in the referenced issues above.

The functions to generate the files are independent of the building process in the `pop_parachain` crate, making it easy to create a separate command for file generation if desired, as discussed in https://github.com/r0gue-io/pop-cli/issues/28.

This PR also addresses https://github.com/r0gue-io/pop-cli/issues/188 by displaying the location and name of the built binary.


evilrobot-01:
FYI: https://github.com/paritytech/polkadot-sdk/tree/master/substrate/bin/utils/chain-spec-builder

evilrobot-01 (changes requested):
Some improvements suggested. I think the use of paths needs special attention,

crates/pop-cli/src/commands/build/parachain.rs
```diff
@@ -26,7 +39,30 @@ impl BuildParachainCommand {
...
 
-		outro("Build Completed Successfully!")?;
+		success("Build Completed Successfully!")?;
+		let binary = binary_path(self.path.as_deref())?;
+		let mut generated_files = vec![format!("Binary generated at: \"{binary}\"")];
+		// If para_id is provided, generate the chain spec
+		if let Some(para_id) = self.para_id {
+			let chain_spec = generate_chain_spec(self.path.as_deref(), para_id)?;
+			generated_files
+				.push(format!("New raw chain specification file generated at: {chain_spec}"));
```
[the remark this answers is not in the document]
AlexD10S: Fixed after the refactor to use Path

crates/pop-parachains/src/build.rs
```diff
@@ -1,13 +1,358 @@
...
+/// # Arguments
+///
+/// * `path` - Location of the parachain project.
+/// * `para_id` - The parachain ID to be replaced in the specification.
+pub fn generate_chain_spec(path: Option<&Path>, para_id: u32) -> Result<String, Error> {
+	let parachain_folder = path.unwrap_or(Path::new("./"));
+	let binary_path = binary_path(path)?;
+	let plain_parachain_spec =
+		format!("{}/plain-parachain-chainspec.json", parachain_folder.display());
+	cmd(&binary_path, vec!["build-spec", "--disable-default-bootnode"])
```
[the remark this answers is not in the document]
AlexD10S: Makes sense. Added a simple check: `check_command_exists` and a custom error if fails

crates/pop-parachains/src/build.rs
```diff
@@ -1,13 +1,358 @@
...
+		.and_then(|i| i.as_str())
+		.ok_or_else(|| Error::Config("expected `name`".into()))?;
+	Ok(name.to_string())
+}
+
+/// Get the current parachain id from the generated chain specification file.
+fn get_parachain_id(plain_parachain_spec: &String) -> Result<u32> {
+	let data = fs::read_to_string(plain_parachain_spec)?;
+	let value = serde_json::from_str::<Value>(&data)?;
+	Ok(value.get("para_id").and_then(Value::as_u64).unwrap_or(1000) as u32)
```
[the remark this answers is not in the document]
AlexD10S: Updated with a comment for better clarity.

crates/pop-parachains/src/build.rs
```diff
@@ -1,13 +1,358 @@
...
+	let binary_path = binary_path(path)?;
+	let wasm_file = format!("{}/para-{}-genesis-state", parachain_folder.display(), para_id);
+	cmd(binary_path.clone(), vec!["export-genesis-state", "--chain", &chain_spec, &wasm_file])
+		.run()?;
+	Ok(wasm_file)
+}
+
+/// Parses the node name from the Cargo.toml file located in the project path.
+fn parse_node_name(path: Option<&Path>) -> Result<String, Error> {
+	let cargo_toml = path.unwrap_or(Path::new("./")).join("node/Cargo.toml");
```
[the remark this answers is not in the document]
AlexD10S: Noted for a future refactor.

evilrobot-01 (changes requested):
There are a lot of hard coded values in the library crates which may make its usefulness limited. I suggest a refactor to externalise those values as parameters, so that the calling code (e.g. pop-cli) can control and provide the specifics.

crates/pop-cli/src/commands/build/parachain.rs
```diff
@@ -26,7 +39,43 @@ impl BuildParachainCommand {
...
+		let binary = binary_path(self.path.as_deref())?;
+		let mut generated_files = vec![format!("Binary generated at: \"{}\"", binary.display())];
+		// If para_id is provided, generate the chain spec
+		if let Some(para_id) = self.id {
+			let chain_spec = generate_chain_spec(self.path.as_deref(), para_id)?;
+			generated_files.push(format!(
+				"Plain text chain specification file generated at: {}",
+				chain_spec.display()
+			));
+			let raw_chain_spec = generate_raw_chain_spec(self.path.as_deref())?;
```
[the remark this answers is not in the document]
AlexD10S: Added, with a check that asserts the `chain_spec` exists (in the functions: `generate_raw_chain_spec`, `export_wasm_file`, `generate_genesis_state_file`)

crates/pop-parachains/src/build.rs
```diff
@@ -1,13 +1,418 @@
...
+	Ok(())
+}
+
+/// Get the path to the node release binary based on the project path.
+///
+/// # Arguments
+/// * `path` - Location of the parachain project.
+pub fn binary_path(path: Option<&Path>) -> Result<PathBuf, Error> {
+	let node_name = parse_node_name(path)?;
+	let release_path = path.unwrap_or(Path::new("./")).join("target/release");
```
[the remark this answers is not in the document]
AlexD10S: Removed the hardcoded value from the library function. 
`pop-cli` will now indicate this path.  For now, we are only building parachains in release mode, not debug, so we can assume it will be this path, right?. Code here: https://github.com/r0gue-io/pop-cli/blob/alex/build-generate-files/crates/pop-cli/src/commands/build/parachain.rs#L47

crates/pop-parachains/src/build.rs
```diff
@@ -1,13 +1,418 @@
...
+/// Generates the plain text chain specification for a parachain.
+///
+/// # Arguments
+/// * `path` - Location of the parachain project.
+/// * `para_id` - The parachain ID to be replaced in the specification.
+pub fn generate_chain_spec(path: Option<&Path>, para_id: u32) -> Result<PathBuf, Error> {
+	let parachain_folder = path.unwrap_or(Path::new("./"));
+	let binary_path = binary_path(path)?;
+	check_command_exists(&binary_path, "build-spec")?;
+	let plain_parachain_spec = parachain_folder.join("plain-parachain-chainspec.json");
```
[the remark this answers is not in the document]
AlexD10S: Done

crates/pop-parachains/src/build.rs
```diff
@@ -1,13 +1,418 @@
...
+			&genesis_file.display().to_string(),
+		],
+	)
+	.run()?;
+	Ok(genesis_file)
+}
+
+/// Parses the node name from the Cargo.toml file located in the project path.
+fn parse_node_name(path: Option<&Path>) -> Result<String, Error> {
+	let cargo_toml = path.unwrap_or(Path::new("./")).join("node/Cargo.toml");
```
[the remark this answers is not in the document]
AlexD10S: As the comment above:
Removed the hardcoded value from the library function.
`pop-cli` will now indicate this path. Is it okay for us to assume the folder is `node`?. Code here: https://github.com/r0gue-io/pop-cli/blob/alex/build-generate-files/crates/pop-cli/src/commands/build/parachain.rs#L49

crates/pop-cli/src/commands/build/parachain.rs
```diff
@@ -26,7 +41,65 @@ impl BuildParachainCommand {
...
 		build_parachain(&self.path)?;
 
-		outro("Build Completed Successfully!")?;
+		success("Build completed successfully!")?;
+		let project_path = self.path.clone().unwrap_or(PathBuf::from("./"));
+		// Assumes the target directory is always `target/release`. If an option to build in debug mode is added in `build_parachain`, this will need to be modified accordingly.
+		let target_path = project_path.join("target/release");
+		// Assumes the node directory is always in `node`.
+		let node_path = project_path.join("node");
+		let binary = binary_path(&target_path, &node_path)?;
```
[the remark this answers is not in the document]
evilrobot-01: Likewise I think `build_parachain` should also take a `release: bool` parameter, which pop-cli can default to true for now.

crates/pop-parachains/src/build.rs
```diff
@@ -1,13 +1,498 @@
...
+/// * `chain_spec_file_name` - The name of the chain specification file to be generated.
+/// * `para_id` - The parachain ID to be replaced in the specification.
+pub fn generate_chain_spec(
+	path: Option<&Path>,
+	binary_path: &Path,
+	chain_spec_file_name: &str,
+	para_id: u32,
+) -> Result<PathBuf, Error> {
+	check_command_exists(&binary_path, "build-spec")?;
+	let plain_parachain_spec = path.unwrap_or(Path::new("./")).join(chain_spec_file_name);
```
evilrobot-01, line 54: We dont need these specified as two separate parameters when they can just be specified as a single path.

The command that is being wrapped needs two inputs: `binary: &Path` and `chain_spec: &Path`. And then the `para_id` for the chainspec amendment.

crates/pop-parachains/src/build.rs
```diff
@@ -1,13 +1,498 @@
...
+
+/// Generates a raw chain specification file for a parachain.
+///
+/// # Arguments
+/// * `path` - Location of the parachain project.
+/// * `chain_spec` - Location of the plain chain specification file.
+/// * `binary_path` - The path to the node binary executable that contains the `build-spec` command.
+/// * `chain_spec_file_name` - The name of the chain specification file to be generated.
+pub fn generate_raw_chain_spec(
+	path: Option<&Path>,
```
evilrobot-01, line 71: Can we assume that the raw chain spec will always be output in the same directory as the plain? If so, we dont need this parameter.

`let raw_chain_spec = plain_chain_spec.parent().join(chain_spec_file_name);`

Alternatively take `raw_chain_spec: &Path` and let the caller be specific about where they want to create it.

crates/pop-parachains/src/build.rs
```diff
@@ -1,13 +1,498 @@
...
+		binary_path,
+		vec![
+			"build-spec",
+			"--chain",
+			&plain_chain_spec.display().to_string(),
+			"--disable-default-bootnode",
+			"--raw",
+		],
+	)
+	.stdout_path(raw_chain_spec.clone())
```
evilrobot-01, line 91: Not sure we need this clone based on the PathBuf type of raw_chain_spec and the type defined on stdout_path. Please double check, along with all other usages.

crates/pop-cli/src/commands/build/parachain.rs
```diff
@@ -26,7 +41,65 @@ impl BuildParachainCommand {
...
+				self.path.as_deref(),
+				&chain_spec,
+				&binary,
+				RAW_CHAIN_SPEC_FILE_NAME,
+			)?;
+			generated_files.push(format!(
+				"New raw chain specification file generated at: {}",
+				chain_spec.display()
+			));
+			let wasm_file_name = format!("para-{}-wasm", para_id);
```
evilrobot-01, line 74: Why not .wasm?

https://fileinfo.com/extension/wasm

crates/pop-parachains/src/build.rs
```diff
@@ -1,13 +1,498 @@
...
+
+/// Export the WebAssembly runtime for the parachain.
+///
+/// # Arguments
+/// * `path` - Location of the parachain project.
+/// * `chain_spec` - Location of the raw chain specification file.
+/// * `binary_path` - The path to the node binary executable that contains the `export-genesis-wasm` command.
+/// * `wasm_file_name` - The name of the wasm runtime file to be generated.
+pub fn export_wasm_file(
+	path: Option<&Path>,
```
evilrobot-01, line 104: Again, either take a path for the wasm file, or use the chain_spec location to output in the same directory.

Please also consider the ordering of these parameters. Logically to me I would specify the binary first, as that is used to generate the wasm blob, then the chain spec file, then resulting wasm path. i.e. the same order as those passed to the cmd function within. Same applies for the other functions.

crates/pop-cli/src/commands/build/parachain.rs
```diff
@@ -26,7 +41,65 @@ impl BuildParachainCommand {
...
 
-		outro("Build Completed Successfully!")?;
+		success("Build completed successfully!")?;
+		let project_path = self.path.clone().unwrap_or(PathBuf::from("./"));
+		// Assumes the target directory is always `target/release`. If an option to build in debug mode is added in `build_parachain`, this will need to be modified accordingly.
+		let target_path = project_path.join("target/release");
+		// Assumes the node directory is always in `node`.
+		let node_path = project_path.join("node");
+		let binary = binary_path(&target_path, &node_path)?;
+		let mut generated_files = vec![format!("Binary generated at: \"{}\"", binary.display())];
```
evilrobot-01, line 51: In the resulting output, this path has quotes ("") but all those for the chainspec/wasm artifacts do not. Easiest to just remove them here for consistency.

crates/pop-parachains/src/build.rs
```diff
@@ -1,13 +1,498 @@
...
+};
+use toml_edit::DocumentMut;
 
-/// Build the parachain located in the specified `path`.
-pub fn build_parachain(path: &Option<PathBuf>) -> anyhow::Result<()> {
+/// Build the parachain.
+///
+/// # Arguments
+/// * `path` - Location of the parachain project.
+pub fn build_parachain(path: &Option<PathBuf>) -> Result<(), Error> {
```
evilrobot-01, line 19: As suggested elsewhere, from a library perspective this should take `release: bool` as a parameter

crates/pop-parachains/src/build.rs
```diff
@@ -1,13 +1,498 @@
...
 		.run()?;
+	Ok(())
+}
+
+/// Constructs the node binary path based on the target path and the node folder path.
+///
+/// # Arguments
+/// * `target_path` - The path where the binaries are expected to be found.
+/// * `node_path` - The path to the node from which the node name will be parsed.
+pub fn binary_path(target_path: &Path, node_path: &Path) -> Result<PathBuf, Error> {
```
evilrobot-01, line 31: As suggested elsewhere, this can be eliminated by including the logic in build_parachain to return the built binary path, which is much simpler for the caller. Fine to keep it as a separate private function for testing if you want, its just the current usage is unnecessarily noisy.

crates/pop-parachains/src/build.rs
```diff
@@ -1,13 +1,498 @@
...
+}
+
+/// Generates the plain text chain specification for a parachain.
+///
+/// # Arguments
+/// * `path` - Location of the parachain project.
+/// * `binary_path` - The path to the node binary executable that contains the `build-spec` command.
+/// * `chain_spec_file_name` - The name of the chain specification file to be generated.
+/// * `para_id` - The parachain ID to be replaced in the specification.
+pub fn generate_chain_spec(
```
evilrobot-01, line 47:
```suggestion
pub fn generate_plain_chain_spec(
```

crates/pop-parachains/src/build.rs
```diff
@@ -1,13 +1,498 @@
...
+}
 
+/// Generate the parachain genesis state.
+///
+/// # Arguments
+/// * `path` - Location of the parachain project.
+/// * `chain_spec` - Location of the raw chain specification file.
+/// * `binary_path` - The path to the node binary executable that contains the `export-genesis-state` command.
+/// * `genesis_file_name` - The name of the genesis state file to be generated.
+pub fn generate_genesis_state_file(
```
evilrobot-01, line 134: As above, consider unnecessary parameters and order.

crates/pop-parachains/src/build.rs
```diff
@@ -1,13 +1,498 @@
...
+			&chain_spec.display().to_string(),
+			&genesis_file.display().to_string(),
+		],
+	)
+	.run()?;
+	Ok(genesis_file)
+}
+
+/// Parses the node name from the Cargo.toml file located in the project path.
+fn parse_node_name(node_path: &Path) -> Result<String, Error> {
```
evilrobot-01, line 159: This isnt specific to a node, its reading the package name from a cargo.toml file. This can be moved to `common` as `parse_package_name`.

crates/pop-parachains/src/build.rs
```diff
@@ -1,13 +1,498 @@
...
+	let cargo_toml = node_path.join("Cargo.toml");
```
evilrobot-01, line 160:
```suggestion
	let manifest = node_path.join("Cargo.toml");
```

crates/pop-parachains/src/build.rs
```diff
@@ -1,13 +1,498 @@
...
+		.get("package")
+		.and_then(|i| i.as_table())
+		.and_then(|t| t.get("name"))
+		.and_then(|i| i.as_str())
+		.ok_or_else(|| Error::Config("expected `name`".into()))?;
+	Ok(name.to_string())
+}
+
+/// Get the current parachain id from the generated chain specification file.
+fn get_parachain_id(plain_parachain_spec: &Path) -> Result<u32> {
```
evilrobot-01, line 173:
```suggestion
fn get_parachain_id(chain_spec: &Path) -> Result<u32> {
```
This isnt specific to a plain chain spec - check the json for a raw chain spec and you will see the para_id still exists.

crates/pop-parachains/src/build.rs
```diff
@@ -1,13 +1,498 @@
...
+	let name = config
+		.get("package")
+		.and_then(|i| i.as_table())
+		.and_then(|t| t.get("name"))
+		.and_then(|i| i.as_str())
+		.ok_or_else(|| Error::Config("expected `name`".into()))?;
+	Ok(name.to_string())
+}
+
+/// Get the current parachain id from the generated chain specification file.
```
evilrobot-01, line 172:
```suggestion
/// Get the parachain id from the chain specification file.
```

crates/pop-parachains/src/build.rs
```diff
@@ -1,13 +1,498 @@
...
+fn get_parachain_id(plain_parachain_spec: &Path) -> Result<u32> {
+	let data = fs::read_to_string(plain_parachain_spec)?;
+	let value = serde_json::from_str::<Value>(&data)?;
+	// Default to 2000, as it is the first number allocated for non-system parachains.
+	Ok(value.get("para_id").and_then(Value::as_u64).unwrap_or(2000) as u32)
+}
+
+/// Replaces the generated parachain id in the chain specification file with the provided para_id.
+fn replace_para_id(parachain_folder: PathBuf, para_id: u32, generated_para_id: u32) -> Result<()> {
```
evilrobot-01, line 181: Why 'folder' (parachain_folder)?

crates/pop-parachains/src/build.rs
```diff
@@ -1,13 +1,498 @@
...
+		.ok_or_else(|| Error::Config("expected `name`".into()))?;
+	Ok(name.to_string())
+}
+
+/// Get the current parachain id from the generated chain specification file.
+fn get_parachain_id(plain_parachain_spec: &Path) -> Result<u32> {
+	let data = fs::read_to_string(plain_parachain_spec)?;
+	let value = serde_json::from_str::<Value>(&data)?;
+	// Default to 2000, as it is the first number allocated for non-system parachains.
+	Ok(value.get("para_id").and_then(Value::as_u64).unwrap_or(2000) as u32)
```
evilrobot-01, line 177: Should the caller of the function not make this decision? Looking at the only usage in generate_chain_spec, a para id is already specified. Default values feels like an antipattern to me, masking issues which should be dealt with in a better way.

This function should return Result<Option<u32>> and then the calling code can just do `?.unwrap_or(para_id)`

evilrobot-01 (changes requested):
A few more suggestions/improvements to clean up the code. There seem to be a number of unnecessary parameters in the library crate.

crates/pop-parachains/src/build.rs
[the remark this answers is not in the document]
evilrobot-01: Kinda addressed this in latest review. The library crate shouldnt make this decision, but it should be left up to the caller, in this case the pop-cli crate. I think it is okay to make this assumption in the pop-cli crate for now. That could easily be refactored out into a cli argument in the future if required, without any changes to the library crate.

crates/pop-parachains/src/build.rs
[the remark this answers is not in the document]
evilrobot-01: A library should be as flexible as possible, within constraints, so that it can easily flex with changes in requirements by library users to remain useful. Adding a release parameter is super simple and makes the code way more flexible. Strictly speaking, we should probably have profiles rather than just debug/releease, but can add that later.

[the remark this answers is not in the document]
AlexD10S: Makes sense to have `Profiles` from now, so won't be a breaking change.

crates/pop-parachains/src/build.rs
AlexD10S: This order makes more sense.

crates/pop-parachains/src/build.rs
AlexD10S: removed the pub! and use only internally

crates/pop-common/Cargo.toml
```diff
@@ -19,6 +19,7 @@ serde.workspace = true
 strum.workspace = true
 thiserror.workspace = true
 tokio.workspace = true
+toml_edit.workspace = true
```
evilrobot-01, line 22: I dont think we need this anymore with the manifest module added, using cargo_toml dependency.

crates/pop-common/src/helpers.rs
```diff
@@ -31,12 +32,57 @@ pub fn replace_in_file(file_path: PathBuf, replacements: HashMap<&str, &str>) ->
 	Ok(())
 }
 
+/// Parses the package name from the `Cargo.toml` file located in the specified node path.
+///
+/// # Arguments
+/// * `node_path` - The path to the node directory containing the `Cargo.toml` file.
+pub fn parse_package_name(node_path: &Path) -> Result<String, Error> {
+	let manifest = node_path.join("Cargo.toml");
```
evilrobot-01, line 40: Can be simplified by using crate::manifest::from_path() which returns a strongly type Manifest object which can be used to get the package name in a much simpler way.

Also suggest moving this function to the manifest module. I am personally not a fan of the 'helpers' pattern as it just becomes a dumping ground for unstructured functions, when they could at least be logically grouped by module.

I also question if we even need this function at all now, as getting the package name from the Manifest should be as simple as manifest.package.name (https://docs.rs/cargo_toml/latest/cargo_toml/struct.Package.html#structfield.name)

crates/pop-common/src/helpers.rs
```diff
@@ -51,4 +97,52 @@ mod tests {
...
+	fn parse_package_name_works() -> Result<()> {
+		let temp_dir = tempfile::tempdir()?;
+		let node_path = generate_mock_node(temp_dir.path())?;
+		let name = parse_package_name(&node_path)?;
+		assert_eq!(name, "parachain-template-node");
+		Ok(())
+	}
+
+	#[test]
+	fn parse_package_name_node_cargo_no_exist() -> Result<()> {
```
evilrobot-01, line 111:
```suggestion
	fn parse_package_name_node_cargo_missing() -> Result<()> {
```

crates/pop-parachains/src/build.rs
```diff
@@ -1,30 +1,68 @@
...
+};
 
-/// Build the parachain located in the specified `path`.
+/// Enum representing the build profile for a parachain.
+#[derive(Debug, PartialEq)]
+pub enum Profile {
+	/// Debug profile, optimized for debugging.
+	Debug,
+	/// Release profile, optimized without any debugging functionality.
+	Release,
```
evilrobot-01, line 20: Can add another one for `Other(String)` which can be used for things like the `production` profile.

This should be defined in `pop-common` imo.

crates/pop-cli/src/commands/build/parachain.rs
```diff
@@ -65,23 +116,50 @@ mod tests {
...
 			for release in [false, true] {
 				for valid in [false, true] {
 					let project = if package.is_some() { "package" } else { "parachain" };
-					let mode = if release { "RELEASE" } else { "DEBUG" };
+					let mode = if release { Profile::Release } else { Profile::Debug };
 					let mut cli = MockCli::new()
 						.expect_intro(format!("Building your {project}"))
 						.expect_warning("NOTE: this may take some time...")
-						.expect_info(format!("The {project} was built in {mode} mode."))
+						.expect_info(format!("The {project} was built in {:?} mode.", mode))
```
evilrobot-01, line 162: Is this not a regression? The cargo contract output uses capitalisation of the mode in its output and I didnt see anywhere where this happens to remain consistent?

crates/pop-parachains/src/build.rs
```diff
@@ -1,30 +1,68 @@
...
+use pop_common::{parse_package_name, replace_in_file};
+use serde_json::Value;
+use std::{
+	collections::HashMap,
+	fs,
+	path::{Path, PathBuf},
+};
 
-/// Build the parachain located in the specified `path`.
+/// Enum representing the build profile for a parachain.
```
evilrobot-01, line 14:
```suggestion
/// Enum representing a build profile.
```

crates/pop-parachains/src/build.rs
```diff
@@ -1,30 +1,68 @@
...
+	/// Release profile, optimized without any debugging functionality.
+	Release,
+}
+
+impl Profile {
+	/// Returns the corresponding path to the target folder.
+	fn target_folder(&self, path: &Path) -> PathBuf {
+		match self {
+			Profile::Release => path.join("target/release"),
+			Profile::Debug => path.join("target/debug"),
```
evilrobot-01, line 28: Keep in mind that adding `Other(String)` will result in `Profile::Other(profile) => path.join("target").join(profile)`

crates/pop-parachains/src/errors.rs
```diff
@@ -25,12 +25,18 @@ pub enum Error {
...
+	#[error("Command {command} doesn't exist in binary {binary}")]
+	MissingCommand { command: String, binary: String },
 	#[error("Orchestrator error: {0}")]
 	OrchestratorError(#[from] OrchestratorError),
 	#[error("Failed to create pallet directory")]
 	PalletDirCreation,
 	#[error("ParseError error: {0}")]
 	ParseError(#[from] url::ParseError),
+	#[error("{0}")]
+	PopCommonError(#[from] pop_common::Error),
```
evilrobot-01, line 39:
```suggestion
	CommonError(#[from] pop_common::Error),
```

We dont need to prefix it with Pop, the context of the crate is pop.

evilrobot-01 (changes requested):
Much cleaner!

I think the only thing to address to conclude this is to remove the need for the toml_edit dependency in common, as the common::manifest module provides a much cleaner way of achieving the same result: resolving the package name. You can probably eliminate the function entirely using the Manifest type which is exported from common.

Also think the build profile should be moved to common.

crates/pop-common/src/helpers.rs
AlexD10S: The function is not needed anymore, can be `let node_name = manifest.package().name();` where it was called.

crates/pop-parachains/src/build.rs
AlexD10S: Agreed on moving it into `pop-common.` I think we should only add new profiles once needed, what's the benefit on adding it now?

crates/pop-cli/src/commands/build/parachain.rs
AlexD10S: Implemented a Debug in Profile to keep it consistent

crates/pop-parachains/src/build.rs
AlexD10S: Same as above, I prefer to add new profiles once needed.

crates/pop-common/src/manifest.rs
```diff
@@ -1,19 +1,8 @@
...
 use std::path::{Path, PathBuf};
-use thiserror::Error;
-
-/// An error specific to reading a manifest.
-#[derive(Error, Debug)]
-pub enum Error {
-	#[error("IO error: {0}")]
-	IO(#[from] std::io::Error),
-	#[error("Failed to get manifest path: {0}")]
-	ManifestPath(String),
```
AlexD10S, line 13: Note: Moved with the rest because some are used in other files

crates/pop-common/Cargo.toml
AlexD10S: Removed

evilrobot-01 (changes requested):
Still a few issues sadly. Should be the last one.

crates/pop-common/src/profile.rs
```diff
@@ -0,0 +1,42 @@
...
+		} else {
+			Profile::Debug
+		}
+	}
+}
+
+impl fmt::Display for Profile {
+	fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+		match self {
+			Self::Debug => write!(f, "Debug"),
```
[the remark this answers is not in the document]
AlexD10S: You are right, It uses `RELEASE` and `DEBUG`.

evilrobot-01 (approved):
LGTM! Thank you for yout patience and persistence in getting this completed. 💪

PR #220, opened 2024-06-24
feat(new): improve new contract with a ContractType for categorization

peterwht:


peterwht:
Will refactor `ContractType` and `Provider` types to a reusable type.

crates/pop-cli/src/commands/new/contract.rs
```diff
@@ -1,20 +1,27 @@
...
 #[derive(Args, Clone)]
 pub struct NewContractCommand {
 	#[arg(help = "Name of the contract")]
 	pub(crate) name: Option<String>,
+	#[arg(
+		help = "Contract type.",
+		default_value = ContractType::Examples.as_ref(),
+		value_parser = crate::enum_variants!(ContractType)
+	)]
+	pub(crate) contract_type: Option<ContractType>,
```
AlexD10S, line 24: I received feedback on a pending PR suggesting that we should make the flags shorter. I like it, do you think makes sense to call it only type? https://github.com/r0gue-io/pop-cli/pull/209#discussion_r1650551879

AlexD10S (approved):
LGTM! just left a comment

crates/pop-cli/src/commands/new/contract.rs
peterwht: Unfortunately, type is a reserved keyword in Rust. Could maybe do `category` instead?

AlexD10S: Let's keep it contract_type then, c_type is a bit confusing

PR #201, opened 2024-06-09
feat: guide user for contract creation and add 4 contract templates

AlexD10S:
This PR includes an interactive user guide for contract generation and have an structure to manage different Contract template.

-  Interactive user guide for contract generation when no attributes has been input by the user: `pop new contract`:
Captura de pantalla 2024-06-10 a las 8 56 25

- Support diverse contract templates: 
              1. Standard: The default contract generate with `cargo-contract`
              2. ERC20: The implementation of the ERC-20 standard in Solidity using ink!
              3. ERC721: The implementation of the ERC-721 standard in Solidity using ink!
              4. ERC1155: The implementation of the ERC-1155 standard in Solidity using ink!
All templates were selected from: https://use.ink/examples/smart-contracts and the repository: https://github.com/use-ink/ink-examples/ 

**Considerations for reviewers**
- To fetch a contract template, the steps are as follows:
     1. Clone the [ink-examples](https://github.com/use-ink/ink-examples/ ) repository into a temporary folder.
     2. Copy only the necessary files from the selected template into the contract folder. 
    As indicated in the [code](https://github.com/r0gue-io/pop-cli/blob/alex/contracts-template/crates/pop-contracts/src/new.rs#L83), if future templates include directories, additional functionality will need to be added to support copying these directories.
- Renaming the template files to have the name chosen by the user. The same process that `cargo-contract` uses to rename the standard contract generation:
     1. Change the name in the Cargo.toml file, as shown [here](https://github.com/use-ink/cargo-contract/blob/master/crates/build/templates/new/_Cargo.toml#L2).
     2. Update the name in the code, as demonstrated [here](https://github.com/use-ink/cargo-contract/blob/master/crates/build/templates/new/lib.rs#L4).

- Since we are now generating contracts without using cargo-contract build function, I have added a check for contract naming with the `is_valid_contract_name` function. This function provides the same functionality as `cargo-contract`, as shown [here](https://github.com/use-ink/cargo-contract/blob/master/crates/build/src/new.rs#L39).

- This implementation introduces a basic common crate that includes all `Git` functionality and the `replace_in_file` function to avoid code duplication. In a future PR, we can refactor this crate to include more functions that are not specific to parachain or contracts. For now, it contains only the functionalities used in both crates. Closes https://github.com/r0gue-io/pop-cli/issues/124.

- Currently, the only provider is the official ink repository. If additional providers are added, such as for parachains, only a small amount of extra code will be required.

PR #221, opened 2024-06-25
refactor(templates): Make templates and providers generic

peterwht:
Prereq: https://github.com/r0gue-io/pop-cli/pull/201 <- please merge this first.
This one will update to point to main after the above is merged.

With the introduction of contract generation, with specific types (ERC, Examples, etc), duplicate code was introduced for Template types (such as Provider and ContractTypes and parachain & contract templates).

This PR introduces two traits to simplify template handling:
- `Template` -> the types of a TemplateType.
- `TemplateType` -> the type, category, or "parent" of a `Template`

For example:
`ERC -> TemplateType` has many types `ERC20, ERC721`
`Pop -> TemplateType` has many types `Assets, Contracts, EVM`


crates/pop-cli/src/commands/new/parachain.rs
```diff
@@ -129,7 +132,7 @@ async fn guide_user_to_generate_parachain() -> Result<NewParachainCommand> {
 		decimals: 12,
 		initial_endowment: "1u64 << 60".to_string(),
 	};
-	if template.matches(&Provider::Pop) {
+	if Provider::Pop.matches(&template) {
```
peterwht, line 135: Note, before individual templates were able to use the method `matches` to see if it belongs to a TemplateType; "does template type match template". This PR has flipped this functionality and now matches the template to the template type; "does this template match the template type".

PR #222, opened 2024-06-25
feat: enable building without project type specification

evilrobot-01:
Enables building without needing to specify the project type - i.e. `pop b -r`

TODO:
- [x] formalise project detection - e.g. reading manifest dependencies properly to check for ink
- [x] update `-p` to support package specification, so it can act as a simple wrapper to `cargo b -p`
- [x] tests

### DIFF
--- a/crates/pop-cli/src/cli.rs
+++ b/crates/pop-cli/src/cli.rs
@@ -21,6 +21,10 @@ pub(crate) mod traits {
 		fn outro(&mut self, message: impl Display) -> Result<()>;
 		/// Prints a footer of the prompt sequence with a failure style.
 		fn outro_cancel(&mut self, message: impl Display) -> Result<()>;
+		/// Prints a success message.
+		fn success(&mut self, message: impl Display) -> Result<()>;
+		/// Prints a warning message.
+		fn warning(&mut self, message: impl Display) -> Result<()>;
 	}
 
 	/// A confirmation prompt.
@@ -74,6 +78,16 @@ impl traits::Cli for Cli {
 	fn outro_cancel(&mut self, message: impl Display) -> Result<()> {
 		cliclack::outro_cancel(message)
 	}
+
+	/// Prints a success message.
+	fn success(&mut self, message: impl Display) -> Result<()> {
+		cliclack::log::success(message)
+	}
+
+	/// Prints a warning message.
+	fn warning(&mut self, message: impl Display) -> Result<()> {
+		cliclack::log::warning(message)
+	}
 }
 
 /// A confirmation prompt using cliclack.
@@ -122,6 +136,8 @@ pub(crate) mod tests {
 		multiselect_expectation:
 			Option<(String, Option<bool>, bool, Option<Vec<(String, String)>>)>,
 		outro_cancel_expectation: Option<String>,
+		success_expectations: Vec<String>,
+		warning_expectations: Vec<String>,
 	}
 
 	impl MockCli {
@@ -134,8 +150,8 @@ pub(crate) mod tests {
 			self
 		}
 
-		pub(crate) fn expect_info(mut self, text: impl Display) -> Self {
-			self.info_expectations.push(text.to_string());
+		pub(crate) fn expect_info(mut self, message: impl Display) -> Self {
+			self.info_expectations.push(message.to_string());
 			self
 		}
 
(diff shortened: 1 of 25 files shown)


crates/pop-contracts/src/build.rs
```diff
@@ -1,12 +1,20 @@
...
 
+use crate::utils::helpers::get_manifest_path;
 use contract_build::{execute, BuildMode, ExecuteArgs};
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
+use thiserror::Error;
 
-use crate::utils::helpers::get_manifest_path;
+#[derive(Error, Debug)]
+pub enum Error {
```
AlexD10S, line 9: Is cleaner to keep the list of errors used in this file here instead of the errors.rs file: https://github.com/r0gue-io/pop-cli/pull/222/files#diff-3dd0c55b6cdcc6839cb0e4ff1f593a4335a4c8d2b7132a32d450f47cdb30285b.

crates/pop-contracts/src/build.rs
```diff
@@ -22,3 +30,21 @@ pub fn build_smart_contract(path: &Option<PathBuf>, build_release: bool) -> anyh
 
 	Ok(formatted_result)
 }
+
+/// Determines whether the manifest at the supplied path is a supported smart contract project.
+pub fn is_supported(path: Option<&Path>) -> Result<bool, Error> {
+	// Resolve manifest path
+	let path = match path {
+		Some(path) => match path.ends_with("Cargo.toml") {
```
AlexD10S, line 38: I see you're handling devs specifying the path to `Cargo.toml` like `cargo-contracts` does. I had this doubt at the beginning, and in the `path` description, I specified that it is the path to the project.
```rs
#[arg(short = 'p', long, help = "Path for the contract project, [default: current directory]")]
pub(crate) path: Option<PathBuf>,
```
It's great that you added this to handle both options seamlessly.

evilrobot-01: This is mostly as the `cargo_toml::Manifest::from_path(..) `didnt work when using `"Cargo.toml",` as is output from `get_manifest_path` the path is `None`, but does work for "./Cargo.toml".

crates/pop-contracts/src/build.rs
evilrobot-01: Its tricky. 

By having an error per module, the resulting error perhaps gives greater context as to the source of the error - e.g. Error::BuildError(build::Error::ManifestError(..)) but then there are also circumstances where there might be repetitive errors that are defined in the base Error already.

It just feels like the main Error is becoming a dumping ground for random error types, but maybe that is okay.

brunopgalvao (commented):
Reviewed. LGTM.

crates/pop-contracts/src/build.rs
evilrobot-01: Note that some of this has been reorganised by moving manifest loading to the common crate.

crates/pop-cli/src/commands/build/mod.rs
```diff
@@ -12,18 +19,72 @@ pub(crate) mod parachain;
...
+	pub(crate) path: Option<PathBuf>,
+	/// The package to be built.
+	#[arg(short = 'p', long)]
+	pub(crate) package: Option<String>,
+	/// For production, always build in release mode to exclude debug features.
+	#[clap(short, long)]
+	pub(crate) release: bool,
 }
 
 /// Build a parachain or smart contract.
```
AlexD10S, line 34: Not sure if it makes sense to add something like: "Build a parachain, smart contract, or any Rust project." Technically, you can even build `pop-cli `, pallets and packages with this command. Up to you.

AlexD10S (commented):
Great job. I just left a few small comments. This will have a lot of conflicts with: https://github.com/r0gue-io/pop-cli/pull/219. If you finish the test, I am happy to merge this first and handle the conflicts myself.

There is no need to include it in this PR. But as a refactor we should remove the short = 'p' in the other commands to maintain consistency and avoid confusing the user.
```
#[arg(short = 'p', long)]
path: Option<PathBuf>,
```

And finally, this PR will close https://github.com/r0gue-io/pop-cli/issues/91

crates/pop-cli/src/cli.rs
```diff
@@ -165,6 +181,16 @@ pub(crate) mod tests {
 			self
 		}
 
+		pub(crate) fn expect_success(mut self, message: impl Display) -> Self {
```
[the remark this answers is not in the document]
evilrobot-01: These should be used soon, as we start to add tests for the `pop-cli` crate.

crates/pop-cli/src/commands/build/mod.rs
evilrobot-01: Will change it simply to "Build a project", only as how it currently reads doesnt take features into account and having different comments for the highlevel command based on features seems like overkill.

evilrobot-01: Looking at it now, a lot of the comments suffer the same fate, so maybe I should address it properly so we have a reference to update the rest.

evilrobot-01: Addressed in latest commit.

AlexD10S (approved):
Great!! LGTM

AlexD10S (approved):

PR #226, opened 2024-06-25
refactor(templates): Make templates and providers generic

peterwht:
With the introduction of contract generation, with specific types (ERC, Examples, etc), duplicate code was introduced for Template types (such as Provider and ContractTypes and parachain & contract templates).

This PR introduces two traits to simplify template handling:
- `Template` -> the types of a TemplateType.
- `TemplateType` -> the type, category, or "parent" of a `Template`

For example:
`ERC -> TemplateType` has many types `ERC20, ERC721`
`Pop -> TemplateType` has many types `Assets, Contracts, EVM`


crates/pop-cli/src/commands/new/contract.rs
```diff
@@ -64,8 +67,8 @@ impl NewContractCommand {
 	}
 }
 
-fn is_template_supported(contract_type: &ContractType, template: &Template) -> Result<()> {
-	if !template.matches(contract_type) {
+fn is_template_supported(contract_type: &ContractType, template: &ContractTemplate) -> Result<()> {
+	if !contract_type.matches(template) {
```
peterwht, line 71: Note, before individual templates were able to use the method matches to see if it belongs to a TemplateType; "does template type match template". This PR has flipped this functionality and now matches the template to the template type; "does this template match the template type".

AlexD10S (approved):
Nicely done. 
I'm not completely convinced by the name `TemplateType`, but it seems the best approach to include both providers for `parachains` and categories for `contracts`.

crates/pop-common/src/templates.rs
```diff
@@ -0,0 +1,77 @@
...
+
+	/// Get provider's name.
+	fn name(&self) -> &str {
+		self.get_message().unwrap_or_default()
+	}
+
+	/// Get the default template of the provider.
+	fn default_template(&self) -> T;
+
+	/// Get the providers detailed description message.
```
evilrobot-01, line 59:
```suggestion
	/// Get the type's description.
```

crates/pop-common/src/templates.rs
```diff
@@ -0,0 +1,77 @@
...
+	fn description(&self) -> &str {
+		self.get_detailed_message().unwrap_or_default()
+	}
+
+	/// Get the list of templates of the provider.
```
evilrobot-01, line 64:
```suggestion
	/// Get the list of templates of the type.
```

crates/pop-common/src/templates.rs
```diff
@@ -0,0 +1,77 @@
...
+	fn templates(&self) -> Vec<&T> {
+		T::VARIANTS
+			.iter()
+			.filter(|t| t.get_str(TEMPLATE_TYPE_PROPERTY) == Some(self.name()))
+			.collect()
+	}
+
+	/// Check the template belongs to a template type.
```
evilrobot-01, line 72:
```suggestion
	/// Check the template belongs to a type.
```

crates/pop-contracts/src/lib.rs
```diff
@@ -15,7 +15,7 @@ pub use call::{
 	call_smart_contract, dry_run_call, dry_run_gas_estimate_call, set_up_call, CallOpts,
 };
 pub use new::{create_smart_contract, is_valid_contract_name};
-pub use templates::{ContractType, Template};
+pub use templates::{ContractTemplate, ContractType};
```
evilrobot-01, line 18: As its in the templates namespace, can we not just have `templates::{Contract, Type}` ?

crates/pop-contracts/src/templates.rs
```diff
@@ -75,80 +49,56 @@ impl ContractType {
...
-	/// A minimal contract template.
+pub enum ContractTemplate {
+	/// A minimalist contract template.
 	#[default]
 	#[strum(
 		serialize = "standard",
 		message = "Standard",
 		detailed_message = "ink!'s 'Hello World': Flipper",
-		props(ContractType = "Examples")
+		props(TemplateType = "Examples")
```
evilrobot-01, line 59:
```suggestion
		props(Type = "Examples")
```
Would read much better as this, hence the suggestion to change the const value in pop_common::templates

crates/pop-contracts/src/templates.rs
```diff
@@ -75,80 +49,56 @@ impl ContractType {
...
-		self.get_str("ContractType") == Some(contract_type.name())
-	}
-}
+impl Template for ContractTemplate {}
 
 #[cfg(test)]
 mod tests {
 	use super::*;
 	use std::{collections::HashMap, str::FromStr};
+	use strum::VariantArray;
```
evilrobot-01, line 94: use ContractTemplate::*; to avoid repetition of the enum prefix.

crates/pop-contracts/src/test.rs
```diff
@@ -46,7 +46,7 @@ mod tests {
 		crate::create_smart_contract(
 			"test_contract",
 			temp_contract_dir.as_path(),
-			&crate::Template::Standard,
+			&crate::ContractTemplate::Standard,
```
evilrobot-01, line 49: Do we need a borrow here?

crates/pop-contracts/src/utils/helpers.rs
```diff
@@ -59,7 +59,7 @@ mod tests {
 		crate::create_smart_contract(
 			"test_contract",
 			temp_contract_dir.as_path(),
-			&crate::Template::Standard,
+			&crate::ContractTemplate::Standard,
```
evilrobot-01, line 62: Do we need a borrow here?

crates/pop-contracts/tests/contract.rs
```diff
@@ -13,7 +13,11 @@ fn setup_test_environment() -> std::result::Result<tempfile::TempDir, Error> {
 	let temp_dir = tempfile::tempdir().expect("Could not create temp dir");
 	let temp_contract_dir = temp_dir.path().join("test_contract");
 	fs::create_dir(&temp_contract_dir)?;
-	create_smart_contract("test_contract", temp_contract_dir.as_path(), &Template::Standard)?;
+	create_smart_contract(
+		"test_contract",
+		temp_contract_dir.as_path(),
+		&ContractTemplate::Standard,
```
evilrobot-01, line 19: Do we need a borrow here?

crates/pop-parachains/src/errors.rs
```diff
@@ -49,7 +48,7 @@ pub enum Error {
 	RustfmtError(std::io::Error),
 
 	#[error("Template error: {0}")]
-	TemplateError(#[from] templates::Error),
+	TemplateError(#[from] pop_common::Error),
```
evilrobot-01, line 51:
```suggestion
	TemplateError(#[from] pop_common::templates::Error),
```

crates/pop-parachains/src/lib.rs
```diff
@@ -15,7 +15,7 @@ pub use errors::Error;
 pub use indexmap::IndexSet;
 pub use new_pallet::{create_pallet_template, TemplatePalletConfig};
 pub use new_parachain::instantiate_template_dir;
-pub use templates::{Config, Provider, Template};
+pub use templates::{Config, ParachainTemplate, Provider};
```
evilrobot-01, line 18:
```suggestion
pub use templates::{Config, Parachain, Provider};
```

To simplify

crates/pop-parachains/src/new_parachain.rs
```diff
@@ -19,22 +22,22 @@ use walkdir::WalkDir;
 /// * `tag_version` - version to use (`None` to use latest).
 /// * `config` - customization values to include in the new parachain.
 pub fn instantiate_template_dir(
-	template: &Template,
+	template: &ParachainTemplate,
```
evilrobot-01, line 25: Do we need a borrow here?

crates/pop-parachains/src/new_parachain.rs
```diff
@@ -19,22 +22,22 @@ use walkdir::WalkDir;
...
 	target: &Path,
 	tag_version: Option<String>,
 	config: Config,
 ) -> Result<Option<String>> {
 	sanitize(target)?;
 
-	if template.matches(&Provider::Pop) {
+	if Provider::Pop.matches(&template) {
```
evilrobot-01, line 32:
```suggestion
	if Provider::Pop.provides(&template) {
```

Might be a better name, or `supports`

crates/pop-parachains/src/templates.rs
```diff
@@ -90,15 +66,15 @@ pub struct Config {
...
+pub enum ParachainTemplate {
 	/// Minimalist parachain template.
 	#[default]
 	#[strum(
 		serialize = "standard",
 		message = "Standard",
 		detailed_message = "A standard parachain",
 		props(
-			Provider = "Pop",
+			TemplateType = "Pop",
```
evilrobot-01, line 77: Naming was better before imo. I think just Type might be okay, but a better solution might be to add a const on the trait, with a default of "Type", so that the implementor can define the value to be used.

crates/pop-parachains/src/templates.rs
```diff
@@ -250,31 +202,27 @@ impl Template {
...
-	RepositoryMissing,
-	#[error("The `Provider` property is missing from the template variant")]
-	ProviderMissing,
-}
-
 #[cfg(test)]
 mod tests {
 	use super::*;
 	use std::{collections::HashMap, str::FromStr};
+	use strum::VariantArray;
```
evilrobot-01, line 209: use ParachainTemplate::*; to remove the enum prefix repetition.

crates/pop-parachains/src/templates.rs
```diff
@@ -356,28 +311,36 @@ mod tests {
...
+		assert_eq!(provider.default_template(), ParachainTemplate::ParityContracts);
 	}
 
 	#[test]
 	fn test_templates_of_provider() {
 		let mut provider = Provider::Pop;
 		assert_eq!(
 			provider.templates(),
-			[&Template::Standard, &Template::Assets, &Template::Contracts, &Template::EVM]
+			[
```
evilrobot-01, line 332: Can this function not return owned values so we dont have to borrow for comparison?

crates/pop-cli/src/commands/new/contract.rs
evilrobot-01: No objections apart from the naming. Suggested 'provides' or 'supports' elsewhere in comments.

crates/pop-cli/src/commands/new/parachain.rs
```diff
@@ -35,9 +38,9 @@ pub struct NewParachainCommand {
 		short = 't',
 		long,
 		help = "Template to use.",
-		value_parser = crate::enum_variants!(Template)
+		value_parser = crate::enum_variants!(ParachainTemplate)
```
evilrobot-01, line 41: This enum_variants! macro should be moved to pop_common. Expect it still exists in both contracts and parachains crates.

crates/pop-cli/src/commands/new/parachain.rs
```diff
@@ -129,7 +132,7 @@ async fn guide_user_to_generate_parachain() -> Result<NewParachainCommand> {
 		decimals: 12,
 		initial_endowment: "1u64 << 60".to_string(),
 	};
-	if template.matches(&Provider::Pop) {
+	if Provider::Pop.matches(&template) {
```
evilrobot-01, line 135: Would prefer provides or supports instead of matches.

evilrobot-01 (changes requested):
Great start, left a number of suggestions for improvement.

crates/pop-contracts/src/test.rs
peterwht: Yes, `create_smart_contract` requires a borrowed template.

crates/pop-contracts/src/utils/helpers.rs
peterwht: Yes, `create_smart_contract` requires a borrowed template.

crates/pop-contracts/tests/contract.rs
peterwht: Yes, `create_smart_contract` requires a borrowed template.

crates/pop-common/src/templates.rs
```diff
@@ -0,0 +1,77 @@
...
+	fn repository_url(&self) -> Result<&str, Error> {
+		self.get_str("Repository").ok_or(Error::RepositoryMissing)
+	}
+	/// Get the list of supported templates.
+	fn templates() -> &'static [Self] {
+		Self::VARIANTS
+	}
+
+	/// Get the TemplateType of the template.
+	fn template_type(&self) -> Result<&str, Error> {
```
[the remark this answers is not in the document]
peterwht: Unfortunately, this is not possible because type is a reserved keyword in rust.

crates/pop-parachains/src/new_parachain.rs
peterwht: If a template was just a regular enum, I would say not. However, because templates have a lot of associated information via Strum, I imagine a copy would not be ideal.

crates/pop-parachains/src/templates.rs
peterwht: I did this, but then reverted the change. The reason is that it introduces unnecessary clones in its usage.

crates/pop-cli/src/commands/new/contract.rs
```diff
@@ -8,7 +8,10 @@ use clap::{
 };
 use cliclack::{clear_screen, confirm, input, intro, log::success, outro, outro_cancel, set_theme};
 use console::style;
-use pop_contracts::{create_smart_contract, is_valid_contract_name, ContractType, Template};
+use pop_common::templates::{Template, TemplateType};
```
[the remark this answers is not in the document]
peterwht: I changed from `ContractTemplate` to `Contract` and left `ContractType` as `Type`. Reason being it introduces some extra complexity for imports. In particular, not being able to do `use templates::{Template, Type}` in either the parachain templates, contract templates, or any file that uses the template .

peterwht:
Comments resolved! Seems much better now :)

brunopgalvao (approved):
LGTM.

evilrobot-01 (approved):
LGTM!

PR #231, opened 2024-06-28
feat: liquid standard templates

al3mart:


crates/pop-parachains/src/new_parachain.rs
```diff
@@ -39,37 +35,50 @@ pub fn instantiate_standard_template(
...
+				}),
+		),
+		tag: tag_version.clone(),
+		..Default::default()
+	};
+
+	let template_generation_args = GenerateArgs {
+		template_path: standard_template_path,
+		name: Some(target.file_name().unwrap().to_str().unwrap().to_string()),
+		destination: Some(target.parent().unwrap().to_path_buf()),
```
al3mart, line 64: The type conversions here could maybe be done in a cleaner way ?

crates/pop-parachains/src/utils/helpers.rs
```diff
@@ -85,32 +95,6 @@ pub(crate) fn write_to_file(path: &Path, contents: &str) -> Result<(), Error> {
 #[cfg(test)]
 mod tests {
 	use super::*;
-	use crate::generator::parachain::ChainSpec;
-	use askama::Template;
-	use tempfile::tempdir;
-
-	#[test]
-	fn test_write_to_file() -> Result<(), Box<dyn std::error::Error>> {
```
al3mart, line 93: We don't do this anymore

crates/pop-parachains/src/utils/helpers.rs
```diff
@@ -24,6 +24,15 @@ pub(crate) fn sanitize(target: &Path) -> Result<(), Error> {
 	Ok(())
 }
 
+/// Check if the decimals input by the user is a valid numeric value.
+///
+/// # Arguments
+///
+/// * `decimals` - decimals input to be checked for validity.
+pub fn is_decimals_valid(decimals: &str) -> bool {
```
AlexD10S, line 32: Add a simple unit test for this function

crates/pop-parachains/src/new_parachain.rs
```diff
@@ -80,12 +89,19 @@ mod tests {
 
 	fn setup_template_and_instantiate() -> Result<tempfile::TempDir> {
 		let temp_dir = tempfile::tempdir().expect("Failed to create temp dir");
+		println!("{:?}", temp_dir);
```
AlexD10S, line 92: Remove

crates/pop-parachains/src/new_parachain.rs
```diff
@@ -39,37 +35,50 @@ pub fn instantiate_standard_template(
 	config: Config,
 	tag_version: Option<String>,
 ) -> Result<Option<String>> {
-	let temp_dir = ::tempfile::TempDir::new_in(std::env::temp_dir())?;
-	let source = temp_dir.path();
+	// Template palceholder definitions
```
AlexD10S, line 38: small typo: placeholder

crates/pop-parachains/src/new_parachain.rs
```diff
@@ -39,37 +35,50 @@ pub fn instantiate_standard_template(
...
+	let mut token_symbol: String = "token-symbol=".to_string();
+	let mut token_decimals: String = "token-decimals=".to_string();
+	let mut initial_endowement: String = "initial-endowment=".to_string();
 
-	let tag = Git::clone_and_degit(template.repository_url()?, source, tag_version)?;
+	// Placeholder customization
+	token_symbol.push_str(&*config.symbol);
```
AlexD10S, line 44: Just small comment, this:
```
let mut token_symbol: String = "token-symbol=".to_string();
token_symbol.push_str(&*config.symbol);
```
can be refactor to:
```rs
let token_symbol: String = format!("token-symbol={}", config.symbol);
```

crates/pop-parachains/src/new_parachain.rs
```diff
@@ -39,37 +35,50 @@ pub fn instantiate_standard_template(
...
+	token_decimals.push_str(&*config.decimals);
+	initial_endowement.push_str(&*config.initial_endowment);
 
-	for entry in WalkDir::new(&source) {
-		let entry = entry?;
+	// Tempalte rendering arguments
```
AlexD10S, line 48: Just small typo: Template

crates/pop-parachains/src/new_parachain.rs
```diff
@@ -39,37 +35,50 @@ pub fn instantiate_standard_template(
...
+	let standard_template_path = TemplatePath {
+		git: Some(
+			template
+				.repository_url()
+				.map_or(String::from("https://github.com/r0gue.io/base-parachain"), |r| {
```
AlexD10S, line 53: The `template.repository_url()` should never produce an error because it is our responsibility to set the repository URL correctly.
Not sure the best approach here, personally I'd throw an error instead of hardcoding the `base-parachain` url, or if we don't want to throw an error return a the DEFAULT value in the `repository_url` function(https://github.com/r0gue-io/pop-cli/blob/main/crates/pop-parachains/src/templates.rs#L225, )

crates/pop-parachains/src/new_parachain.rs
```diff
@@ -39,37 +35,50 @@ pub fn instantiate_standard_template(
...
+		destination: Some(target.parent().unwrap().to_path_buf()),
+		define: vec![token_symbol, token_decimals, initial_endowement],
+		..Default::default()
+	};
 
-		let source_path = entry.path();
-		let destination_path = target.join(source_path.strip_prefix(&source)?);
+	// Template rendering
+	let target_template_path = cargo_generate::generate(template_generation_args)
+		.expect("Couldn't render liquid tempalte");
```
AlexD10S, line 71: type: template

AlexD10S (changes requested):
Nice!
About generating the Zombienet file, we can keep the previous work for that no? In the `instantiate_standard_template`:
```rs
// Add network configuration
use askama::Template;
let network = Network { node: "name-we-want".into() };
write_to_file(&target.join("network.toml"), network.render().expect("infallible").as_ref())?;
```
And keep the `generator/parachain.rs` file with:
```
#[derive(Template)]
#[template(path = "base/network.templ", escape = "none")]
pub(crate) struct Network {
	pub(crate) node: String,
}
```

crates/pop-parachains/src/new_parachain.rs
al3mart: Thanks, this makes more sense!

PR #232, opened 2024-06-30
refactor: ensure the UX for new contracts is consistent with the parachains.

AlexD10S:
Ensure the UX for new contracts is consistent with the parachains. Remove the `--path` flag for contracts to match the behavior of parachains. Restructure the user guidance for generating a new contract to follow the same order as parachains: 1. template to use 2. name (path) of the project.

### DIFF
--- a/crates/pop-cli/src/commands/new/contract.rs
+++ b/crates/pop-cli/src/commands/new/contract.rs
@@ -8,8 +8,13 @@ use clap::{
 };
 use cliclack::{clear_screen, confirm, input, intro, log::success, outro, outro_cancel, set_theme};
 use console::style;
+use pop_common::get_project_name_from_path;
 use pop_contracts::{create_smart_contract, is_valid_contract_name, ContractType, Template};
-use std::{env::current_dir, fs, path::PathBuf, str::FromStr};
+use std::{
+	fs,
+	path::{Path, PathBuf},
+	str::FromStr,
+};
 use strum::VariantArray;
 
 #[derive(Args, Clone)]
@@ -24,8 +29,6 @@ pub struct NewContractCommand {
 		value_parser = crate::enum_variants!(ContractType)
 	)]
 	pub(crate) contract_type: Option<ContractType>,
-	#[arg(short = 'p', long, help = "Path for the contract project, [default: current directory]")]
-	pub(crate) path: Option<PathBuf>,
 	#[arg(
 		short = 't',
 		long,
@@ -46,11 +49,13 @@ impl NewContractCommand {
 		} else {
 			self.clone()
 		};
-		let name = &contract_config
+		let path_project = &contract_config
 			.name
 			.clone()
 			.expect("name can not be none as fallback above is interactive input; qed");
-		is_valid_contract_name(name)?;
+		let path = Path::new(path_project);
+		let name = get_project_name_from_path(path, "my_contract");
+		is_valid_contract_name(&name)?;
 		let contract_type = &contract_config.contract_type.clone().unwrap_or_default();
 		let template = match &contract_config.template {
 			Some(template) => template.clone(),
@@ -59,7 +64,7 @@ impl NewContractCommand {
 
 		is_template_supported(contract_type, &template)?;
 
-		generate_contract_from_template(name, contract_config.path, &template)?;
+		generate_contract_from_template(&name, &path, &template)?;
 		Ok(())
 	}
 }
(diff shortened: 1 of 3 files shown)


brunopgalvao (commented):
When I run `cargo run new contract` and select ERC20, for some reason it hangs on:
```
┌   Pop CLI : Generating "duck" using Erc20!
│
◓  Generating contract...   
```

AlexD10S:
> When I run `cargo run new contract` and select ERC20, for some reason it hangs on:
> 
> ```
> ┌   Pop CLI : Generating "duck" using Erc20!
> │
> ◓  Generating contract...   
> ```

Hmmm It works for me. What is doing is cloning the ERC20 from https://github.com/use-ink/ink-examples/tree/main/erc20
Is possible your internet connection is poor?

crates/pop-cli/src/commands/new/contract.rs
```diff
@@ -104,10 +100,14 @@ async fn guide_user_to_generate_contract() -> anyhow::Result<NewContractCommand>
 
 	let template = display_select_options(contract_type)?;
 
+	let name: String = input("Name of your contract?")
+		.placeholder("./my_contract")
```
evilrobot-01, line 104: Note that this is not consistent with https://github.com/r0gue-io/pop-cli/blob/9644bb46b1582972315cc3987f87f274d794916a/crates/pop-cli/src/commands/new/parachain.rs#L126 which uses a hypen, which is the goal of this PR.

crates/pop-common/src/helpers.rs
```diff
@@ -31,6 +31,18 @@ pub fn replace_in_file(file_path: PathBuf, replacements: HashMap<&str, &str>) ->
 	Ok(())
 }
 
+/// Gets the last component (name of a project) of a path or returns a default value if the path has no valid last component.
+///
+/// # Arguments
+/// * `path` - Location path of the project.
+/// * `default` - The default string to return if the path has no valid last component.
+pub fn get_project_name_from_path(path: &Path, default: &str) -> String {
+	path.file_name()
```
evilrobot-01, line 40: path.components().last().map_or() might also be possible. Not sure which is best.

crates/pop-common/src/helpers.rs
```diff
@@ -31,6 +31,18 @@ pub fn replace_in_file(file_path: PathBuf, replacements: HashMap<&str, &str>) ->
 	Ok(())
 }
 
+/// Gets the last component (name of a project) of a path or returns a default value if the path has no valid last component.
+///
+/// # Arguments
+/// * `path` - Location path of the project.
+/// * `default` - The default string to return if the path has no valid last component.
+pub fn get_project_name_from_path(path: &Path, default: &str) -> String {
```
evilrobot-01, line 39: I think you can return a &str here, negating the need to convert into a new string.

crates/pop-common/src/lib.rs
```diff
@@ -4,5 +4,5 @@ pub mod helpers;
 
 pub use errors::Error;
 pub use git::{Git, GitHub, Release};
-pub use helpers::replace_in_file;
+pub use helpers::{get_project_name_from_path, replace_in_file};
 static APP_USER_AGENT: &str = concat!(env!("CARGO_PKG_NAME"), "/", env!("CARGO_PKG_VERSION"));
```
evilrobot-01, line 8:
```suggestion

static APP_USER_AGENT: &str = concat!(env!("CARGO_PKG_NAME"), "/", env!("CARGO_PKG_VERSION"));
```

evilrobot-01 (changes requested):
A few improvements still required.

crates/pop-common/src/helpers.rs
AlexD10S: Seems to produce the same output as `file_name`, so I'll keep it for now.

AlexD10S:
> Do we need to clone it? We have logic for downloading source code archives from GitHub, which is much faster, especially as we don't need the commit history for the created contract?
> 
> Perhaps the sourcing module should be moved from parachains to common?

This is out of scope for this PR. 
I agree that it's an improvement to make. I'm working on it in a different PR: https://github.com/r0gue-io/pop-cli/pull/239.

crates/pop-cli/src/commands/new/contract.rs
AlexD10S: You are right, changed. I'll keep the same you suggested  for consistency. The alternative was: `Where should your contract be created?`

AlexD10S:
> When I run `cargo run new contract` and select ERC20, for some reason it hangs on:
> 
> ```
> ┌   Pop CLI : Generating "duck" using Erc20!
> │
> ◓  Generating contract...   
> ```

Can you try this PR: https://github.com/r0gue-io/pop-cli/pull/239 to see if you still experience this issue?

brunopgalvao:
> cargo run new contract

Same issue for me. This is also happening on `main` for me as well.

brunopgalvao:
> When I run `cargo run new contract` and select ERC20, for some reason it hangs on:
> 
> ```
> ┌   Pop CLI : Generating "duck" using Erc20!
> │
> ◓  Generating contract...   
> ```

This is now working for me. I believe it was an internet issue. I will continue to monitor this in case it happens again and discover the source of the issue.

AlexD10S:
As Frank suggested in a comment above, downloading source code archives from GitHub is much faster than cloning the repository. I have initiated this improvement in another PR: : https://github.com/r0gue-io/pop-cli/pull/239 However, it will require more time for a proper refactor of sourcing logic.

If you encounter the error again, please debug to gather more information to replicate it accurately.

brunopgalvao (approved):

evilrobot-01 (approved):

evilrobot-01:
Note that this is missing the audit warning that we have in the parachains ux, which should be added as a separate PR.

Issue #196, opened 2024-05-30
Build artefacts cannot be found after successful build


bolajahmad:
Running the `pop build..` command does not work as expected.

**Expected result:** Assuming I am in my project directory which has a root Cargo.toml file. When I do `pop build..` command, the resulting build artefact(s) should be in the target directory of my project.

**Observed result:** Build is successful but target folder is not created in my project directory. Instead it is located somewhere else entirely, determined by `CARGO_TARGET_DIR` env. 

To recreate,

* Install pop-cli
* Create a new contracts project and write a contract
* Set CARGO_TARGET_DIR to some other pat
* Build your contract using `pop build contract ..`

We should not be depending on the CARGO_TARGET_DIR for the build of the contract, maybe we can try CARGI_MANIFEST_DIR instead? This configuration should also only be project based, it should not be used by the cli everytime I compile some other project because this is also what is currently happening.

al3mart:
I was able to reproduce, thanks for reporting @bolajahmad

al3mart:
We seem to be pulling the correct manifest path in https://github.com/r0gue-io/pop-cli/blob/7135ef57bcb769a31d17c3202089928318ea6dfb/crates/pop-contracts/src/build.rs#L9
On my tests the manifest was retrieving the one on the expected location `ManifestPath { path: "Cargo.toml" }`

It seems that when executing the build, cargo contracts it just ignores the `manifest_path` given in favor of the system `CARGO_TARGET_DIR` configuration.

Including a local cargo configuration for the project when executing `pop new ...` could a way of avoiding this
https://doc.rust-lang.org/cargo/reference/config.html#hierarchical-structure

bolajahmad:
It's not just cargo contracts though, the same thing happen when I do `pop build parachain` as well. When I do the same build but with only cargo build, it works as expected.

I think we just need a local configuration file that we can read to determine such values like (the target directory, the root manifest TOML file). This also affects issues like #34 and #176 because there could be a build but we can't find it in the anticipated locations.

al3mart:
Yeah, I believe my comment above won't really be of any help either, as cargo will take value of `CARGO_TARGET_DIR` over whatever we might be able to provide in a config for that project

AlexD10S:
> We seem to be pulling the correct manifest path in
> 
[...]

PR #234, opened 2024-07-01
fix: set variable CARGO_TARGET_DIR to build contract in contract project

AlexD10S:
Closes https://github.com/r0gue-io/pop-cli/issues/196

Fixes the bug reported above setting the env variable `CARGO_TARGET_DIR` before building the contract, as `cargo-contract` enforce the take the root_package to place the artifacts when building the contract. See [here](https://github.com/use-ink/cargo-contract/blob/master/crates/build/src/lib.rs#L740) and [here](https://github.com/use-ink/cargo-contract/blob/master/crates/build/src/crate_metadata.rs#L70).

Added a unit test that ensures this variable is set.

### DIFF
--- a/crates/pop-contracts/src/build.rs
+++ b/crates/pop-contracts/src/build.rs
@@ -1,13 +1,15 @@
 // SPDX-License-Identifier: GPL-3.0
 
+use crate::{errors::Error, utils::helpers::get_manifest_path};
 use contract_build::{execute, BuildMode, ExecuteArgs};
-use std::path::PathBuf;
-
-use crate::utils::helpers::get_manifest_path;
+use std::{env, path::PathBuf};
 
 /// Build the smart contract located at the specified `path` in `build_release` mode.
-pub fn build_smart_contract(path: &Option<PathBuf>, build_release: bool) -> anyhow::Result<String> {
-	let manifest_path = get_manifest_path(path)?;
+pub fn build_smart_contract(path: &Option<PathBuf>, build_release: bool) -> Result<String, Error> {
+	let manifest_path = get_manifest_path(&path)?;
+	// Set the CARGO_TARGET_DIR variable to this project target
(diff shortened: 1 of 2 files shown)


al3mart (approved):
Looks great! Thanks :sparkling_heart:

brunopgalvao (commented):
Not sure I follow what @bolajahmad is saying. `CARGO_TARGET_DIR` should be used.  In general, `CARGO_TARGET_DIR` is the location of where to place all generated artifacts, relative to the current working directory. See [here](https://doc.rust-lang.org/cargo/reference/environment-variables.html).

IIU this PR correctly, the reason we want to set the `CARGO_TARGET_DIR` is because if the default is not specified then the directory named `target` located at the root of the workspace will be used. As stated [here](https://doc.rust-lang.org/cargo/reference/config.html#buildtarget-dir).

AlexD10S:
> Not sure I follow what @bolajahmad is saying. `CARGO_TARGET_DIR` should be used. In general, `CARGO_TARGET_DIR` is the location of where to place all generated artifacts, relative to the current working directory. See [here](https://doc.rust-lang.org/cargo/reference/environment-variables.html).
> 
> IIU this PR correctly, the reason we want to set the `CARGO_TARGET_DIR` is because if the default is not specified then the directory named `target` located at the root of the workspace will be used. As stated [here](https://doc.rust-lang.org/cargo/reference/config.html#buildtarget-dir).

In the issue there is a discussion on how to fix it. But in this PR, you understood it correctly the fix is set the `CARGO_TARGET_DIR` variable to don't use the default one when generating the contract. 
What is your concern? I'm open to considering other approaches.

AlexD10S:
Closing it for now. Reason: https://github.com/r0gue-io/pop-cli/issues/196#issuecomment-2202543346

PR #239, opened 2024-07-03
refactor: downloading contracts code instead of cloning

AlexD10S:
For contract templates, we were previously cloning the repository. This PR refactors the process to download source code archives from GitHub instead, as it is faster and cleaner.

To achieve this, a bit of refactoring was needed, moving some logic from the `pop_parachains` crate to the `pop_common` crate, such as the `from_archive` function and the `Status` trait.

_There are a lot of file changes because the `create_smart_contract` function was refactored to be async, which required updating all the related tests._

### DIFF
--- a/crates/pop-common/src/git.rs
+++ b/crates/pop-common/src/git.rs
@@ -296,30 +296,23 @@ pub async fn from_archive(
 	contents: &[(&str, PathBuf)],
 	status: &impl Status,
 ) -> Result<(), Error> {
-	println!("Downloading from {url}...");
-	println!("{:?}", contents);
 	// Download archive
 	status.update(&format!("Downloading from {url}..."));
 	let response = reqwest::get(url).await?.error_for_status()?;
 	let mut file = tempfile()?;
 	file.write_all(&response.bytes().await?)?;
 	file.seek(SeekFrom::Start(0))?;
-	println!("Extracting.");
 	// Extract contents
 	status.update("Extracting from archive...");
 	let tar = GzDecoder::new(file);
-	println!("decoded.");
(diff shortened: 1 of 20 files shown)


crates/pop-common/src/git.rs
```diff
@@ -278,6 +285,37 @@ pub struct Release {
 	pub commit: Option<String>,
 }
 
+/// Source binary by downloading and extracting from an archive.
+///
+/// # Arguments
+/// * `url` - The url of the archive.
+/// * `contents` - The contents within the archive which are required.
+/// * `status` - Used to observe status updates.
+pub async fn from_archive(
```
evilrobot-01, line 294: Why in the git folder and why not just move the whole sourcing module?  This code downloads a source code archive over http from GitHub, and has nothing to do with git.

We now have the same dependencies defined in both places - old location and new.

My suggestion would be to move the whole sourcing module to common as a separate PR, so that both contracts and parachains crates can then use.

AlexD10S: The git file has the implementation of both `Git` and `GitHub` logic, `from_archive` has not been included in the specific Git logic.
The refactor to move the whole sourcing implementation was a bit tricky due to the tight coupling of some logic with parachains. For instance, `from_github_archive`  which sources and builds the binary (https://github.com/r0gue-io/pop-cli/blob/main/crates/pop-parachains/src/up/sourcing.rs#L294). That's why, as an initial step, I decided to migrate only the necessary logic required for contracts.

AlexD10S:
Closing this to redo it after merging this PR: https://github.com/r0gue-io/pop-cli/pull/258.

PR #240, opened 2024-07-03
fix: remove unused folders after download contracts node binary

AlexD10S:
Once the `substrate-contract-node` is downloaded, the binary is extracted to `artifacts/substrate-contracts-node-mac/substrate-contracts-node`. This made the cargo clean cache process more complex as it was failing when trying to remove the folder. 
This PR fixes that by copying the binary to the root cache folder and removing the unnecessary folders.

### DIFF
--- a/crates/pop-contracts/src/utils/contracts_node.rs
+++ b/crates/pop-contracts/src/utils/contracts_node.rs
@@ -4,6 +4,7 @@ use flate2::read::GzDecoder;
 use pop_common::GitHub;
 use std::{
 	env::consts::OS,
+	fs,
 	io::{Seek, SeekFrom, Write},
 	path::PathBuf,
 	process::{Child, Command},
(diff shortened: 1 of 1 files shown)


crates/pop-contracts/src/utils/contracts_node.rs
```diff
@@ -61,6 +62,10 @@ pub async fn run_contracts_node(cache: PathBuf) -> Result<Child, Error> {
 		let tar = GzDecoder::new(file);
 		let mut archive = Archive::new(tar);
 		archive.unpack(cache.clone())?;
+		// Copy the file into the cache folder and remove the folder artifacts
```
evilrobot-01, line 65: As mentioned in another, moving the sourcing module to common means some of this code can be eliminated to reuse the functionality exposed there. Having duplicate code is bad for maintenance.

AlexD10S: Agreed with the refactor, but should be tackled in another PR. This PR is a quick fix to avoid introducing issues in `pop clean` before releasing.

brunopgalvao (approved):

evilrobot-01 (approved):

PR #241, opened 2024-07-04
fix: handle IO error if rename fails

hitchhooker:
image
this should fix there errs. not sure about cause, probably something related to /tmp permissions


AlexD10S:
Thanks for reporting this issue and coming up with a fix! I'm looking into what caused the error and checking if it might affect other parts of the system.

evilrobot-01:
Echoing Alex's comment, thanks for reporting and for the fix!

AlexD10S (approved):

PR #246, opened 2024-07-08
fix: check if contracts needs to be build before deploy

AlexD10S:
A bug was found when running the` pop up contract`. The check to determine if the contract had been built was failing, causing the contract to be rebuilt every time.
I have added the fix and included a test to ensure this issue does not occur again.

### DIFF
--- a/crates/pop-cli/src/commands/up/contract.rs
+++ b/crates/pop-cli/src/commands/up/contract.rs
@@ -3,6 +3,7 @@
 use crate::style::style;
 use clap::Args;
 use cliclack::{clear_screen, confirm, intro, log, log::success, outro, outro_cancel};
+use pop_common::manifest::from_path;
 use pop_contracts::{
 	build_smart_contract, dry_run_gas_estimate_instantiate, dry_run_upload,
 	instantiate_smart_contract, is_chain_alive, parse_hex_bytes, run_contracts_node,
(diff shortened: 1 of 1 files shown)


crates/pop-cli/src/commands/up/contract.rs
```diff
@@ -196,11 +192,20 @@ impl From<UpContractCommand> for UpOpts {
 	}
 }
 
+/// Checks if a contract has been built by verifying the existence of the build directory.
+///
+/// # Arguments
+/// * `path` - An optional path to the project directory. If no path is provided, the current directory is used.
+pub fn has_contract_been_built(path: Option<&Path>) -> bool {
+	path.unwrap_or_else(|| Path::new(".")).join("target/ink").exists()
```
evilrobot-01, line 200: Sorry to say, I dont think this is an accurate enough check. If I delete the contract artifacts within target/ink, leaving the directory, it passes this check and then fails with another error. I think it should look for all the required artifacts and if any one of them is missing, trigger a build.

evilrobot-01 (approved):
Very clean!

crates/pop-cli/src/commands/up/contract.rs
AlexD10S: You are right, checking only the .contract file as is the one used for the `cargo-contract` crate

evilrobot-01 (approved):

PR #250, opened 2024-07-09
refactor: default suri

evilrobot-01:
Ensures consistent with the behaviour in `pop up contract` and simplifies devex with one less parameter to specify.


AlexD10S (approved):
LGTM!  I noticed this during a demo.

PR #257, opened 2024-07-19
feat: pop build spec

al3mart:
Add a new subcommand to build the spec of an existing appchain project.

Users can now run `pop build spec` to build their chain specs (plain and raw) along with some useful artifacts like:
- genesis wasm
- genesis state

---

Lacks tests at the minute.


crates/pop-parachains/src/build.rs
```diff
@@ -72,15 +72,19 @@ fn binary_path(target_path: &Path, node_path: &Path) -> Result<PathBuf, Error> {
...
 ) -> Result<(), Error> {
 	check_command_exists(&binary_path, "build-spec")?;
-	cmd(binary_path, vec!["build-spec", "--disable-default-bootnode"])
-		.stdout_path(plain_chain_spec)
-		.run()?;
+	let mut args = vec!["build-spec"];
+	if default_bootnode {
+		args.push("--disable-default-bootnode");
+	}
+	cmd(binary_path, args).stdout_path(plain_chain_spec).run()?;
```
evilrobot-01, line 87: Check elsewhere for usages on redirecting std_err to avoid the process output messing up pop's output.

crates/pop-parachains/src/build.rs
```diff
@@ -196,6 +200,36 @@ fn replace_para_id(chain_spec: PathBuf, para_id: u32, generated_para_id: u32) ->
 	Ok(())
 }
 
+/// Replaces the generated relay spec name with the given one in an existing chain spec.
+pub fn replace_relay_spec(
```
evilrobot-01, line 204: This approach might be much simpler using serde_json::Value, especially if we are able to read in an existing file to amend in place.

evilrobot-01 (commented):
Looking good, added some suggestions.

crates/pop-cli/src/commands/build/spec.rs
```diff
@@ -0,0 +1,281 @@
...
+		.interact()?;
+
+	// Prompt for relay chain.
+	let mut prompt =
+		cliclack::select("Choose the relay chain your chain will be connecting to: ".to_string());
+	for (i, relay) in RelayChain::VARIANTS.iter().enumerate() {
+		if i == 0 {
+			prompt = prompt.initial_value(relay);
+		}
+		prompt = prompt.item(relay, relay.clone().into_str(), "");
```
evilrobot-01, line 231: We should be using the description from the enum - see templates for examples.

crates/pop-cli/src/commands/build/spec.rs
```diff
@@ -0,0 +1,281 @@
...
+		.placeholder("2000")
+		.default_input("2000")
+		.interact()?;
+
+	// Prompt for relay chain.
+	let mut prompt =
+		cliclack::select("Choose the relay chain your chain will be connecting to: ".to_string());
+	for (i, relay) in RelayChain::VARIANTS.iter().enumerate() {
+		if i == 0 {
+			prompt = prompt.initial_value(relay);
```
evilrobot-01, line 229: Initial value should probably default to some existing value if we are able to read from some existing specified file.

crates/pop-cli/src/commands/build/spec.rs
```diff
@@ -0,0 +1,281 @@
...
+			},
+		};
+		Ok("spec")
+	}
+
+	/// Builds a parachain spec.
+	///
+	/// # Arguments
+	/// * `cli` - The CLI implementation to be used.
+	fn build(self, cli: &mut impl cli::traits::Cli) -> anyhow::Result<&'static str> {
```
AlexD10S, line 115: Does it make sense to check if it's a parachain before attempting the build with the `spec` command? I tried with a contract, and it only showed an error after the entire process was completed.

We have already the method: https://github.com/r0gue-io/pop-cli/blob/main/crates/pop-parachains/src/build.rs#L43 and the example on when is called: https://github.com/r0gue-io/pop-cli/blob/main/crates/pop-cli/src/commands/build/mod.rs#L65

crates/pop-cli/src/commands/build/spec.rs
```diff
@@ -0,0 +1,281 @@
...
+		cli.intro(format!("Building your chain spec"))?;
+
+		// Either a para id was already provided or user has been guided to provide one.
+		let para_id = self.id.unwrap_or(2000);
+
+		// Show warning if specified as deprecated.
+		if !self.valid {
```
AlexD10S, line 122: Agree with Frank's comment above, we don't need this logic.

crates/pop-cli/src/commands/build/spec.rs
```diff
@@ -0,0 +1,281 @@
...
+			None => {
+				cli.info(format!("The binary was not found. The project will be built locally."))?;
+				cli.warning("NOTE: this may take some time...")?;
+				build_parachain(&project_path, None, &mode, None).unwrap()
+			},
+		};
+
+		// Create output dir.
+		let mut output_path = project_path.clone();
+		output_path.push("target/pop");
```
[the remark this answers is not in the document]
AlexD10S: I agree, we can let the user specify where they want to create the output directory, but if not, why not use the `project_path` by default instead of `target/pop`?

crates/pop-cli/src/commands/build/spec.rs
```diff
@@ -0,0 +1,281 @@
...
+		cli.outro(format!(
+			"Need help? Learn more at {}\n",
+			style("https://learn.onpop.io").magenta().underlined()
+		))?;
+
+		Ok("spec")
+	}
+}
+
+/// Guide the user to generate thier chain specification.
```
AlexD10S, line 208: Just a small typo here: Guide the user to generate the chain specification.

crates/pop-cli/src/commands/build/spec.rs
```diff
@@ -0,0 +1,281 @@
...
+	// Prompt for chain type.
+	let mut prompt = cliclack::select("Choose the chain type: ".to_string());
+	for (i, chain_type) in ChainType::VARIANTS.iter().enumerate() {
+		if i == 0 {
+			prompt = prompt.initial_value(chain_type);
+		}
+		prompt = prompt.item(chain_type, chain_type.clone().into_str(), "");
+	}
+	let ct = prompt.interact()?;
+	let chain_type = ct.clone();
```
AlexD10S, line 245: Is not better to use: `let chain_type = prompt.interact()?;`

AlexD10S, line 245: Just a question here, is the `clone` the best way to get the value?

al3mart: The execution of the command requires a the type `RelayChain` or `ChainType` respectively, the interaction method is returning a reference.

Used clone to get me going quickly, but I'll see if I can get this implemented in a cleaner way, didn't love it when I first write it, but at least allowed me to continue.

crates/pop-cli/src/commands/build/spec.rs
al3mart: As for the last commit, it will silently exit without triggering any extra execution if the project is not a supported type.

AlexD10S: I see!  You added the code there already: https://github.com/r0gue-io/pop-cli/pull/257/files#diff-5ab1db70ee18975f80c440aa7fd49171bee41eda0fef6120ac04830d8d1b908fR154 will only be to add an error if is not supported: `cli::Cli.outro_cancel("🚫 Error.")?;`

al3mart: Appreciated :+1:

AlexD10S:
The CI failled in 3 points:
- `cargo deny`: Merge `main` branch, where `openssl` library has been upgraded.
- `cargo test doc`: There has been a change in the code for `generate_plain_chain_spec(&binary_path, &plain_chain_spec_path, para_id);` to fix it simply update the readme here https://github.com/r0gue-io/pop-cli/blob/main/crates/pop-parachains/README.md?plain=1#L47 adding the new parameter. To test it locally you can run `cargo test --doc` in the `pop_parachains` crate.
- parachain integration tests: Update the method is called to build the parachain: https://github.com/r0gue-io/pop-cli/blob/main/crates/pop-cli/tests/parachain.rs#L39 to run `pop build spec` ....

crates/pop-cli/src/commands/build/spec.rs
```diff
@@ -0,0 +1,399 @@
...
+			vec![format!("Specification and artifacts generated at: {}", &output_path.display())];
+		generate_plain_chain_spec(&binary_path, &plain_chain_spec, para_id, self.default_bootnode)?;
+		generated_files.push(format!(
+			"Plain text chain specification file generated at: {}",
+			plain_chain_spec.display()
+		));
+
+		// Customize spec based on input.
+		let relay = self.relay.unwrap_or(RelayChain::PaseoLocal).to_string();
+		let _ = replace_relay_spec(&plain_chain_spec, &relay, "rococo-local");
```
AlexD10S, line 234: Use `?` instead of `_`:
```rs
replace_relay_spec(&plain_chain_spec, &relay, "rococo-local")?;
```

crates/pop-cli/src/commands/build/spec.rs
```diff
@@ -0,0 +1,399 @@
...
+		let chain_type = self.chain_type.unwrap_or(ChainType::Development).to_string();
+		let _ = replace_chain_type(&plain_chain_spec, &chain_type, "Local");
```
AlexD10S, line 236: Same as above

crates/pop-cli/src/commands/build/spec.rs
```diff
@@ -0,0 +1,399 @@
...
+		if self.protocol_id.is_some() {
+			let protocol_id = self.protocol_id.unwrap_or("template-local".to_string());
+			let _ = replace_protocol_id(&plain_chain_spec, &protocol_id, "template-local");
```
AlexD10S, line 239: Same

AlexD10S (approved):

crates/pop-cli/src/commands/build/mod.rs
```diff
@@ -46,6 +48,10 @@ pub(crate) enum Command {
 	#[cfg(feature = "contract")]
 	#[clap(alias = "c")]
 	Contract(BuildContractCommand),
+	/// Build a chain' specification and its genesis artifacts.
```
evilrobot-01, line 51:
```suggestion
	/// Build a chain specification and its genesis artifacts.
```

crates/pop-cli/tests/parachain.rs
```diff
@@ -13,10 +13,10 @@ use tokio::time::{sleep, Duration};
 async fn parachain_lifecycle() -> Result<()> {
 	let temp = tempfile::tempdir().unwrap();
 	let temp_dir = temp.path();
-	// let temp_dir = Path::new("./"); //For testing locally
+	//let temp_dir = Path::new("./"); //For testing locally
 	// Test that all templates are generated correctly
 	generate_all_the_templates(&temp_dir)?;
-	// pop new parachain test_parachain (default)
+	//pop new parachain test_parachain (default)
```
evilrobot-01, line 19:
```suggestion
	// pop new parachain test_parachain (default)
```

crates/pop-cli/tests/parachain.rs
```diff
@@ -13,10 +13,10 @@ use tokio::time::{sleep, Duration};
 async fn parachain_lifecycle() -> Result<()> {
 	let temp = tempfile::tempdir().unwrap();
 	let temp_dir = temp.path();
-	// let temp_dir = Path::new("./"); //For testing locally
+	//let temp_dir = Path::new("./"); //For testing locally
```
evilrobot-01, line 16:
```suggestion
	// let temp_dir = Path::new("./"); //For testing locally
```

crates/pop-cli/src/commands/build/spec.rs
```diff
@@ -0,0 +1,381 @@
...
+	)]
+	Kusama,
+	#[strum(
+		serialize = "kusama-local",
+		message = "Kusama Local",
+		detailed_message = "Local configuration for Kusama network."
+	)]
+	KusamaLocal,
+	#[strum(serialize = "rococo", message = "Rococo", detailed_message = "Parity's test network.")]
+	Rococo,
```
evilrobot-01, line 83: I'm not sure we should include Rococo/Local with its imminent retirement - perhaps replace with Westend/Local so this list is more accurate? Description should maybe read 'Parity's test network for protocol testing' or something.

crates/pop-cli/src/commands/build/spec.rs
```diff
@@ -0,0 +1,381 @@
...
+	Display,
+	EnumString,
+	EnumMessage,
+	ValueEnum,
+	VariantArray,
+	Eq,
+	PartialEq,
+)]
+/// Supported relay chains that can be included in the resulting chain spec.
+pub(crate) enum RelayChain {
```
evilrobot-01, line 69: Can we reconsider the order of the variants please? A suggestion is ordering by likelihood of selection, or the natural progression - i.e. Paseo/Westend to Kusama to Polkadot.

crates/pop-cli/src/commands/build/spec.rs
```diff
@@ -0,0 +1,381 @@
...
+			let genesis_file_name = format!("para-{}-genesis-state", para_id);
+			let genesis_state_file =
+				generate_genesis_state_file(&binary_path, &raw_chain_spec, &genesis_file_name)?;
+			generated_files.push(format!(
+				"Genesis State file exported at: {}",
+				genesis_state_file.display().to_string()
+			));
+		}
+
+		console::Term::stderr().clear_last_lines(1)?;
```
evilrobot-01, line 278:
```suggestion

```

Dont think we need this anymore due to the spinner. It essentially removes the line break under the intro for inconsistent UI.

crates/pop-cli/src/commands/build/spec.rs
```diff
@@ -0,0 +1,381 @@
...
+		Ok("spec")
+	}
+}
+
+/// Guide the user to generate their chain specification.
+async fn guide_user_to_generate_spec() -> anyhow::Result<BuildSpecCommand> {
+	Cli.intro("Generate your chain spec")?;
+
+	// Confirm output path
+	let output_file: String = input("Name of the chain spec file. If a path is given, the necessary directories will be created:")
```
evilrobot-01, line 298:
```suggestion
	let output_file: String = input("Name of the plain chain spec file. If a path is given, the necessary directories will be created:")
```

crates/pop-cli/src/commands/build/spec.rs
```diff
@@ -0,0 +1,381 @@
...
+		}
+		plain_chain_spec.set_extension("json");
+
+		// Locate binary, if it doesn't exist trigger build.
+		let mode: Profile = self.release.into();
+		let cwd = current_dir().unwrap_or(PathBuf::from("./"));
+		let binary_path = match binary_path(&mode.target_folder(&cwd), &cwd.join("node")) {
+			Ok(binary_path) => binary_path,
+			_ => {
+				cli.info(format!("No node was not found. The project will be built locally."))?;
```
evilrobot-01, line 217:
```suggestion
				cli.info(format!("Node was not found. The project will be built locally."))?;
```

crates/pop-cli/src/commands/build/spec.rs
```diff
@@ -0,0 +1,381 @@
...
+			vec![format!("Specification and artifacts generated at: {}", &output_path.display())];
+		generate_plain_chain_spec(&binary_path, &plain_chain_spec, para_id, self.default_bootnode)?;
+		generated_files.push(format!(
+			"Plain text chain specification file generated at: {}",
+			plain_chain_spec.display()
+		));
+
+		// Customize spec based on input.
+		let relay = self.relay.unwrap_or(RelayChain::PaseoLocal).to_string();
+		replace_relay_spec(&plain_chain_spec, &relay, "rococo-local")?;
```
evilrobot-01, line 235: There are no guarantees that it will always be `rococo-local` to replace. I maintain that this should be done properly via serde or some json editing crate.

crates/pop-cli/src/commands/build/spec.rs
```diff
@@ -0,0 +1,381 @@
...
+			prompt = prompt.item(
+				chain_type,
+				chain_type.get_message().unwrap_or(chain_type.as_ref()),
+				chain_type.get_detailed_message().unwrap_or_default(),
+			);
+		}
+		chain_type = prompt.interact()?.clone();
+	}
+
+	// Prompt for default bootnode if chian type is Local or Live.
```
evilrobot-01, line 344:
```suggestion
	// Prompt for default bootnode if chain type is Local or Live.
```

crates/pop-parachains/src/build.rs
```diff
@@ -196,6 +211,63 @@ fn replace_para_id(chain_spec: PathBuf, para_id: u32, generated_para_id: u32) ->
 	Ok(())
 }
 
+/// Replaces the generated relay spec name with the given one in an existing chain spec.
+/// # Arguments
+/// * `chain_spec` - The path to the specification to be modified.
+/// * `relay_name` - The new value for the relay chian field in the specification.
```
evilrobot-01, line 217:
```suggestion
/// * `relay_name` - The new value for the relay chain field in the specification.
```

crates/pop-cli/src/commands/build/spec.rs
```diff
@@ -0,0 +1,381 @@
...
+		generated_files.push(format!(
+			"Plain text chain specification file generated at: {}",
+			plain_chain_spec.display()
+		));
+
+		// Customize spec based on input.
+		let relay = self.relay.unwrap_or(RelayChain::PaseoLocal).to_string();
+		replace_relay_spec(&plain_chain_spec, &relay, "rococo-local")?;
+		let chain_type = self.chain_type.unwrap_or(ChainType::Development).to_string();
+		replace_chain_type(&plain_chain_spec, &chain_type, "Local")?;
```
evilrobot-01, line 237: Likewise, surely the default value depends on what chain_spec.rs outputs, which could vary project to project.

crates/pop-cli/src/commands/build/spec.rs
```diff
@@ -0,0 +1,381 @@
...
+		if self.protocol_id.is_some() {
+			let protocol_id = self.protocol_id.unwrap_or("template-local".to_string());
+			replace_protocol_id(&plain_chain_spec, &protocol_id, "template-local")?;
```
evilrobot-01, line 240: Same again

crates/pop-cli/src/commands/build/spec.rs
```diff
@@ -0,0 +1,381 @@
...
+	}
+}
+
+/// Guide the user to generate their chain specification.
+async fn guide_user_to_generate_spec() -> anyhow::Result<BuildSpecCommand> {
+	Cli.intro("Generate your chain spec")?;
+
+	// Confirm output path
+	let output_file: String = input("Name of the chain spec file. If a path is given, the necessary directories will be created:")
+		.placeholder("./template-chain-spec.json")
```
evilrobot-01, line 299: Should these not be using the constant at the top of this file?

crates/pop-cli/src/commands/build/spec.rs
```diff
@@ -0,0 +1,381 @@
...
+	let relay_chain = prompt.interact()?.clone();
+
+	// Prompt for chain type.
+	// If relay is Kusama or Polkadot, then Live type is used and user is not prompted.
+	let chain_type: ChainType;
+	if relay_chain == RelayChain::Polkadot || relay_chain == RelayChain::Kusama {
+		chain_type = ChainType::Live;
+	} else {
+		let mut prompt = cliclack::select("Choose the chain type: ".to_string());
+		for (i, chain_type) in ChainType::VARIANTS.iter().enumerate() {
```
evilrobot-01, line 331: Should we maybe filter out the chain types based on relay selected? Justification is preventing user from making a mistake.

crates/pop-cli/src/commands/build/spec.rs
```diff
@@ -0,0 +1,381 @@
...
+
+	// Prompt for default bootnode if chian type is Local or Live.
+	let default_bootnode = match chain_type {
+		ChainType::Development => true,
+		_ => confirm(format!("Would you like to use local host as a bootnode ?")).interact()?,
+	};
+
+	// Prompt for protocol-id.
+	let protocol_id: String =
+		input("Choose the protocol-id that will identify your network on the networking layer: ")
```
evilrobot-01, line 352: Think its worth mentioning p2p networking layer?

evilrobot-01 (changes requested):
Great job! Left comments, suggesetions and questions.

I still have reservations on the hacky replacement of values, which seems especially brittle if a dev has customised their chain spec code to default to some other expected value - e.g. paseo-local instead rococo-local which you have hard coded. We have done this in another project for example. It would therefore not update properly, the user wouldnt know and would then have problems later down the line when attempting to use the file and things not working, leading to frustration.

al3mart:
I do share the concerns with the replacements implementations we have right now. I did try moving in the direction suggested and using serde_json::value alst night, and see the benefits over what we have right now.

crates/pop-cli/src/commands/build/spec.rs
al3mart: Makes sense, though keep in mind that mostly everybody testing coretime is doing so in some flavor of Rococo. Although Paseo should integrate coretime rather soon, so I think we can get rid of Rococo anyway.

evilrobot-01: Good point. Maybe we decorate them as deprecated? Both via https://doc.rust-lang.org/reference/attributes/diagnostics.html#the-deprecated-attribute and via the description which appears in the UI.

crates/pop-cli/src/commands/build/spec.rs
al3mart: I included it to get rid of the last message from the spinner as once the generation is complete it doesn't really add much information.
But it is true than when building on release it removes the last break like and it doesn't look too good.

:+1:

crates/pop-cli/src/commands/build/spec.rs
al3mart: It feels better not including it than including them new with a deprecated flag.

crates/pop-cli/src/commands/build/spec.rs
al3mart: Ah, good one, completely overlooked

crates/pop-cli/src/commands/build/spec.rs
```diff
@@ -0,0 +1,423 @@
...
+				cli.info(format!("Node was not found. The project will be built locally."))?;
+				cli.warning("NOTE: this may take some time...")?;
+				build_parachain(&cwd, None, &mode, None)?
+			},
+		};
+
+		// Generate plain spec.
+		spinner.set_message("Generating plain chain specification...");
+		let mut generated_files =
+			vec![format!("Specification and artifacts generated at: {}", &output_path.display())];
```
AlexD10S, line 241: Sorry for another change. But do you think makes sense to remove this message? `format!("Specification and artifacts generated at: {}", &output_path.display())`

I tried `pop build spec` in the parachain, and I see:
```
◆  Generated files:
│  ● Specification and artifacts generated at: .
```
Which seems a bit weird

al3mart: Yeah fair

crates/pop-parachains/src/build.rs
AlexD10S: This can be resolved

crates/pop-cli/src/commands/build/spec.rs
AlexD10S: This can be resolved too no?

AlexD10S (approved):

evilrobot-01 (approved):
Awesome, tysm! I know it was a decent amount of extra work, but hope that you see as much value in the outcome that I do. This seems way more useful and beneficial as a tool to me, so the additional effort is hugely appreciated. Thanks!

crates/pop-cli/src/commands/build/spec.rs
evilrobot-01: This has now been addressed in this PR.

PR #254, opened 2024-07-18
feat(contract-e2e): auto-source substrate-contracts-node with e2e tests

peterwht:
Solves https://github.com/r0gue-io/pop-cli/issues/247

- `pop test contract --features e2e-tests` is now deprecated in favor of `pop test contract --e2e`.
- `pop test contract --e2e` will now download (if needed) `substrate-contracts-node` for usage in e2e tests.
    - If the `substrate-contracts-node` standalone command exists, this will be used.
    - Otherwise, the binary will be downloaded to the pop cache.
- `pop up contract` was refactored to use the existing functionality for sourcing binaries in `pop-parachains`. 

Merge Prerequisite: https://github.com/r0gue-io/pop-cli/pull/258

Review this PR first before #258. #258 moves the sourcing functionality to `pop-common`.

PR #258, opened 2024-07-19
refactor(sourcing): move sourcing functionality to `pop-common`

peterwht:
Sourcing functionality is used by both `pop-parachains` and, now, `pop-contracts` for `substrate-contracts-node` .So, this PR refactors the `Source` and `Binary` structs (and related components) into `pop-common` for better usability and reduced duplication.

Review https://github.com/r0gue-io/pop-cli/pull/254 first. But merge this PR first.


crates/pop-common/src/sourcing/binary.rs
```diff
@@ -0,0 +1,253 @@
+use crate::sourcing::{
```
evilrobot-01, line 1: Missing license. Please double check all new files added.

crates/pop-common/src/sourcing/binary.rs
```diff
@@ -0,0 +1,253 @@
...
+				Archive { .. } | Source::Url { .. } => None,
+			},
+		}
+		.map(|r| r.as_str())
+	}
+}
+
+/// A descriptor of a remote repository.
+#[derive(Debug, PartialEq)]
+pub struct Repository {
```
evilrobot-01, line 182: Debatable on whether this should have been moved as its currently only being used by the parachains crate. Probably fine. Questionable whether it should be in the binary module, apart from the package field which links to a buildable Rust package which _may_ be a binary.

crates/pop-common/src/sourcing/binary.rs
```diff
@@ -0,0 +1,253 @@
...
+			None => crate::GitHub::name(&url)?,
+		}
+		.to_string();
+
+		Ok(Self { url, reference, package })
+	}
+}
+
+/// Trait for observing status updates.
+pub trait Status {
```
evilrobot-01, line 216: This should be in the crate root imo, its a general purpose trait for reporting status updates and not specific to a binary. Likewise the default impl for ().

crates/pop-common/src/sourcing/binary.rs
```diff
@@ -0,0 +1,253 @@
...
+	fn update(&self, status: &str);
+}
+
+impl Status for () {
+	// no-op: status updates are ignored
+	fn update(&self, _: &str) {}
+}
+
+/// Determines the target triple based on the current platform.
+pub fn target() -> Result<&'static str, Error> {
```
evilrobot-01, line 227: Should be in sourcing/mod.rs, or higher.

crates/pop-common/src/sourcing/mod.rs
```diff
@@ -0,0 +1,3 @@
+pub mod binary;
+pub mod errors;
+pub mod sourcing;
```
evilrobot-01, line 3: Think this adds unneccessary complexity/duplication - sourcing::sourcing::*. Suggest moving the content of sourcing.rs to sourcing/mod.rs so its easier to find and easier to navigate.

crates/pop-contracts/Cargo.toml
```diff
@@ -36,7 +36,6 @@ contract-extrinsics.workspace = true
 
 #  pop
 pop-common = { path = "../pop-common", version = "0.2.0" }
-pop-parachains = { path = "../pop-parachains", version = "0.2.0" }
```
evilrobot-01, line 39: wtf

crates/pop-cli/tests/contract.rs
```diff
@@ -43,8 +44,11 @@ async fn contract_lifecycle() -> Result<()> {
 	assert!(temp_dir.join("test_contract/target/ink/test_contract.json").exists());
 
 	// Run the contracts node
-	let cache = temp_dir.join("cache");
-	let process = run_contracts_node(cache, None).await?;
+	// temp_dir gets dropped prematurely, so manually create a tmp directory.
+	let cache = dirs::cache_dir().expect("cache_dir failed").join("pop_tmp");
```
evilrobot-01, line 48: This seems weird as it should only drop when `temp` goes out of scope.

crates/pop-cli/src/commands/up/parachain.rs
```diff
@@ -8,7 +8,7 @@ use cliclack::{
 };
 use console::{Emoji, Style, Term};
 use duct::cmd;
-use pop_common::sourcing::binary::Status;
+use pop_common::sourcing::Status;
```
evilrobot-01, line 11: As suggested elsewhere, Status could/should be moved to root module as its fairly abstract to general purpose status reporting.

crates/pop-common/src/sourcing/errors.rs
```diff
@@ -0,0 +1,23 @@
+// SPDX-License-Identifier: GPL-3.0
```
evilrobot-01, line 1: A whole module for a single enum seems unnecesssary, especially when it is named errors but there is/will probably only ever be a single error type. This pattern has developed from previous implementations and not sure it should be followed. May be subjective however.

evilrobot-01 (changes requested):
Looks great, thanks for taking this on! Have made some suggestions on improving the placement of things, mostly for visibility/accessibility for crate users. Should be quick to address if no objections.

crates/pop-contracts/Cargo.toml
peterwht: This was an intermediate implementation that I used in PR 254.

crates/pop-cli/tests/contract.rs
peterwht: It's because the temp_dir handle gets moved in the subsequent functioned calls, so, although the original handle still exists (when using clone), the clones get dropped which result in the directory getting prematurely cleaned.

Passing references introduces unnecessary complexity in terms of manually specifying lifetimes on the `Binary` struct. So, I felt manually creating a temp_dir is the least overhead for a test.

peterwht:
- Moved `Status` to `pop_common -> lib.rs`
- Moved `Repository` to `pop_common -> git`
- Moved sourcing error to `sourcing -> mod.rs`
- Moved `target()` to `common -> helpers`

AlexD10S (approved):
LGTM!

Issue #260, opened 2024-07-23
avoid linking to substrate.io


kianenigma:
I see that the pallet template links a lot to substrate.io, and I want to suggest you to update to linking to better rust-docs instead of those. 

PaperMoon and a few other areas already in the process of deprecating the website. 

I have worked with Daan already to show that a pallet template can be "backed" by links to polkadot-sdk-docs in a reasonably good way, for example https://github.com/paritytech/polkadot-sdk/pull/4684/files

related https://github.com/paritytech/polkadot-sdk/pull/5118

tsenovilla:
Hey @kianenigma 😄. I've been working in the pallet template these days and just submitted a PR:

-  #261 

In this version there's no links to substrate.io but to the polkadot-sdk-docs. I haven't included a link for everything though, just a link to those docs. I can update it to include some other links and include the same PR with this issue

AlexD10S:
Closed with https://github.com/r0gue-io/pop-cli/pull/272

Issue #253, opened 2024-07-18
Improve template pallet generation


tsenovilla:
Currently, the command 
```bash
pop new pallet
```
is just creating a static template. IMO, this command would be more useful if it provided the user with an interactive interface to design its pallet template according to its needs (eg, how many storage items are needed, does the user wanna a default configuration template for the config trait to be included?, ...). Otherwise, it's almost the same as publishing the template on Github. 

The main point here is to decide where we draw the line. If this interactive interface offers too much, it may become useless. For example, if it forces the user to add all the config types with their trait bounds, there wouldn't be much difference compared to writing it directly on the pallet, and the process will get longer and boring for the user❌. In the other hand, if it just offers the possibility to write the names of those (or some of them) types, the process is more straightforward and the user may be attracted to use the CLI by the fact of creating a template with at least some structure in its config trait, without spending so much time in defining all the trait bounds in the CLI, that can be done later.

IMO, there should be an equilibrium making this tool attractive for the user in terms of time, as well as offering the ability of customizing a little bit the template from the CLI, instead of using a static template. 

I'm already working on a potential solution 😆 will PR it asap!

evilrobot-01:
Agreed, the pallet story definitely needs some love and is next up on our roadmap, starting with adding existing pallets to a runtime. I think you nailed the balance that we strive for, so really looking forward to your PR!

AlexD10S:
Closed with https://github.com/r0gue-io/pop-cli/pull/261

PR #261, opened 2024-07-23
feat: improve pallet template generation

tsenovilla:
# Description

- Related to #253 
- Related to #260

This PR gives a facelift to the command 

```bash
pop new pallet
```

This command was creating a static pallet template, now it's interactive, giving the user the possibility to shape its pallet's template accordingly to its special needs. Here's how the new flow works:

1. The user may choose whether to include a default configuration for its pallet's config trait. If that's selected, the user will be able to decide whether the config trait's types are included in the default config. All the infrastructure needed to use this config will be created, including the ``config_preludes`` module where the types selected to be included in the default may be set up and the derived configuration ready to be edited in the mock runtime.

2. Then, the user may choose among some types that may appear frequently in pallet's. The available options in this first approach are: 

- ```RuntimeEvent```: Nothing special to say here, it's almost everywhere.
- ```RuntimeOrigin```: This type is useful when using custom origins. If this is added to the pallet, the user may also add an internal custom origin later on. If the user doesn't create an internal custom origin, it can still be useful to define some external custom origins.
- ```Currency```: This type is useful if the pallet's going to interact with the native currency of the blockchain. It creates a loosely coupling with the ```Balances``` pallet and all the infrastructure needed to manage it properly.

3. The user may add some other custom types to its config trait. Here we only ask for their names, as adding all the trait bounds would make the CLI an awful tool. So why should the CLI have this option? Basically because the user is able to annotate which of these types are going to be included in the default config and a lot of boilerplate lines are written following these decisions, when it comes to play with this default config. If the user isn't using the default config, this step isn't that useful IMO, but it's up to the user to use it.

4. The user now works with the storage. This step consists simply in picking which type of storage's needed and give it a name. The structure for this type of storage will be autogenerated with the user's choices.

5. The user may choose whether using a genesis state for the pallet.

6. If ```RuntimeOrigin``` has been added, the user may choose to create a custom internal origin. If that's selected, the user will introduce all the variants' names for that enum. This also creates a template for implementing the ```EnsureOrigin``` trait to this custom origin and some template functions that may be used to convert the ```OriginFor``` origin received from the outer world into this custom origin.

## Some other design choices

- I chose to don't include events/errors in this process. The only thing we may ask from the user would be the variants' names of those enums, the user may directly write that in the code. I think this is exactly the kind of things that should be avoided to make this tool attractive.

- I didn't include any template dispatch. This is the custom logic and then including something like 
```rust
fn do_something
```
isn't worthy IMO.

- I created a modular folder, trying to keep as much stuff as possible out of lib.rs/tests.ts. I think it's nice to start building pallets in this way instead of having huge files. Specially for lib.rs, all the stuff which isn't under a
```rust
#[pallet::something]
```
attribute can be moved out.

- I opted for using the new frame umbrella crate in all the templates. 

- Related to #260, links to substrate.io has been replaced by links to polkadot-sdk-docs

# Potential future work

- As I described in #253, the main point here is to find a balance between the time spent creating the pallet and the work we take away from developers, in order to make the use of this CLI attractive. This is just a first approach, based purely in what I found interesting to add, so probably it's far from the ideal.

- Keep adding common config types and their respective stuff to the pallet generation. With the work done in this PR, it's as simple as adding a new enum variant to 
```rust
pub enum TemplatePalletConfigCommonTypes
```
and edit the templates accordingly :)

- Maybe add support for adding external custom origins to the config trait (?) I'm not sure if this would be worth it at all.

- Maybe add support for dispatchables (?). As this is the custom logic it's hard to define it from the CLI, but maybe adding some "structure" to it could be interesting.

- Move the pallet stuff out of `pop-parachains`. I think that it's independent enough to live in its own crate, and the code would be clearer.

# Notes
Many of the modified files only include formatting by `cargo fmt`.


crates/pop-cli/src/commands/utils/helpers.rs
```diff
@@ -0,0 +1,245 @@
+use cliclack::input;
+
+/// A macro to facilitate the selection of an options among some pre-selected ones (represented as
+/// variants from a fieldless enums) and the naming of that selection. This process is repetead
```
brunopgalvao, line 4:
```suggestion
/// variants from fieldless enums) and the naming of that selection. This process is repeated
```

crates/pop-cli/src/commands/utils/helpers.rs
```diff
@@ -0,0 +1,245 @@
...
+/// and give names to those storages.
+///
+/// # Parameters
+///
+/// - `$enum`: The enums types to be iterated over for selection. This enums must implement
+///   `IntoEnumIterator` and `EnumMessage` traits from the `strum` crate. Each variant is
+///   responsible of its own messages. At least one enum type has to be passed in, but several can
+///   be used if for non-conflicting choices may be selected by the user for the same thing.
+/// - `$prompt_message`: The messages displayed to the user at the very beginning of each iteration.
+///   They must implement the `Dispay` trait. Each enum must have associated a prompt message.
```
brunopgalvao, line 18:
```suggestion
///   They must implement the `Display` trait. Each enum must have an associated prompt message.
```

crates/pop-cli/src/commands/utils/helpers.rs
```diff
@@ -0,0 +1,245 @@
...
+                    }),+,
+                    selected_name)
+                );
+            }
+            output
+        }
+    }
+}
+
+/// A macro to facilitate the select multiple variant of an enum and store them inside a Vec.
```
brunopgalvao, line 147:
```suggestion
/// A macro to facilitate the select multiple variant of an enum and store them inside a `Vec`.
```

crates/pop-cli/src/commands/utils/helpers.rs
```diff
@@ -0,0 +1,245 @@
...
+/// - `$enum`: The enum type to be iterated over for the selection. This enum must implement
+///   `IntoEnumIterator` and `EnumMessage` traits from the `strum` crate. Each variant is
+///   responsible of its own messages.
+/// - `$prompt_message`: The message displayed to the user. It must implement the `Dispay` trait.
```
brunopgalvao, line 151:
```suggestion
/// - `$prompt_message`: The message displayed to the user. It must implement the `Display` trait.
```

crates/pop-cli/src/commands/utils/helpers.rs
```diff
@@ -0,0 +1,245 @@
...
+			.interact()?
+			.iter()
+			.map(|byte| unsafe { std::mem::transmute(*byte) })
+			.collect::<Vec<TemplatePalletConfigCommonTypes>>()
+	}};
+}
+
+/// This function performs a loop allowing the user to input as many things as wanted and collecting
+/// them into a `Vec`. In order to stop the loop, the user may enter Quit. # Parameters
+///
```
brunopgalvao, line 232:
```suggestion
/// them into a `Vec`. In order to stop the loop, the user may enter Quit.
///
/// # Parameters
///
```

crates/pop-parachains/templates/pallet/src/config_preludes.rs.templ
```diff
@@ -0,0 +1,139 @@
...
+    type ^^config_type.2^^ = todo!();
+    {{- when TemplatePalletConfigTypesDefault::NoDefaultBounds }}
+    type ^^config_type.2^^ = todo!();
+    {{- else -}}
+    {{- endmatch -}}
+    {{- endfor }}
+} 
+
+/// Default configurations of this pallet in a solochain environment.
+pub struct ParaChainDefaultConfig;
```
brunopgalvao, line 109:
```suggestion
pub struct ParachainDefaultConfig;
```

crates/pop-parachains/templates/pallet/src/config_preludes.rs.templ
```diff
@@ -0,0 +1,139 @@
...
+
+#[derive_impl(frame_system::config_preludes::ParaChainDefaultConfig, no_aggregated_types)]
```
brunopgalvao, line 111:
```suggestion
#[derive_impl(frame_system::config_preludes::ParachainDefaultConfig, no_aggregated_types)]
```

crates/pop-parachains/templates/pallet/src/config_preludes.rs.templ
```diff
@@ -0,0 +1,139 @@
...
+impl frame::deps::frame_system::DefaultConfig for ParaChainDefaultConfig {}
```
brunopgalvao, line 112:
```suggestion
impl frame::deps::frame_system::DefaultConfig for ParachainDefaultConfig {}
```

crates/pop-parachains/templates/pallet/src/config_preludes.rs.templ
```diff
@@ -0,0 +1,139 @@
...
+
+#[register_default_impl(ParaChainDefaultConfig)]
```
brunopgalvao, line 114:
```suggestion
#[register_default_impl(ParachainDefaultConfig)]
```

crates/pop-parachains/templates/pallet/src/config_preludes.rs.templ
```diff
@@ -0,0 +1,139 @@
...
+impl DefaultConfig for ParaChainDefaultConfig{
```
brunopgalvao, line 115:
```suggestion
impl DefaultConfig for ParachainDefaultConfig{
```

crates/pop-parachains/templates/pallet/src/pallet_logic.rs.templ
```diff
@@ -0,0 +1,4 @@
+mod try_state;
```
brunopgalvao, line 1: Please include some documentation in the file so developers know what this file is for and what `mod try_state` is.

crates/pop-parachains/templates/pallet/src/pallet_logic/try_state.rs.templ
```diff
@@ -0,0 +1,9 @@
+use crate::pallet::*;
+
+impl<T: Config> Pallet<T> {
+	#[cfg(feature = "try-runtime")]
+	pub(crate) fn do_try_state() -> Result<(), sp_runtime::TryRuntimeError> {
+		// Write here your try_state logic if needed
+		Ok(())
+	}
+}
```
brunopgalvao, line 9: Please add a reference to Polkadot SDK Docs for this file.

crates/pop-cli/src/commands/new/pallet.rs
```diff
@@ -22,13 +31,52 @@ pub struct NewPalletCommand {
...
-		clear_screen()?;
-		intro(format!(
-			"{}: Generating new pallet \"{}\"!",
-			style(" Pop CLI ").black().on_magenta(),
-			&self.name,
-		))?;
-		set_theme(Theme);
+		Cli.intro("Let's create a new pallet!")?;
+
+		Cli.info("First, we're defining the pallet's config trait.")?;
```
brunopgalvao, line 36:
```suggestion
		Cli.info("First, let's define the pallet's config trait.")?;
```

brunopgalvao (commented):
Thank you for your work.

This work will benefit the community and make Pop CLI more meaningful for development.

My feedback is mainly around DevEx. Let's improve the interactive prompts so that if one is new to pallet development they have some learning resources to go to.

Ideally point to Polkadot SDK Docs as @kianenigma mentioned here: https://github.com/r0gue-io/pop-cli/issues/260

For example:

```
┌   Pop CLI : Let's create a new pallet!
│
⚙  First, we're defining the pallet's config trait.
│  
◇  Would you like to add a derivable default configuration for your pallet's config trait?
│
│    Learn more about config traits here: [Link to Polkadot SDK Docs]
│
│  Yes 
│
◇  Are you interested in adding one of these types and their usual configuration to your pallet? Pick an option by pressing the spacebar. Press enter when you're done!
│  RuntimeEvent 
│  RuntimeOrigin 
│  Currency 
│
⚙  If you want to add some other config types, this is the moment. Keep adding them until you're done!
│
│   Learn more about config types here: [Link to Polkadot SDK Docs]
│  
◇ You're adding a new type to your pallet's Config trait! Should it be included into the metadata?
│
│    Learn more about metadata here: [Link to Polkadot SDK Docs]
│
│  Not included into the metadata 
│
◇  Should it be included in the default configuration?
│
│    Learn more about default configuration here: [Link to Polkadot SDK Docs]
│
│  No 
│
│  test
│
◇  Your adding a new type to your pallet's Config trait! Should it be included into the metadata?
│  Quit 
│
⚙  Now, let's work on your pallet's storage.
│  
◇  You're adding a new storage item to your pallet! Select a storage type to create an instance of it:
│
│  Learn more about pallet storage here: [Link to Polkadot SDK Docs]
│
│  StorageMap 
│
│  mymap
│
◇  You're adding a new storage item to your pallet! Select a storage type to create an instance of it:
│  StorageValue 
│
│  myvalue
│
◇  You're adding a new storage item to your pallet! Select a storage type to create an instance of it:
│  Quit 
│
◇  Would you like to add a genesis state for your pallet?
│  Yes 
│
◇  Would you like to add a custom internal origin? If yes, you'll be asked to add the variants of the Origin enum.
│
│     Learn more about origins here: [Link to Polkadot SDK Docs]
│
│  No 
│
◇  Generation complete
│
└  cd into "pallet-template" and enjoy hacking! 🚀 Don't forget to complete the todo's! 🙈
```

AlexD10S (changes requested):
Fantastic PR, thanks for contributing to `pop-cli`. 
As you pointed out in the issue, we should find the balance between making this tool useful and easy to use for the user. The functionality you added is really cool and very useful. However, I think we should make it even easier for users. Here are my suggestions:

To me, the main thing that helps new devs is having something up and running with one command, similar to what we did for parachains and contracts. With just two commands and no code, users can have their chain up and running. I'd suggest using the same approach here. The generated pallet code should have all `TODOs` removed so that the user can type `pop new pallet`, customize it via the `pop-cli `guide, and then run `pop build` to have it work seamlessly.

From the flow, I would probably move point 1 to the end, as Bruno suggested.  Point 6 should come right after point 2 to prompt if RuntimeOrigin has been added. Point 4 in the flow is very useful for me; the only feedback is if we can somehow ask the user for the type to be stored so we don't have a TODO afterward and it can compile.

Regarding your design points, I disagree with not including any template dispatch. Personally, I find it useful to have a default dispatch as an example of how it works. However, this is just my opinion, and I noticed that the Polkadot documentation has also moved in this direction too: [Polkadot SDK - Your First Pallet](https://paritytech.github.io/polkadot-sdk/master/polkadot_sdk_docs/guides/your_first_pallet/index.html). I'm happy to consider other people's views, but you can keep it as you did for now.

Another point is trying to keep as much as possible out of `lib.rs`. Again, this is just my personal opinion, and I’m open to hearing other views. However, for the basic pallet we will be generating, I’m happy with having everything under lib.rs as I think it’s easier for a new developer.

I also added some comments in the code, mainly to keep it consistent with the rest of the commands. I know it might seem like a lot of feedback and comments, but I'm very excited about merging this functionality. Thanks again for contributing.

crates/pop-parachains/templates/pallet/src/config_preludes.rs.templ
tsenovilla: This wouldn't work right now, frame system's config_preludes module defines `ParaChainDefaultConfig`:frame_system::ParaChainDefaultConfig . And yeah, it also defines `SolochainDefaultConfig`:
frame_system::SolochainDefaultConfig.

It's a bit confusing, but we have to derive in the same way it's defined in the polkadot-sdk, so it's better to keep the same struct name in the template generated config_prelude IMO

tsenovilla:
Following feedback by @brunopgalvao and @AlexD10S (thank u so much for it btw), I'd like to open a discussion about which path should be followed.

There's two points to discuss I think, and the final merged design will depend on what you think it's better:

1. Is this tool thought to be used by everyone, or just by new FRAME developers?

I think we should aim to build a tool useful for both, new and advanced FRAME developers. And honestly, I think that the power of the tool actually lives in advanced users, cause new FRAME developers are probably happy with a static template doing the basic stuff. But there's still some value for them with the basic, dynamic structure.

Related to @AlexD10S feedback, that's why I think moving stuff out `lib.rs` is a good idea. If this is thought just for new users, I agree it's unnecessary. But if someone is including all the more advanced stuff, it's nice to have this structure ready, as keeping it under the `#[frame::pallet]` macro leads to a longer compilation time.

You both suggested to move the default config prompt to the end because it may be frightening for new FRAME devs. I agree with that, but moving it to the end isn't possible at all, since when the user is inputting its config types (in this current version), the prompt will ask whether to include those types in the default configuration depending on this choice. However, we can abstract this thing, putting it into a `clap` command instead of being part of the UX. This will be helpful for advanced users but also new users wouldn't have to think about it. The same applies for the genesis state and the custom origin as well. In addition, this would reduce the prompt time which is nice! 

I haven't considered doing it this way honestly, so thank you for bringing up the discussion with your feedback @brunopgalvao 😁. Regarding the point of prompting the user or not, I still think it may be useful in some cases, specially when we need the user to input the same kind of information several times (config types, storage,...). What you propose about adding commands such as `pop add pallet storage MyValue StorageValue` isn't compatible with prompting in `pop new pallet` at all. I think that this kind of commands may be useful once the pallet is created if you want to add something else (but I'm not sure it's better to directly writing it in the code ), but if you have a first idea of what your pallet should include, why don't offer the user the possibility of putting it into the template with that first command instead of running five times `pop add pallet storage ...`? 

Additionally, I totally agree with you that maybe at the very beginning you don't know at all what to include, and you brainstorm afterwards, but isn't it nice to have a tool allowing us to shape our first thoughts, even if they're wrong? This is the whole point of this PR actually, otherwise taking a static template and build on top of it works perfectly IMO.

2. Do we need to have a functional pallet ready to compile (then no `todo!()`), or should we have a tool taking time away from developers (specially advanced ones)?

This is actually tricky, and I already discussed a bit about it with @AlexD10S on Telegram. In one hand, all the templates out there are ready to run. In the other hand, they are so general and not likely close to what you need.

The React template is a good example: you get something that creates a running page, but it's just the react logo and a counter button. This logic is removed 99'99% of the times cause you don't need it. What is useful about these kind of templates then? Actually the structure, that's what you don't wanna want write yourself. 

IMO, having a non-compiling template isn't that bad deal if its structure is closer to what we need, and to do so we need to keep `todo!()` in the code. 

Think about the storage: if you're planning (on the very first pallet design) to have 5 storage types, would you prefer to create a template with a single, compiling storage value (as in [the current template](https://github.com/r0gue-io/pop-cli/blob/main/crates/pop-parachains/templates/pallet/src/lib.rs.templ#L36)) or to have your 5 storage types already defined, that look like what you need, for example:

```rust 
	/// Some docs for this storage type
	#[pallet::storage]
	pub type MyValue = StorageValue<Value = todo!()>;

	/// Some docs for this storage type
	#[pallet::storage]
	pub type MyMap = StorageMap<todo!(), Key = todo!(), Value = todo!()>;

	/// Some docs for this storage type
	#[pallet::storage]
	pub type AnotherMap = StorageMap<todo!(), Key = todo!(), Value = todo!()>;

	/// Some docs for this storage type
	#[pallet::storage]
	pub type MyDoubleMap = StorageDoubleMap<
		Hasher1 = todo!(),
		Key1 = todo!(),
		Hasher2 = todo!(),
		Key2 = todo!(),
		Value = todo!(),
	>;

	/// Some docs for this storage type
	#[pallet::storage]
	pub type MyNMap = StorageNMap<
		Key = (NMapKey<todo!(), todo!()>, NMapKey<todo!(), todo!()>, NMapKey<todo!(), todo!()>),
		Value = todo!(),
	>;
```

We really need the `todo!()` here if we opt for making it dynamic, IMO asking the user to input all this information breaks the usability of the tool, the user will prefer to write it directly in the code.

Going back to the React template, the dev will surely delete the provided code to write its own code, so why is that better than removing some `todo!()`? 

This also applies to the config trait (and the default config if included) and to all the stuff related to the custom origin.

So this is the discussion I'd like to open, I designed the PR in the way I did having my opinions about this in mind. So now that I explained my point about these decisions, I'd really love to listen your opinions about it. I think the design of this command will completely change depending on your opinions, cause I completely get the points you stated, so now it's time to choose the best path to follow 😋🤪

tsenovilla:
> Thank you for your work.
> 
> This work will benefit the community and make Pop CLI more meaningful for development.
> 
> My feedback is mainly around DevEx. Let's improve the interactive prompts so that if one is new to pallet development they have some learning resources to go to.
> 
> Ideally point to Polkadot SDK Docs as @kianenigma mentioned here: #260
> 
> For example:
> 
> ```
> ┌   Pop CLI : Let's create a new pallet!
> │
> ⚙  First, we're defining the pallet's config trait.
> │  
> ◇  Would you like to add a derivable default configuration for your pallet's config trait?
> │
> │    Learn more about config traits here: [Link to Polkadot SDK Docs]
> │
> │  Yes 
> │
> ◇  Are you interested in adding one of these types and their usual configuration to your pallet? Pick an option by pressing the spacebar. Press enter when you're done!
> │  RuntimeEvent 
> │  RuntimeOrigin 
> │  Currency 
> │
> ⚙  If you want to add some other config types, this is the moment. Keep adding them until you're done!
> │
> │   Learn more about config types here: [Link to Polkadot SDK Docs]
> │  
> ◇ You're adding a new type to your pallet's Config trait! Should it be included into the metadata?
> │
> │    Learn more about metadata here: [Link to Polkadot SDK Docs]
> │
> │  Not included into the metadata 
> │
> ◇  Should it be included in the default configuration?
> │
> │    Learn more about default configuration here: [Link to Polkadot SDK Docs]
> │
> │  No 
> │
> │  test
> │
> ◇  Your adding a new type to your pallet's Config trait! Should it be included into the metadata?
> │  Quit 
> │
> ⚙  Now, let's work on your pallet's storage.
> │  
> ◇  You're adding a new storage item to your pallet! Select a storage type to create an instance of it:
> │
> │  Learn more about pallet storage here: [Link to Polkadot SDK Docs]
> │
> │  StorageMap 
> │
> │  mymap
> │
> ◇  You're adding a new storage item to your pallet! Select a storage type to create an instance of it:
> │  StorageValue 
> │
> │  myvalue
> │
> ◇  You're adding a new storage item to your pallet! Select a storage type to create an instance of it:
> │  Quit 
> │
> ◇  Would you like to add a genesis state for your pallet?
> │  Yes 
> │
> ◇  Would you like to add a custom internal origin? If yes, you'll be asked to add the variants of the Origin enum.
> │
> │     Learn more about origins here: [Link to Polkadot SDK Docs]
> │
> │  No 
> │
> ◇  Generation complete
> │
> └  cd into "pallet-template" and enjoy hacking! 🚀 Don't forget to complete the todo's! 🙈
> ```

I think this is an awesome idea. I'll add it asap! Thanks!

crates/pop-cli/src/commands/new/pallet.rs
```diff
@@ -22,13 +31,52 @@ pub struct NewPalletCommand {
...
+                (TemplatePalletConfigTypesMetadata ,"You're adding a new type to your pallet's Config trait! Should it be included into the metadata?")
+            )
+                .into_iter()
+                .map(|(to_metadata, config_type)| (to_metadata, TemplatePalletConfigTypesDefault::Default, config_type))
+                .collect::<Vec<(TemplatePalletConfigTypesMetadata, TemplatePalletConfigTypesDefault, String)>>();
+		}
+
+		Cli.info("Now, let's work on your pallet's storage.")?;
+
+		let pallet_storage = pick_options_and_give_name!(
```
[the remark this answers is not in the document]
tsenovilla: Agree! But not sure at all how to do it 😅 Any suggestion?

crates/pop-cli/src/commands/utils/helpers.rs
```diff
@@ -0,0 +1,245 @@
+use cliclack::input;
```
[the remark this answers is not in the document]
tsenovilla: Yeah! I'll include that in the next commit :)

crates/pop-cli/src/commands/new/pallet.rs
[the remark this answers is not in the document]
AlexD10S: Checking it the input is a valid variable name will be enough. We can manually add a function that checks: Contains only alphanumeric characters or underscores and does not start with a digit.
Or we can research if there is something already, (for example:  https://stackoverflow.com/questions/58937136/in-a-procedural-macro-how-can-i-check-if-a-string-is-a-valid-variable-name-and)

brunopgalvao:
# Proper Development

I would very much like for Pop CLI to be a tool for **proper** Polkadot development however it also has to be a tool for onboarding new Polkadot developers of varying skill level. With a good amount of work, we can fulfill both scenarios, the latter being required, while the former is more of a nice-to-have but could have the biggest gains longterm.

The base-level target audience for Pop CLI should be a newcomer with intermediate Rust skills and little Web3 experience. Specifically, the bare minimum that we should cater for in regards to pallet development are developers who are new to Polkadot, skimmed through the Docs, have intermediate Rust knowledge, and want to write a basic pallet e.g. Uniswap V1 DEX pallet. But also, based on your wonderful feedback, Pop CLI can also be a powerful tool to accommodate advanced developers who already know more or less what they want and need to spin up something quickly to hit-the-ground running.

## How do we accommodate for both audiences?

**Add ability to elevate privileges to an advanced mode.**
`pop new pallet` accepts an "advanced" mode. Perhaps through a keyboard shortcut (I am thinking of VIM) or by typing in "advanced mode" or "God Mode" (I am thinking of Doom).

With advanced mode, we can create the ideal developer environment for a serious Polkadot developer.
There should be a disclaimer: "You are going off the rails. You are going rogue. This mode is for advanced developers".
Advanced mode should still compile. I understand your argument with the `todo!`'s however I am wondering if there is another way to achieve the same result. 

Perhaps you can solve the problem by creating a custom, universal, "placeholder" macro:
```rust
macro_rules! oDOT {
    () => {{
        println!("TODO: implement this!");
        Default::default()
    }};
}
```

Prompts should still be re-ordered as discussed in previous comments.

**The default however will be "easy mode".**
Easy mode does not need to be selected. It is the default.
The default easy mode pallet should be for educational purposes but powerful enough that it can be used as the base for a basic pallet.
Easy mode will create boilerplate example code (no `todo!()'s`) for each modular component along with inline comments and resource-rich docs that the newcomer can go to for learning purposes.
Easy mode compiles.
Easy mode should be tightly-coupled with the Polkadot SDK Docs.
Easy mode should only have one prompt --> the pallet name and in the near future should accommodate for pallet templates e.g. NFT pallet, DEX Uniswap V1 pallet,  DEX Uniswap V2 pallet, Delegated Staking pallet, etc.

Easy mode should leverage `#[frame_support::pallet(dev_mode)]`.

tsenovilla:
thanks for replying @brunopgalvao ! I really like this approach of having a simple and an advanced mode! However, I'm still not sure that we can have a compiling version in the advanced mode if we want to offer the user the flexibility I gave in the first commit. The `todo!()` appear in places such as the config trait definition

```rust
/// Some docs for this type
#[pallet::constant]
type ConfigType1 = todo!();

/// Some docs for this type
#[pallet::no_default_bounds]
type ConfigType2 = todo!();
```

or inside the definition of storage items


```rust
/// Some docs for this storage type
#[pallet::storage]
pub type MyMap = StorageMap<Hasher = todo!(), Key = todo!(), Value = todo!()>;

/// Some docs for this storage type
#[pallet::storage]
pub type MyDoubleMap = StorageDoubleMap<
Hasher1 = todo!(),
Key1 = todo!(),
Hasher2 = todo!(),
Key2 = todo!(),
Value = todo!(),
>;
```

A macro like the one you proposed won't compile in those places either. The best I can imagine we can do to have a compiling template is to place some default stuff there, eg:

```rust
/// Some docs for this type
#[pallet::constant]
type ConfigType1 = Get<u32>;

/// Some docs for this type
#[pallet::no_default_bounds]
type ConfigType2 = Get<u32>;


...

/// Some docs for this storage type
#[pallet::storage]
pub type MyMap = StorageMap<Hasher = Blake2_128Concat, Key = u32, Value = u32>;

/// Some docs for this storage type
#[pallet::storage]
pub type MyDoubleMap = StorageDoubleMap<
Hasher1 = Blake2_128Concat, Key1 = u32, Hasher2 = Blake2_128Concat, Key2 = u32,Value = u32
>;
```

But this is as far from reality as using `todo!()` and probably will be more confusing. Someone may think: "Why is this template assuming my trait bounds are all `Get<u32>`?", while the `todo!()` is clearer. I'd love to find a way a flexible/compiling advanced mode, but it seems that if we improve one of those aspects the other one is affected, a tradeoff we have to solve.

brunopgalvao:
Why can you not create macro placeholders for the various types that the template generation supports?

tsenovilla:
> Why can you not create macro placeholders for the various types that the template generation supports?

I mean, each `todo!()` place's different.

Inside the `Config` trait it's not possible to expand a macro to define an associated type cause the parser don't accept the symbol !, you'll get an error like this one: 

```bash
error: expected one of `(`, `+`, `::`, `;`, `<`, `=`, or `where`, found `!`
  |
  | trait Config: frame_system::Config{
  |             - while parsing this item list starting here
  |     type A: oDOT!();
  |                 ^ expected one of 7 possible tokens
  | }
```

For storage it's possible to do something like this and this part compiles: 

```rust 

macro_rules! oDOT1{
    () =>{
        Blake2_128Concat
    }
}
macro_rules! oDOT2{
    () => {
        u32
    };
}

...

#[pallet::storage]
pub type MyMap =
StorageMap<Hasher = oDOT1!(), Key = oDOT2!(), Value = oDOT2!()>;
```

But this only works with these simple macros, almost the same as just placing this defaults directly in the template. We should have to create different placeholder macros for several places across the template (and this would make the code very messy IMO), and we still have the problem with the `Config` trait.

Maybe I didn't get your point correctly tho, but if it's something like this I'm not sure that it's a better approach.

crates/pop-parachains/templates/pallet/advanced_mode/src/config_preludes.rs.templ
```diff
@@ -0,0 +1,180 @@
+{{- let using_runtime_event = pallet_common_types|contains(TemplatePalletConfigCommonTypes::RuntimeEvent) -}}
```
kianenigma, line 1: I've pinged @gupnik to give feedback on this, but he is out for another week or so :)

crates/pop-parachains/templates/pallet/simple_mode/src/lib.rs.templ
```diff
@@ -0,0 +1,189 @@
+//! # ^^name|capitalize^^ Pallet
```
kianenigma, line 1: The links here look good, thanks for taking in the feedback! 

I about all the pallet code that gets generated via templates, what would be cool is to have integrity tests that generates all possible pallets based on the `.template` files, and make sure they compile. I would guess you already have this, but double checking :)

crates/pop-parachains/templates/pallet/advanced_mode/src/mock.rs.templ
```diff
@@ -0,0 +1,174 @@
+{{- let using_runtime_event = pallet_common_types|contains(TemplatePalletConfigCommonTypes::RuntimeEvent) -}}
+{{- let using_runtime_origin = pallet_common_types|contains(TemplatePalletConfigCommonTypes::RuntimeOrigin) -}}
+{{- let using_currency = pallet_common_types|contains(TemplatePalletConfigCommonTypes::Currency) -}}
+use crate::{self as pallet_^^name|lower^^, *};
+use pop_dev::{PlaceholderType, pop_todo};
+
+use frame::{
+	deps::frame_support::{
```
kianenigma, line 8: See `Usage section here https://paritytech.github.io/polkadot-sdk/master/polkadot_sdk_frame/index.html

it is all about `preludes` :) 

and whatever is not in prelude, but reasonable should be: 

1. temporarily fetch it via `dpes`
2. either PR or, make an issue for us to fix. 

In general we should try and avoid using `deps`, if that makes sense.

crates/pop-parachains/templates/pallet/advanced_mode/src/tests.rs.templ
```diff
@@ -0,0 +1,7 @@
+use crate::{mock::*, *};
+use frame::deps::frame_support::{assert_noop, assert_ok};
```
kianenigma, line 2: the `testing::prelude` should cover this.

crates/pop-parachains/templates/pallet/advanced_mode/src/benchmarking.rs.templ
```diff
@@ -0,0 +1,16 @@
+use super::*;
+
+#[allow(unused)]
+use crate::{frame_system::RawOrigin, Pallet, mock::{new_test_ext, Test}};
+use frame::deps::frame_benchmarking::v2::*;
```
kianenigma, line 5: not sure if we have one, but we should make a `frame::benchmarking::prelude`, if a good one doesn't exist. 

Should have been part of https://github.com/paritytech/polkadot-sdk/pull/4406/files

kianenigma (commented):
Good luck guys, and please feel free to come to me, @gupnik and Oliver for support.

crates/pop-parachains/templates/pallet/advanced_mode/src/mock.rs.templ
tsenovilla:
```rust 
use frame::{prelude::*, runtime::prelude::*, testing_prelude::*, deps::frame_support::runtime};
```

compiles with `cargo test`, but it doesn't without the last import. It seems the `runtime` isn't in the preludes

crates/pop-parachains/templates/pallet/advanced_mode/src/tests.rs.templ
tsenovilla: Committed!

tsenovilla:
Ok so after merging 

- #272 

I added the advanced mode to this PR. Now, the customizable template is only available in that mode, and the prompts have mostly been changed by commands. They are still there when the user is repeatedly asked to introduce something (eg, the user uses the flag -s to specify storage so he's prompted to define the pallet's storage).

Following a previous discussion with @brunopgalvao, the template is now compiling by using some placeholders that lives in a new crate called `pop-dev`, so the template imports and uses them. This crate isn't merged of course, so to test this functionality you'd need to change the `Cargo.toml` of the template to point to your local folder.


IMO, this is messier than the version using the `todo!()` macro: this advanced mode should be used by advanced developers, so those imports feel more of a hindrance than anything else. It's something that will be removed anyway. I'd prefer to find the non-compiling template being neater. Anyway, going back to the `todo!()` is easy in case we want to do so, the important thing here is the advanced mode

crates/pop-parachains/templates/pallet/simple_mode/src/lib.rs.templ
tsenovilla: I'd say that's not implemented, so it could be a nice issue for another PR

crates/pop-parachains/templates/pallet/advanced_mode/src/benchmarking.rs.templ
tsenovilla: It seems there's not a `benchmarking` module in the frame crate yet: https://paritytech.github.io/polkadot-sdk/master/polkadot_sdk_frame/index.html?search=benchmarking

tsenovilla: I have to look at this a little bit more, `frame::benchmarking::prelude` isn't compiling, but maybe there's another way without using deps.

crates/pop-parachains/templates/pallet/advanced_mode/src/mock.rs.templ
kianenigma: This is the PR that you are missing: 

https://github.com/paritytech/polkadot-sdk/pull/5151

Hopefully you will have it backported soon enough :)

crates/pop-parachains/src/new_pallet.rs
```diff
@@ -2,55 +2,98 @@
...
 /// Metadata for the Template Pallet.
+#[derive(Debug)]
 pub struct TemplatePalletConfig {
+	/// The name of the pallet
 	pub name: String,
+	/// The authors of the pallet
 	pub authors: String,
+	/// The pallet description
 	pub description: String,
+	/// A `bool` indicating if the pallet is contained in a workspace
```
AlexD10S, line 28: I would remove:  `A `bool` indicating`, an example of using bool flags in `pop build contract`: https://github.com/r0gue-io/pop-cli/blob/main/crates/pop-cli/src/commands/build/contract.rs#L18
Suggestion:
```rs
/// The pallet is contained in a workspace
pub pallet_in_workspace: bool,
/// Use the advanced mode
pub pallet_advanced_mode: bool,
/// Include a default config for the pallet.
pub pallet_default_config: bool,
/// Include a genesis config
pub pallet_genesis: bool,
/// Include a custom origin
pub pallet_custom_origin: bool,
```

crates/pop-parachains/src/new_pallet.rs
```diff
@@ -2,55 +2,98 @@
...
 	File::create(format!("{}/Cargo.toml", pallet.display()))?;
 	File::create(format!("{}/lib.rs", src.display()))?;
 	File::create(format!("{}/benchmarking.rs", src.display()))?;
 	File::create(format!("{}/tests.rs", src.display()))?;
 	File::create(format!("{}/mock.rs", src.display()))?;
-	File::create(format!("{}/weights.rs", src.display()))?;
+	if config.pallet_advanced_mode {
+		create_dir(&pallet_logic)?;
+		create_dir(&tests)?;
+		File::create(format!("{}/pallet_logic.rs", src.display()))?;
```
AlexD10S, line 84: `pallet_logic` and `try_state` files I am not sure if makes sense to add them.
@brunopgalvao  any thoughts?

crates/pop-parachains/templates/pallet/advanced_mode/src/tests.rs.templ
```diff
@@ -0,0 +1,6 @@
+use crate::{*, mock::*};
```
AlexD10S, line 1: Does make sense to have the tests like the polkadot-sdk docs: https://paritytech.github.io/polkadot-sdk/master/polkadot_sdk_docs/guides/your_first_pallet/index.html#your-first-test-runtime

@brunopgalvao

crates/pop-cli/src/commands/new/pallet.rs
```diff
@@ -20,23 +29,70 @@ pub struct NewPalletCommand {
...
 
+#[derive(Subcommand)]
+pub enum Mode {
+	/// Using the advanced mode will unlock all the POP CLI potential. You'll be able to fully customize your pallet template!. Don't use this mode unless you exactly know what you want for your pallet
+	Advanced(AdvancedMode),
+}
+
+#[derive(Args)]
+pub struct AdvancedMode {
+	#[arg(short, long, value_enum, help = "Add common types to your config trait from the CLI.")]
```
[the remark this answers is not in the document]
AlexD10S: This is not needed anymore! You fixed it

crates/pop-cli/src/commands/new/pallet.rs
```diff
@@ -20,23 +29,70 @@ pub struct NewPalletCommand {
...
+	pub(crate) config_common_types: Option<Vec<TemplatePalletConfigCommonTypes>>,
+	#[arg(short, long, help = "Use a default configuration for your config trait.")]
+	pub(crate) default_config: bool,
+	#[arg(short, long, value_enum, help = "Add storage items to your pallet from the CLI.")]
```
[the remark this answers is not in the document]
AlexD10S: Not needed anymore, you fixed

crates/pop-cli/src/commands/new/pallet.rs
```diff
@@ -1,15 +1,24 @@
...
+use pop_parachains::{
+	create_pallet_template, resolve_pallet_path, TemplatePalletConfig,
+	TemplatePalletConfigCommonTypes, TemplatePalletStorageTypes,
+};
+use std::{fs, process::Command};
+use strum::{EnumMessage, IntoEnumIterator};
 
 #[derive(Args)]
 pub struct NewPalletCommand {
+	#[command(subcommand)]
```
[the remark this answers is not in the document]
tsenovilla: So I'm not expecting new modes to be added, but using a subcommand allows us to include the related advanced commands only if the option `advanced` is specified. Using a bool would be also possible with a refactor I think, but this way looks better IMO

crates/pop-parachains/src/new_pallet.rs
tsenovilla: This is just to create a modular template, so the advanced developer doesn't have to separate its logic from a huge lib.rs file. I think this is nice but can be reverted if you don't like it

crates/pop-parachains/templates/pallet/advanced_mode/src/tests.rs.templ
tsenovilla: Same as above, creating a test submodule separated from lib.rs is nice when you have plenty of tests. I'd keep it as it's in the PR right now, but can be changed if you prefer the other way

crates/pop-common/src/manifest.rs
```diff
@@ -61,25 +66,26 @@ pub fn add_crate_to_workspace(workspace_toml: &Path, crate_path: &Path) -> anyho
...
+					_ => false,
+				});
+				if !already_in_array {
+					array.push(crate_relative_path);
 				}
 			} else {
-				workspace_table["members"] = value(
-					crate_relative_path.to_str().expect("target's always a valid string; qed"),
-				);
+				return Err(anyhow::anyhow!("Corrupted workspace"));
```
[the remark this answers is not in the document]
tsenovilla: Maybe open a different issue to address this??

crates/pop-cli/src/commands/new/pallet.rs
```diff
@@ -20,23 +29,70 @@ pub struct NewPalletCommand {
 	pub(crate) path: Option<String>,
 }
 
+#[derive(Subcommand)]
+pub enum Mode {
+	/// Using the advanced mode will unlock all the POP CLI potential. You'll be able to fully customize your pallet template!. Don't use this mode unless you exactly know what you want for your pallet
```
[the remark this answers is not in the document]
brunopgalvao:
```suggestion
	/// Advanced mode enables detailed customization of pallet development.
```

crates/pop-parachains/src/new_pallet.rs
brunopgalvao: I do like the modular approach.
I would like @kianenigma or @gupnik to chime in here as to what we would like to see Polkadot developers using. Modular approach or something else?

crates/pop-parachains/templates/pallet/advanced_mode/src/tests.rs.templ
brunopgalvao: ditto as my prev comment.

brunopgalvao:
1. The `--help` prompt order should be consistent with the order of how the commands (`OPTIONS`, `NAME` `COMMAND`) are typed. So this:
```
Usage: pop new pallet [OPTIONS] [NAME] [COMMAND]

Commands:
  advanced  Using the advanced mode will unlock all the POP CLI potential. You'll be able to fully customize your pallet template!. Don't use this mode unless you exactly know what you want for your pallet
  help      Print this message or the help of the given subcommand(s)

Arguments:
  [NAME]  Name of the pallet [default: pallet-template]

Options:
  -a, --authors <AUTHORS>          Name of authors [default: Anonymous]
  -d, --description <DESCRIPTION>  Pallet description [default: "Frame Pallet"]
  -p, --path <PATH>                Path to the pallet, [default: current directory]
  -h, --help                       Print help
  ````
  should be revised to this:
  ```
Usage: pop new pallet [OPTIONS] [NAME] [COMMAND]

Options:
  -a, --authors <AUTHORS>          Name of authors [default: Anonymous]
  -d, --description <DESCRIPTION>  Pallet description [default: "Frame Pallet"]
  -p, --path <PATH>                Path to the pallet, [default: current directory]
  -h, --help                       Print help

Arguments:
  [NAME]  Name of the pallet [default: pallet-template]
  
Commands:
  advanced  Using the advanced mode will unlock all the POP CLI potential. You'll be able to fully customize your pallet template!. Don't use this mode unless you exactly know what you want for your pallet
  help      Print this message or the help of the given subcommand(s)
````
a) From an end-user perspective (one who has no knowledge of clap or CLI's in general), it is confusing the difference between `[OPTIONS]`, `[NAME]`, and `[COMMAND]`. From a UX standpoint, we need to simplify this. The name of the pallet should be an option. The name should be part of the interactive prompt so that I can call advanced mode like so: `pop new pallet advanced` and the first prompt will be the name.

b) When running advanced mode for the first time, it was not clear to me that the syntax should be `pop new pallet advanced` instead I was typing `pop new pallet --advanced`. I understand that `advanced` is a command and not an option but I could see others (who are not familiar with clap or CLI syntax) making the same mistake. I would solve this by adding an  "example" section similar to what polkadot-sdk does for its subcommands.

So the new revised prompt would look:
```
Usage: pop new pallet [OPTIONS]
       pop new pallet [OPTIONS] [COMMAND]

Commands:
  advanced         Advanced mode enables detailed customization of pallet development.
  help                  Print this message or the help of the given subcommand(s)

Options:
  -n, --name <NAME>  Pallet name [default: "pallet-template"]
  -a, --authors <AUTHORS>          Name of authors [default: Anonymous]
  -d, --description <DESCRIPTION>  Pallet description [default: "Frame Pallet"]
  -p, --path <PATH>                Path to the pallet, [default: current directory]
  -h, --help                       Print help
  
Examples:
  pop new pallet
    Interactive mode via prompts.
  pop new pallet -n my-pallet
    Non-interactive mode. Specify options manually.
  pop new pallet advanced 
    Interactive advanced mode via prompts.
  pop new pallet advanced  -c runtime-origin,currency
    Non-interactive advance mode. Specify options manually. `pop new pallet advanced --help` for a list of options available.
```

2) When I went through the `pop new pallet advanced` interactive mode, it did not prompt me for a default config nor a genesis config, was this intentional? Because I solely went through the interactive mode, I did not even know that this was an option that was available in the non-interactive mode. If I had known that then I may have chosen to go through the non-interactive mode.

**Comment by AlexD10S** [comment]
> 1. The `--help` prompt order should be consistent with the order of how the commands (`OPTIONS`, `NAME` `COMMAND`) are typed. So this:
> 
> ```
> Usage: pop new pallet [OPTIONS] [NAME] [COMMAND]
> 
> Commands:
>   advanced  Using the advanced mode will unlock all the POP CLI potential. You'll be able to fully customize your pallet template!. Don't use this mode unless you exactly know what you want for your pallet
>   help      Print this message or the help of the given subcommand(s)
> 
> Arguments:
>   [NAME]  Name of the pallet [default: pallet-template]
> 
> Options:
>   -a, --authors <AUTHORS>          Name of authors [default: Anonymous]
>   -d, --description <DESCRIPTION>  Pallet description [default: "Frame Pallet"]
>   -p, --path <PATH>                Path to the pallet, [default: current directory]
>   -h, --help                       Print help
> ```
> 
> should be revised to this:
> 
> ```
> Usage: pop new pallet [OPTIONS] [NAME] [COMMAND]
> 
> Options:
> -a, --authors <AUTHORS>          Name of authors [default: Anonymous]
> -d, --description <DESCRIPTION>  Pallet description [default: "Frame Pallet"]
> -p, --path <PATH>                Path to the pallet, [default: current directory]
> -h, --help                       Print help
> 
> Arguments:
> [NAME]  Name of the pallet [default: pallet-template]
> 
> Commands:
> advanced  Using the advanced mode will unlock all the POP CLI potential. You'll be able to fully customize your pallet template!. Don't use this mode unless you exactly know what you want for your pallet
> help      Print this message or the help of the given subcommand(s)
> ```
> 
> a) From an end-user perspective (one who has no knowledge of clap or CLI's in general), it is confusing the difference between `[OPTIONS]`, `[NAME]`, and `[COMMAND]`. From a UX standpoint, we need to simplify this. The name of the pallet should be an option. The name should be part of the interactive prompt so that I can call advanced mode like so: `pop new pallet advanced` and the first prompt will be the name.
> 
> b) When running advanced mode for the first time, it was not clear to me that the syntax should be `pop new pallet advanced` instead I was typing `pop new pallet --advanced`. I understand that `advanced` is a command and not an option but I could see others (who are not familiar with clap or CLI syntax) making the same mistake. I would solve this by adding an "example" section similar to what polkadot-sdk does for its subcommands.
> 
> So the new revised prompt would look:
> 
> ```
> Usage: pop new pallet [OPTIONS]
>        pop new pallet [OPTIONS] [COMMAND]
> 
> Commands:
>   advanced         Advanced mode enables detailed customization of pallet development.
>   help                  Print this message or the help of the given subcommand(s)
> 
> Options:
>   -n, --name <NAME>  Pallet name [default: "pallet-template"]
>   -a, --authors <AUTHORS>          Name of authors [default: Anonymous]
>   -d, --description <DESCRIPTION>  Pallet description [default: "Frame Pallet"]
>   -p, --path <PATH>                Path to the pallet, [default: current directory]
>   -h, --help                       Print help
>   
> Examples:
>   pop new pallet
>     Interactive mode via prompts.
>   pop new pallet -n my-pallet
>     Non-interactive mode. Specify options manually.
>   pop new pallet advanced 
>     Interactive advanced mode via prompts.
>   pop new pallet advanced  -c runtime-origin,currency
>     Non-interactive advance mode. Specify options manually. `pop new pallet advanced --help` for a list of options available.
> ```
> 
> 2. When I went through the `pop new pallet advanced` interactive mode, it did not prompt me for a default config nor a genesis config, was this intentional? Because I solely went through the interactive mode, I did not even know that this was an option that was available in the non-interactive mode. If I had known that then I may have chosen to go through the non-interactive mode.

Regarding your first point, I'm not sure if this change is possible or we are depending in `clap`, we'll need to do some research. However, I believe it should be a separate PR. Just keep in mind that we have this pattern in the other commands as well:
```
Usage: pop new contract [OPTIONS] [NAME]

Arguments:
  [NAME]  The name of the contract

Options:
  -c, --contract-type <CONTRACT_TYPE>  The type of contract [default: examples] [possible values: examples, erc, psp]
  -t, --template <TEMPLATE>            The template to use [possible values: standard, erc20, erc721, erc1155, PSP22, PSP34, dns, cross-contract-calls, multisig]
  -h, --help                           Print help
```
a) Why should `name` be an option rather than an argument? From a UX perspective, it seems more intuitive to use `pop new pallet pallet_name` rather than `pop new pallet --name pallet_name`. 
That said, if you’re suggesting changing name to be an argument instead of an option, it should be handled in a separate PR. In this PR, we're maintaining consistency with the other commands.
About prompt the name I agree, I mentioned it here: https://github.com/r0gue-io/pop-cli/pull/261#discussion_r1721680467 but I disagree with about the first prompt will be the name. We should keep it consistent with the rest of the commands where name is the last prompt .
b) I'm not convinced of what you say that `pop new pallet --advanced` is more intuitive than `pop new pallet advanced`.  But adding an example of its usage in the `--help` is a good idea.

**Comment by brunopgalvao** [comment]
> a) Why should `name` be an option rather than an argument? From a UX perspective, it seems more intuitive to use `pop new pallet pallet_name` rather than `pop new pallet --name pallet_name`.

Good point on the usage of name in the wider context of other Pop CLI commands. That being said, let's keep name as it is.

For the advanced mode, it is not clear which options are already included in the interactive prompt and which are not.

That being said, let's include **all** options, `bool` and non-`bool`, conditionally in the prompt.

```
┌   Pop CLI : Generate a pallet
│
⚙  Additional configuration options
│  
◆  Are you interested in adding one of these types and their usual configuration to your pallet? 
Pick an option by pressing the spacebar. Press enter when you're done!
│  ◻ DefaultConfig (Use a default configuration for your config trait.)
│  ◻ GenesisConfig (Add a genesis config to your pallet.) 
│  ◻ Custom Origin (Add a custom origin to your pallet.) 
└  
```
The above prompt would be created conditionally. For example, if the user did not include storage items then the above prompt should not prompt for `DefaultConfig`. If the user is not interested in any of the options they can simply press ENTER.

Of course, those users who would like to specify all the options via the command line manually, that is also available. 

The only alteration to the `pop new pallet my-pallet advanced --help` prompt would be to add an "example" section:
```
Examples:
  pop new pallet my-pallet advanced
    Interactive advanced mode. 
  pop new pallet my-pallet advanced --config-common-types runtime-origin,currency --storage storage-value,storage-map
    Non-interactive advanced mode. Specify options manually.
```

tsenovilla:
Thank you so much for your great feedback @brunopgalvao @AlexD10S . I've pushed a new version taking all this stuff into account :) Let me know if you come up with another needed improvement

crates/pop-parachains/src/new_pallet.rs
kianenigma: While pallet splitting was a cool experiment and it certainly has its niche use-case, I don't see it is particularly useful for beginner devs as IDE experience is worse, and you might encounter weird compiler issues.

kianenigma: Sorry, my comment was a but mislead bc I thought you are using `#[pallet::import_section]` and so on.

Splitting purely into `.rs` files is rather subjective. Similarly, my opinion is that it is better to small with one file rather than few, but I am sure some devs prefer it the other way. 

Contrary, something that worsens the IDE experience (`#[import_section]`) is more objective.

brunopgalvao: I completely understand. If we do move forward with the modular approach, does the current modular setup (the way that it is structured) look good to you?

tsenovilla: As this is a subjective topic, probably users may tell us if they like one approach or another one after merging. I personally like this one due it moves a lot of logic out of the macro expansion, but also understand the other point :)

brunopgalvao (approved):

AlexD10S (approved):
Excellent job.
I left a small comment, but the only thing left to do is merge `main`, there are some conflicts to resolve. This will also fix the test that's failing in the CI.

PR #272, opened 2024-08-01
New pallet template (default)

tsenovilla:
This is a sub-PR of 

- #261 

Following what @brunopgalvao said here: https://github.com/r0gue-io/pop-cli/pull/261#issuecomment-2257954156 I updated the current template to be closer to the one proposed in https://github.com/paritytech/polkadot-sdk/blob/master/templates/parachain/pallets/template but with some small modifications. 

This gives pop cli an updated template for the command `pop new pallet` allowing newcomers to generate a default template full of docs, ideal to start learning about FRAME development. 

The main PR is still open and will be used to propose the advanced, dynamic mode for `pop new pallet`


brunopgalvao (approved):

AlexD10S (approved):
Looks good! Thanks for contributing to pop-cli.

Closes https://github.com/r0gue-io/pop-cli/issues/260

Issue #275, opened 2024-08-04
feat: adding new packages to workspace `Cargo.toml` automatically 


chungquantin:
I try to create a new pallet and a new contract in my parachain but seems like those packages are not added to `workspace.members`. To reproduce: 

```
pop new parachain 

cd ./test-parachain

pop new contract 

cd ./test-contract

pop build

┌   Pop CLI : Building your contract
│
Error: Error invoking `cargo metadata` for Cargo.toml

Caused by:
    `cargo metadata` exited with an error: error: current package believes it's in a workspace when it's not:
    current:   /Users/chungquantin/Developer/pop-network/examples/test-parachain/my_contract/Cargo.toml
    workspace: /Users/chungquantin/Developer/pop-network/examples/test-parachain/Cargo.toml

    this may be fixable by adding `my_contract` to the `workspace.members` array of the manifest located at: /Users/chungquantin/Developer/pop-network/examples/test-parachain/Cargo.toml
    Alternatively, to keep it out of the workspace, add the package to the `workspace.exclude` array, or add an empty `[workspace]` table to the package's manifest.
```

To enhance the experience of the developers, we should add the package to `Cargo.toml` automatically because I believe most of the time, they will forget to add it.

AlexD10S:
Solved in https://github.com/r0gue-io/pop-cli/pull/277

PR #277, opened 2024-08-06
feat: adding new packages to workspace Cargo.toml automatically

tsenovilla:
Related to: 
- #275 

While working in the PR #261 I've been thinking that the feature described in the issue is a must. I took the time to work on a potential solution, hope you find it useful :)


crates/pop-cli/src/commands/new/contract.rs
```diff
@@ -75,6 +78,12 @@ impl NewContractCommand {
 
 		is_template_supported(contract_type, &template)?;
 		generate_contract_from_template(name, &path, &template)?;
+
+		// If the contract is part of a workspace, add it to that workspace
```
AlexD10S, line 82: There's a bug I've encountered. To replicate it: Inside the pop-cli folder, if I run `cargo run new contract test`, it doesn't add the contract to the workspace.

crates/pop-cli/src/common/helpers.rs
```diff
@@ -0,0 +1,302 @@
...
+
+/// This function is used to determine if a Path is contained inside a workspace, and returns a PathBuf to the workspace Cargo.toml if found
+/// # Parameters
+/// - target_dir: A directory that may be contained inside a workspace
+/// # Output
+/// Option<PahBuf>: Some if the input is contained in a workspace, the inner PathBuf points to the workspace Cargo.toml. Otherwise, None
+pub(crate) fn find_workspace_toml(target_dir: &Path) -> Option<PathBuf> {
+	let mut dir = target_dir;
+	while let Some(parent) = dir.parent() {
+		let cargo_toml = dir.join("Cargo.toml");
```
AlexD10S, line 18: The bug mentioned above is fixed if I change this line: 
```
let cargo_toml = parent.join("Cargo.toml");
```
However, I'm not sure why this issue only affects adding contracts, while adding a pallet works as expected.

crates/pop-cli/src/common/helpers.rs
AlexD10S, line 1: The functionality used for the contracts and pallets crates makes sense to be moved to the pop_common crate, which is a collection of essential utilities and shared functionality for Pop. You can find it here: [pop-common](https://github.com/r0gue-io/pop-cli/tree/main/crates/pop-common).

Feel free to add it in any way that makes sense. A suggestion might be to place it in the manifest file: https://github.com/r0gue-io/pop-cli/blob/main/crates/pop-common/src/manifest.rs

crates/pop-cli/src/common/helpers.rs
```diff
@@ -0,0 +1,302 @@
+// SPDX-License-Identifier: GPL-3.0
+
+use anyhow;
+use std::{
+	fs::{read_to_string, write},
+	path::{Path, PathBuf},
+};
+use toml_edit::{value, DocumentMut, Item, Value};
+
+/// This function is used to determine if a Path is contained inside a workspace, and returns a PathBuf to the workspace Cargo.toml if found
```
AlexD10S, line 10: Just for consistency, we use this format for comments, example: https://github.com/r0gue-io/pop-cli/blob/main/crates/pop-common/src/manifest.rs#L7

crates/pop-cli/src/common/helpers.rs
```diff
@@ -0,0 +1,302 @@
...
+					return Some(cargo_toml);
+				}
+			}
+		}
+		dir = parent;
+	}
+	None
+}
+
+/// This function is used to add a crate to a workspace.
```
AlexD10S, line 31: Same as above, just a comment for consistency with the rest of the code. We don't indicate the errors.

AlexD10S (changes requested):
Nice work!!
Left few comments, most of it regarding consistency with the rest of the code.
Only thing to fix is the functionality for contracts.

crates/pop-cli/src/common/helpers.rs
tsenovilla: This is clearly a mistake in my PR 🤪I did wanna use the line you mentioned of course. Fixing it

AlexD10S (approved):
Great job! Thanks again for contributing to `pop-cli`.

PR #279, opened 2024-08-09
feat: clean dmpq state at genesis

al3mart:
If HRMP channels need to be open at genesis the dmq state on the relay needs to match the one in the parachains.
This is not the case in the current state of zombienet, so the changes in this MR take care of cleaning the dmq state in the relay so HRMP channels can be open at genesis.

Follow up work:

A sensible next step would be reading the values from the relevant keys in the parachains and writing those in the relay state. As this change stands right now it would work nicely with parachains that have clean state for the dmqs. Usually that's the state of a parachain that is to be launched, but just in case the mentioned improvements should fit all scenarios.


crates/pop-parachains/src/errors.rs
```diff
@@ -41,4 +41,6 @@ pub enum Error {
 	UnsupportedCommand(String),
 	#[error("Failed to locate the workspace")]
 	WorkspaceLocate,
+	#[error("Can't set environment variable")]
+	EnvVarSetError,
```
evilrobot-01, line 45: We dont need this anymore.

crates/pop-parachains/src/utils/helpers.rs
```diff
@@ -24,6 +25,57 @@ pub(crate) fn sanitize(target: &Path) -> Result<(), Error> {
 	Ok(())
 }
 
+pub async fn clear_dmpq(
```
evilrobot-01, line 28: Please can you add doc comments to function signature.

crates/pop-parachains/src/utils/helpers.rs
```diff
@@ -24,6 +25,57 @@ pub(crate) fn sanitize(target: &Path) -> Result<(), Error> {
...
+	client: OnlineClient<PolkadotConfig>,
+	para_ids: &[u32],
+) -> Result<(), Box<dyn std::error::Error>> {
+	use subxt_signer::sr25519::dev;
+
+	#[subxt::subxt(runtime_metadata_path = "./src/utils/artifacts/paseo-local.scale")]
```
evilrobot-01, line 34: This file is rather large. I believe there is a way to filter the metadata which is included based on only those explicit pallets required. Can we limit it to the system pallet only?

evilrobot-01 (changes requested):
Looks great, just a few minor improvements left please.

crates/pop-parachains/src/utils/helpers.rs
al3mart: Ideally we could generate the interface from the given client on the fly ? I don't know how feasible is that

evilrobot-01 (approved):

crates/pop-parachains/src/utils/helpers.rs
evilrobot-01: Will see how small the resulting metadata is with only the required pallet first, as generating it on every run or caching it brings additional work/complexities.

evilrobot-01: FYI - reduced from 308KB to 48 KB with:

`subxt metadata --url wss://rpc.ibp.network/paseo --pallets System,Sudo > paseo.scale`

evilrobot-01: There is a problem with the filtering sadly: https://github.com/paritytech/subxt/issues/1659

PR #281, opened 2024-08-14
ci: add clippy checks

evilrobot-01:
Discovered that we missed clippy checks in the ci workflow so this PR adds it, along with a commit to address the warnings/errors found.


al3mart (approved):

AlexD10S (approved):
Nice to have this in the CI!

AlexD10S (approved):

PR #284, opened 2024-08-16
feat: integration openzeppelin evm template

AlexD10S:
WIP: issue when pop up the new EVM template

### DIFF
--- a/crates/pop-common/src/templates/mod.rs
+++ b/crates/pop-common/src/templates/mod.rs
@@ -3,6 +3,8 @@
 use strum::{EnumMessage, EnumProperty, VariantArray};
 pub use thiserror::Error;
 
+pub mod extractor;
+
 #[derive(Error, Debug)]
 pub enum Error {
 	#[error("The `Repository` property is missing from the template variant")]
--- a/crates/pop-contracts/src/new.rs
+++ b/crates/pop-contracts/src/new.rs
@@ -1,14 +1,10 @@
 // SPDX-License-Identifier: GPL-3.0
 
-use crate::{
-	errors::Error,
-	utils::helpers::{canonicalized_path, copy_dir_all},
-	Contract,
-};
+use crate::{errors::Error, utils::helpers::canonicalized_path, Contract};
 use anyhow::Result;
 use contract_build::new_contract_project;
 use heck::ToUpperCamelCase;
-use pop_common::{replace_in_file, templates::Template, Git};
+use pop_common::{extract_template_files, replace_in_file, templates::Template, Git};
 use std::collections::HashMap;
 use std::path::{Path, PathBuf};
 use url::Url;
@@ -67,28 +63,23 @@ fn create_template_contract(
 	// Retrieve only the template contract files.
 	if template == &Contract::PSP22 || template == &Contract::PSP34 {
 		// Different template structure requires extracting different path
-		extract_contract_files(String::from(""), temp_dir.path(), canonicalized_path.as_path())?;
+		extract_template_files(
+			String::from(""),
+			temp_dir.path(),
+			canonicalized_path.as_path(),
+			None,
+		)?;
 	} else {
-		extract_contract_files(
+		extract_template_files(
 			template.to_string(),
 			temp_dir.path(),
 			canonicalized_path.as_path(),
+			Some(vec!["frontend".to_string()]),
 		)?;
 	}
 
 	// Replace name of the contract.
 	rename_contract(name, canonicalized_path, template)?;
-	Ok(())
-}
-
-fn extract_contract_files(
-	contract_name: String,
-	repo_folder: &Path,
-	target_folder: &Path,
-) -> Result<()> {
-	let contract_folder = repo_folder.join(contract_name);
-	// Recursively copy all folders and files within. Ignores frontend folders.
-	copy_dir_all(&contract_folder, target_folder)?;
 	Ok(())
 }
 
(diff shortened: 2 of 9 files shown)


ozgunozerk:
Thanks for that! Looking great to me as well

Cargo.toml
```diff
@@ -60,8 +60,8 @@ toml_edit = { version = "0.22", features = ["serde"] }
 symlink = "0.1"
 serde_json = { version = "1.0", features = ["preserve_order"] }
 serde = { version = "1.0", features = ["derive"] }
-zombienet-sdk = "0.2.7"
-zombienet-support = "0.2.7"
+zombienet-sdk = { git = "https://github.com/paritytech/zombienet-sdk", branch = "fix-stash-gen-key"}
+zombienet-support = { git = "https://github.com/paritytech/zombienet-sdk", branch = "fix-stash-gen-key"}
```
pepoviola, line 64:
```suggestion
zombienet-sdk = "0.2.10"
zombienet-support = "0.2.10"
```

Also, which types are you using from `support` (since those should be re-exported by the sdk)?

AlexD10S: Got it, `LocalFileSystem` can be exported from the `zombienet-sdk`.

crates/pop-parachains/src/new_parachain.rs
```diff
@@ -75,6 +78,20 @@ pub fn instantiate_standard_template(
...
 
+pub fn instantiate_openzeppelin_template(
+	template: &Parachain,
+	target: &Path,
+	tag_version: Option<String>,
+) -> Result<Option<String>> {
+	let temp_dir = ::tempfile::TempDir::new_in(std::env::temp_dir())?;
+	let source = temp_dir.path();
+
+	let tag = Git::clone_and_degit(template.repository_url()?, source, tag_version)?;
```
evilrobot-01, line 89: Any reason we are still cloning when we know its a GH repo and downloading and extracting from an archive is faster?

crates/pop-parachains/src/templates.rs
```diff
@@ -136,6 +136,21 @@ pub enum Parachain {
 		)
```
evilrobot-01, line 136: Whilst technically out of scope, should the generic OZ template supporte versions include v2.0.1?

crates/pop-parachains/src/new_parachain.rs
```diff
@@ -120,4 +137,31 @@ mod tests {
...
 		Ok(())
 	}
+
+	#[test]
+	fn test_parachain_instantiate_openzeppelin_template() -> Result<()> {
+		let temp_dir = tempfile::tempdir().expect("Failed to create temp dir");
+		instantiate_openzeppelin_template(&Parachain::OpenZeppelinEVM, temp_dir.path(), None)?;
+
+		// Verify that the generated chain_spec.rs file contains the expected content
+		let generated_file_content =
```
evilrobot-01, line 147: Surely this block doesnt apply to this template? Seems as though it has been copied from another test implementation without further consideration.

evilrobot-01 (commented):
A quick review, a few minor comments.

crates/pop-parachains/src/templates.rs
AlexD10S: You're right, I missed that. We need to improve this so we don't have to manually add it each time. We should implement something like `SupportedVersionsFrom`.

crates/pop-parachains/src/new_parachain.rs
AlexD10S: Yes, it was taken from the above and kept it since I don't see any harm, but I agree is an irrelevant test. I removed it.

crates/pop-parachains/src/new_parachain.rs
AlexD10S: Consistency with the rest of the parachain generations, I felt it was out of scope for this PR. It'll be refactored in a separate PR: https://github.com/r0gue-io/pop-cli/issues/312.

crates/pop-parachains/src/templates.rs
```diff
@@ -123,19 +123,34 @@ pub enum Parachain {
 	EVM,
 	// OpenZeppelin
 	#[strum(
-		serialize = "polkadot-generic-runtime-template",
+		serialize = "generic-template",
```
evilrobot-01, line 126: Is this not a breaking change for command line arguments? Will it not affect the analytics?

crates/pop-parachains/src/templates.rs
```diff
@@ -123,19 +123,34 @@ pub enum Parachain {
...
-			SupportedVersions = "v1.0.0",
+			SupportedVersions = "v1.0.0,v2.0.1",
 			IsAudited = "true",
 			License = "GPL-3.0"
 		)
 	)]
 	OpenZeppelinGeneric,
+	// OpenZeppelin EVM
+	#[strum(
+		serialize = "evm-template",
```
evilrobot-01, line 141: Whilst this is done for ease of use in locating the directory within the repo, in terms of the list of template names there is nothing here to disambiguate it from an evm template from any other provider, potentially resulting in a naming clash. As it is a new template it is an opportunity to specify something more useful.

crates/pop-parachains/src/templates.rs
```diff
@@ -232,7 +247,8 @@ mod tests {
 			("contracts".to_string(), Contracts),
 			("evm".to_string(), EVM),
 			// openzeppelin
-			("polkadot-generic-runtime-template".to_string(), OpenZeppelinGeneric),
+			("generic-template".to_string(), OpenZeppelinGeneric),
```
evilrobot-01, line 250: Implies a breaking change.

crates/pop-parachains/src/templates.rs
AlexD10S: Renamed to `openzeppelin/evm-template` and introduced a `template_name_without_provider` method, which will also be useful for Tanssi integrations too.

crates/pop-parachains/src/templates.rs
AlexD10S: You are right. We'll keep `polkadot-generic-runtime-template` but as deprecated working without showing it to the users `--help`

al3mart (commented):
Looks good, just left a small nit :)

crates/pop-common/src/templates/mod.rs
```diff
@@ -99,3 +109,15 @@ macro_rules! enum_variants {
 		.try_map(|s| <$e>::from_str(&s).map_err(|e| format!("could not convert from {s} to type")))
 	}};
 }
+
+#[macro_export]
+macro_rules! enum_variants_for_help {
```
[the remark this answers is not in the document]
AlexD10S: It makes sense the change. More clear, thanks.

al3mart (approved):

crates/pop-contracts/src/new.rs
```diff
@@ -65,15 +65,10 @@ fn create_template_contract(
...
-			String::from(""),
-			temp_dir.path(),
-			canonicalized_path.as_path(),
-			None,
-		)?;
+		extract_template_files("", temp_dir.path(), canonicalized_path.as_path(), None)?;
 	} else {
 		extract_template_files(
-			template.to_string(),
+			&template.to_string(),
```
evilrobot-01, line 71: Clippy warning needs addressing.

evilrobot-01 (approved):

evilrobot-01 (approved):

PR #285, opened 2024-08-16
feat: instantiate tanssi container chains

al3mart:
This change allow users to instantiate Tanssi compatible container chain templates with `pop new parachain`.


crates/pop-common/src/templates/extractor.rs
```diff
@@ -1,26 +1,33 @@
...
 use std::{fs, io, path::Path};
 
+use anyhow::Result;
+
 /// Extracts the specified template files from the repository folder to the target folder.
 ///
 /// # Arguments
-/// * `template_name` - The name of the template to extract.
+/// * `template_file` - The name of the template to extract.
 /// * `repo_folder` - The path to the repository folder containing the template.
```
evilrobot-01, line 11: Directory and folder are used interchangeably, where perhaps we should stick to one. This obviously isnt the goal of this PR, but it make make sense to move towards a standard term for any code that is changed.

My suggestion would be directory.

crates/pop-parachains/src/new_parachain.rs
```diff
@@ -75,12 +80,72 @@ pub fn instantiate_standard_template(
...
+		template_path,
+		temp_dir.path(),
+		&owned_target.join("runtime").as_path(),
+		None,
+	)?;
+
+	// Add network configuration.
+	use askama::Template;
+	let network = Network { node: format!("container-chain-{}-node", template.to_string()) };
+	write_to_file(&target.join("network.toml"), network.render().expect("infallible").as_ref())?;
```
al3mart, line 136: This might need a zn config template specific for container chains so it includes a tanssi node chain by default. That we can then source and run.

Tanssi release not only includes `tanssi-node` but the template nodes we are adding in this PR.

crates/pop-parachains/src/new_parachain.rs
```diff
@@ -75,12 +80,72 @@ pub fn instantiate_standard_template(
 	Ok(tag)
 }
 
+/// Create a new tanssi compatible container chain.
+///
+/// # Arguments
+///
+/// * `template` - template to generate the parachain from.
```
evilrobot-01, line 87:
```suggestion
/// * `template` - template to generate the container-chain from.
```

Container-chain seems to be the preferred styling, at least from their readme

crates/pop-parachains/src/new_parachain.rs
```diff
@@ -75,12 +80,72 @@ pub fn instantiate_standard_template(
...
+/// * `target` - location where the parachain will be created.
+/// * `tag_version` - version to use (`None` to use latest).
+fn instantiate_tanssi_template(
+	template: &Parachain,
```
evilrobot-01, line 91: Minor, but could alias ContainerChain to Parachain so this reads slightly better.

crates/pop-parachains/src/new_parachain.rs
```diff
@@ -75,12 +80,72 @@ pub fn instantiate_standard_template(
...
+	target: &Path,
+	tag_version: Option<String>,
+) -> Result<Option<String>> {
+	let temp_dir = ::tempfile::TempDir::new_in(std::env::temp_dir())?;
+	let source = temp_dir.path();
+	let tag = Git::clone_and_degit(template.repository_url()?, &source, tag_version)?;
+
+	// Relevant files to extract.
+	let files =
+		[".rustfmt.toml", "Cargo.toml", "Cargo.lock", "LICENSE", "README.md", "rust-toolchain"];
```
evilrobot-01, line 101: Feels like this should be defined on the template via enum properties ultimately, once a stable solution is identified.

crates/pop-parachains/src/new_parachain.rs
```diff
@@ -75,12 +80,72 @@ pub fn instantiate_standard_template(
...
+	// Templates are located in Tanssi's repo as follows:
+	// │
+	// ├┐ container-chains
+	// │├┬ nodes
+	// ││└ <template node directories>
+	// │└┬ runtime-templates
+	// │  └ <tempalte runtime directories>
+
+	// Step 1: extract template node.
+	let template_path = format!("container-chains/nodes/{}", template.to_string());
```
evilrobot-01, line 117: As above, ideally encoded in the template enum once stabilised.

crates/pop-parachains/src/templates.rs
```diff
@@ -154,6 +166,30 @@ pub enum Parachain {
...
+			Repository = "https://github.com/moondance-labs/tanssi",
+			Network = "./network.toml"
+		)
+	)]
+	TanssiSimple,
+	/// Tanssi EVM enabled container chain template.
+	#[strum(
+		serialize = "frontier",
+		message = "EVM",
+		detailed_message = "EVM enabled container chain template.",
```
evilrobot-01, line 185: https://github.com/moondance-labs/tanssi/blob/1fe3bc169163b778e3b4251698de22131b60c985/container-chains/runtime-templates/frontier/Cargo.toml#L4 might be sligthly nicer, only as its something they have defined rather than us.

crates/pop-parachains/src/new_parachain.rs
evilrobot-01: Do they provide cross-platform builds? If not we can possibly provide, but really need to improve the automation of updates there as it's still manual triggers.

crates/pop-common/src/templates/extractor.rs
AlexD10S: @al3mart if you are working in this change, do it directly here: https://github.com/r0gue-io/pop-cli/pull/283 so will be applied for the OZ PR too.

crates/pop-parachains/src/new_parachain.rs
al3mart: We might need to provide at least `apple-darwin`

crates/pop-parachains/src/new_parachain.rs
al3mart: Though, as it stands right now the "only" option would be using a single string that includes all these files under in one concrete property. Scrum docs advise against using a big number of properties, and support for types that are not strings is experimental. :/

crates/pop-common/src/templates/extractor.rs
al3mart: https://github.com/r0gue-io/pop-cli/pull/286

al3mart:
This is now ready for review @AlexD10S

Instantiation should be sorted. Leaving everything related to deployment for a following PR where we can take care of sourcing, preparing the specs and other nits required to fit Tanssi use case.

crates/pop-parachains/src/new_parachain.rs
al3mart: To be addressed in the following PR implementing the deployment logic.

evilrobot-01:
Great stuff, looking forward to trying it out.

Suggest we hold off merging until the deployment PR is ready. Perhaps it could be a PR based on this one so it's easier to test when ready.

crates/pop-parachains/src/templates.rs
```diff
@@ -154,8 +166,38 @@ pub enum Parachain {
...
+		serialize = "simple",
+		message = "Simple",
+		detailed_message = "Generic container chain template.",
+		props(
+			Provider = "Tanssi",
+			Repository = "https://github.com/moondance-labs/tanssi",
+			Network = "./network.toml",
+			NodeDirectory = "container-chains/nodes",
+		    RuntimeDirectory = "container-chains/runtime-templates",
+			ExtractableFiles = ".rustfmt.toml,Cargo.toml,Cargo.lock,LICENSE,README.md,rust-toolchain",
```
AlexD10S, line 180: Can we add the .`gitignore` too?

crates/pop-parachains/templates/container/Cargo.templ
```diff
@@ -0,0 +1,43 @@
+[workspace]
```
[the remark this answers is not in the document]
al3mart: Agreed, we can depend solely on the source manifest to craft these. It will require some tweaks in the logic. But nothing impossible.

al3mart:
[sc-1097]

PR #289, opened 2024-08-22
docs: add community section to README

brunopgalvao:
ditto

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,20 @@
 
 <img src="https://github.com/r0gue-io/pop-cli/blob/main/.icons/logo.jpeg"></img>
 
-An all-in-one tool for Polkadot development.
+<div align="center">
+  
+[![Twitter URL](https://img.shields.io/twitter/follow/Pop?style=social)](https://x.com/onpopio/)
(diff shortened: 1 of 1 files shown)


AlexD10S (commented):
Nice.
But I think is a better idea to add the Telegram and all the links as we have it in the pop network repository: https://github.com/r0gue-io/pop-node

brunopgalvao:
> Nice. But I think is a better idea to add the Telegram and all the links as we have it in the pop network repository: https://github.com/r0gue-io/pop-node

Yes we should add it there as well. Inspiration came from Scio Labs ink!athon repo:
- https://github.com/scio-labs/inkathon?tab=readme-ov-file#inkathon-boilerplate

AlexD10S:
> > Nice. But I think is a better idea to add the Telegram and all the links as we have it in the pop network repository: https://github.com/r0gue-io/pop-node
> 
> Yes we should add it there as well. Inspiration came from Scio Labs ink!athon repo:
> 
> * https://github.com/scio-labs/inkathon?tab=readme-ov-file#inkathon-boilerplate

The Telegram link is in the `pop-node` README. My proposal is to add it here as well, using the same format and including it with the badges there.

AlexD10S (approved):
Thanks! LGTM

PR #296, opened 2024-08-28
feat: deploy tanssi container chains

al3mart:
This PR extends `al3mart/feat-tanssi-container-chains` so that container-chains can be run in a local environment without complexities added.

A Tanssi node will be launched with a chain spec that includes in its genesis the container-chain the user wants to run locally.

As per the comments below, the flow imitates the one provided by Tanssi [here](https://github.com/moondance-labs/tanssi/blob/master/test/scripts/build-spec.sh). And although including a custom commands for container chains might give us more flexibility, this should be good enough to verify user interest.

The limiting factor at the moment is for users to launch their container-chains locally in combination with for instance, system chains. Or other scenarios that have a combination of container-chains and non container-chains.


AlexD10S:
Following the Tanssi guides, before spawning the network, we need to build the specs for the `container chains` and the `tanssi-node`. Instead of running the build-spec script ([See the Tanssi script](https://github.com/moondance-labs/tanssi/blob/master/test/scripts/build-spec.sh)), I've added code to handle this by parsing the `network.toml` file to extract values like `chain_spec_path` and the collator names from it in the `pop up parachain` functionality.

_Considerations for my approach:_

1- About how to triggering Tanssi logic: My current approach is to have a hardcoded value in the `pop-cli` crate (frontend), and trigger the logic automatically if the Zombienet file contains a `tanssi-node`.  See: https://github.com/r0gue-io/pop-cli/pull/296/files#diff-676bc9da9a6a8b135164d15340c548a94dcad29b8cd784ebda6a9e3ef2313622R94

Another approach could be allowing users to explicitly specify it by using a flag, like `pop up parachain --container-chain`, to spawn the container-chain environment. Which approach do you think is better?

2- Handling collator injection: In the[ build-spec script](https://github.com/moondance-labs/tanssi/blob/master/test/scripts/build-spec.sh), Tanssi manually injects collator names into `pallet-invulnerables` and includes a comment in the Zombienet file to update the script if changes are made in the .toml file (See here: https://github.com/moondance-labs/tanssi/blob/master/test/configs/zombieTanssi.json#L51).  This is a bit tricky because if user changes the name of the collators it has to be manually changed.
My current approach parses all collators from the Zombienet config and only injects those whose names start with "Collator". Thoughts on this approach?

al3mart:
[sc-1097]

Issue #293, opened 2024-08-23
bug: extra "0x" in outputted text


brunopgalvao:
Typo when using `pop up contract -u`:

“Contract uploaded: The code hash is "0x0x6180559e28b316b4f426f7fffe473539715c213ed9b3b188f4fa5cdb1056e52e””

There is an extra `0x`

PR #298, opened 2024-08-29
fix: remove extra 0x in outputted text when pop up

AlexD10S:
Closes https://github.com/r0gue-io/pop-cli/issues/293

### DIFF
--- a/crates/pop-contracts/src/up.rs
+++ b/crates/pop-contracts/src/up.rs
@@ -192,7 +192,7 @@ pub async fn upload_smart_contract(
 		let code_hash: String =
 			upload_exec.code().code_hash().iter().map(|b| format!("{:02x}", b)).collect();
 		Err(Error::UploadContractError(format!(
-			"This contract has already been uploaded with code hash: {code_hash}"
+			"This contract has already been uploaded with code hash: 0x{code_hash}"
 		)))
 	}
 }
--- a/crates/pop-contracts/src/up.rs
+++ b/crates/pop-contracts/src/up.rs
@@ -362,7 +362,7 @@ mod tests {
 
 		// Only upload a Smart Contract
 		let upload_result = upload_smart_contract(&upload_exec).await?;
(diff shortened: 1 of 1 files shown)


brunopgalvao (approved):

crates/pop-contracts/src/up.rs
```diff
@@ -187,12 +187,12 @@ pub async fn upload_smart_contract(
...
 		.map_err(|error_variant| Error::UploadContractError(format!("{:?}", error_variant)))?;
 	if let Some(code_stored) = upload_result.code_stored {
-		return Ok(format!("0x{:?}", code_stored.code_hash));
+		return Ok(format!("{:?}", code_stored.code_hash));
 	} else {
 		let code_hash: String =
 			upload_exec.code().code_hash().iter().map(|b| format!("{:02x}", b)).collect();
 		Err(Error::UploadContractError(format!(
-			"This contract has already been uploaded with code hash: 0x{code_hash}"
+			"This contract has already been uploaded with code hash: {code_hash}"
```
brunopgalvao, line 195:
```suggestion
			"This contract has already been uploaded with code hash: 0x{code_hash}"
```



Tested this branch. Here is the output on error:
```
┌   Pop CLI : Deploy a smart contract
│
▲  NOTE: contract has not yet been built.
│  
◇  Your contract artifacts are ready. You can find them in: /Users/bruno/src/pop-cli/target/ink/flipper
│
▲  An error occurred uploading your contract: This contract has already been uploaded with code hash: 326159331e77a975d5de42b8d4d1ada972ccd036f7cd27ec79b42bd21f489f9a
│
└  🚫 Deployment failed.
```
Looks like we need the `0x` here.

brunopgalvao (changes requested):

crates/pop-contracts/src/up.rs
AlexD10S: Good catch! Fixed

Issue #237, opened 2024-07-02
fix: set a License default value per Template


AlexD10S:
In the `pop new parachain` command, we fetch the license of the template the user is going to use.
```
   Pop CLI : Generate a parachain
│
◇  Select a template provider: 
│  OpenZeppelin 
│
◇  Select the type of parachain:
│  Generic Runtime Template 
│
⚙  Template License: GPL-3.0
│  
```
In case of the user surpass the GitHub API limit when generating the parachain it throws an error: 
```
Error: HTTP status client error (403 Forbidden) for url (https://api.github.com/repos/org/project/license)
```
The limit is very high so is unlikely to happen, but it looks bad to throw an error because we can't fetch the license. My suggestion is to have a default value for each template. If the API call fails, we use this default value. 
Alternatively, if we want to be very strict about this and don't want to return a default value, we can return an error string, but it shouldn't stop the parachain generation.

The error is thrown here: https://github.com/r0gue-io/pop-cli/blob/main/crates/pop-common/src/git.rs#L219

aaronbassett:
I just hit this rate limit and wasn't doing anything too excessive. I was attempting to view the differences in some of the templates. It was on the 3rd `pop new parachain` call in around 15 mins that I received the error.

The primary rate limit for unauthenticated requests is quite low, [60 requests per hour](https://docs.github.com/en/rest/using-the-rest-api/rate-limits-for-the-rest-api?apiVersion=2022-11-28). I do not think I exceeded that, but I might have exceeded the secondary rate limit as it appears that each time the `pop new parachain` command is run it can generate multiple requests in a short period of time.

> My suggestion is to have a default value for each template. If the API call fails, we use this default value.

I would suggest not using the GitHub API at all. Clone and cache each of the template repositories. You can query these repositories to get the tags, their Cargo.toml files will include their license, and so on. Stale caches could be updated by pulling changes whenever a user runs `pop new parachain` and by providing a `pop update cache` command alongside the `pop clean cache`

aaronbassett:
As an extra note, the rate-limiting isn't a rolling period or similar. I was limited over an hour ago, and I'm still getting the same error. It could be that the lockout period is greater than an hour, or the lockout timer resets if you attempt another request during the lockout period, or `pop new parachain` can sometimes cause an immediate lockout in a single run.

[...]

PR #299, opened 2024-08-29
fix: limit API calls when generating parachain

AlexD10S:
In the `pop new parachain` command, we perform several sequential operations to gather repository information based on the user's selection:
1. Fetch the template license.
2. Retrieve latest releases to display the user the last 3.
3. Fetch commit hashes: For each of these three releases, we retrieve the commit hash associated with the specified tag in the GitHub repository.
```
◇  Select a template provider: 
│  Pop 
│
◇  Select the type of parachain:
│  Standard 
│
⚙  Template License: Unlicense
│  
◆  Select a specific release:
│  ● Polkadot v1.14 (polkadot-v.1.14.0 / 4a6e8ef)
│  ○ Polkadot v1.13 
│  ○ Polkadot v1.12 
└  
```
This process results in a total of five API calls before we even generate the parachain. If the user performs this operation twice in a short period, it results in 10 API calls, which could quickly hit rate limits on the API:
```
Error: HTTP status client error (403 Forbidden) for url (https://api.github.com/repos/org/project/license)
```
The limit per hour seems high enough [60 requests per hour](https://docs.github.com/en/rest/using-the-rest-api/rate-limits-for-the-rest-api?apiVersion=2022-11-28), but it seems it also restricts the call per minute.
I reached the limit while working on an integration which means I got stuck and can generate more parachains for a certain period of time. Same happened to another user: https://github.com/r0gue-io/pop-cli/issues/237#issuecomment-2291114617

**Solution**
By default, the command won't fetch the `license` (since it's unlikely to change and can be hardcoded) or the `commit hashes`, cutting the API queries down to just one. If users still want to verify this info, they can use the `--verify` flag. This approach keeps the _"Don't Trust, Verify"_ principle intact while reducing API calls for most users, making parachain generation smoother and avoiding rate limit issues.
I couldn’t find a way to fetch all the information in a single call, which would be ideal, so I think this is the best approach for now. However, I’m open to other ideas if you have any suggestions.

Closes: https://github.com/r0gue-io/pop-cli/issues/237

### DIFF
--- a/crates/pop-cli/src/commands/new/parachain.rs
+++ b/crates/pop-cli/src/commands/new/parachain.rs
@@ -197,8 +197,8 @@ async fn generate_parachain_from_template(
 	if verify && tag.is_some() {
 		let url = url::Url::parse(&template.repository_url()?).expect("valid repository url");
 		let repo = GitHub::parse(url.as_str())?;
-		let commit = repo.get_commit_sha_from_release(&tag.clone().unwrap()).await?;
-		verify_note = format!(" ✅ Fetched the latest release of the template along with its license based on the commit SHA for the release ({}).", commit);
+		let commit = repo.get_commit_sha_from_release(&tag.clone().unwrap()).await;
+		verify_note = format!(" ✅ Fetched the latest release of the template along with its license based on the commit SHA for the release ({}).", commit.unwrap_or_default());
 	}
 	success(format!(
 		"Generation complete{}",
--- a/crates/pop-cli/src/commands/new/parachain.rs
+++ b/crates/pop-cli/src/commands/new/parachain.rs
@@ -61,6 +61,12 @@ pub struct NewParachainCommand {
 		default_value = DEFAULT_INITIAL_ENDOWMENT
 	)]
 	pub(crate) initial_endowment: Option<String>,
+	#[arg(
+		short = 'v',
+		long,
+		help = "Fetches the latest license, release, and commit SHA data from GitHub."
+	)]
+	pub(crate) verify: bool,
 }
 
 impl NewParachainCommand {
@@ -68,7 +74,7 @@ impl NewParachainCommand {
 	pub(crate) async fn execute(self) -> Result<Parachain> {
 		// If user doesn't select the name guide them to generate a parachain.
 		let parachain_config = if self.name.is_none() {
-			guide_user_to_generate_parachain().await?
+			guide_user_to_generate_parachain(self.verify).await?
 		} else {
 			self.clone()
 		};
@@ -99,7 +105,7 @@ impl NewParachainCommand {
 }
 
 /// Guide the user to generate a parachain from available templates.
-async fn guide_user_to_generate_parachain() -> Result<NewParachainCommand> {
+async fn guide_user_to_generate_parachain(verify: bool) -> Result<NewParachainCommand> {
 	Cli.intro("Generate a parachain")?;
 
 	// Prompt for template selection.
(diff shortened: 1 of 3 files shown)


brunopgalvao:
LGTM

When running `pop new parachain --verify` or `pop new parachain my-parachain --verify` the output screen should have some indication that the `--verify` option succeeded.

Added to Pop Docs.

AlexD10S:
> LGTM
> 
> When running `pop new parachain --verify` or `pop new parachain my-parachain --verify` the output screen should have some indication that the `--verify` option succeeded.
> 
> Added to Pop Docs.
Good idea. What do you think is the best way to go about it? Maybe a note?

brunopgalvao:
> > LGTM
> > When running `pop new parachain --verify` or `pop new parachain my-parachain --verify` the output screen should have some indication that the `--verify` option succeeded.
> > Added to Pop Docs.
> > Good idea. What do you think is the best way to go about it? Maybe a note?

Yes, a note in the output with what was verified would be nice.

Example:
```
┌   Pop CLI : Generating "my-parachain" using Standard from Pop!
│
◆  Generation complete
│  Version: polkadot-v1.9.0 - ✅ Fetched the latest release of the template along with its license based on the commit SHA for the release (4a6e8ef5cade26e0da1fe74ab8bf3509d7f99d59).
│  
▲  NOTE: the resulting parachain is not guaranteed to be audited or reviewed for security vulnerabilities.
│  Please consult the source repository at https://github.com/r0gue-io/base-parachain to assess production suitability and licensing restrictions.
│  
◆  Next Steps:
│  ● cd into "my-parachain" and enjoy hacking! 🚀
│  ● Use `pop build` to build your parachain.
│  ● Use `pop up parachain -f ./network.toml` to launch your parachain on a local network.
│  
└  Need help? Learn more at https://learn.onpop.io
```

crates/pop-cli/src/commands/new/parachain.rs
```diff
@@ -61,14 +61,20 @@ pub struct NewParachainCommand {
 		default_value = DEFAULT_INITIAL_ENDOWMENT
 	)]
 	pub(crate) initial_endowment: Option<String>,
+	#[arg(
+		short = 'v',
+		long,
+		help = "Fetches the latest license, release, and commit SHA data from GitHub."
```
brunopgalvao, line 67:
```suggestion
		help = "Fetches the latest license and release from GitHub by verifying the GitHub commit SHA."
```

AlexD10S:
> > > LGTM
> > > When running `pop new parachain --verify` or `pop new parachain my-parachain --verify` the output screen should have some indication that the `--verify` option succeeded.
> > > Added to Pop Docs.
> > > Good idea. What do you think is the best way to go about it? Maybe a note?
> 
> Yes, a note in the output with what was verified would be nice.
> 
> Example:
> 
> ```
> ┌   Pop CLI : Generating "my-parachain" using Standard from Pop!
> │
> ◆  Generation complete
> │  Version: polkadot-v1.9.0 - ✅ Fetched the latest release of the template along with its license based on the commit SHA for the release (4a6e8ef5cade26e0da1fe74ab8bf3509d7f99d59).
> │  
> ▲  NOTE: the resulting parachain is not guaranteed to be audited or reviewed for security vulnerabilities.
> │  Please consult the source repository at https://github.com/r0gue-io/base-parachain to assess production suitability and licensing restrictions.
> │  
> ◆  Next Steps:
> │  ● cd into "my-parachain" and enjoy hacking! 🚀
> │  ● Use `pop build` to build your parachain.
> │  ● Use `pop up parachain -f ./network.toml` to launch your parachain on a local network.
> │  
> └  Need help? Learn more at https://learn.onpop.io
> ```

I added the note that way, I like it

brunopgalvao (approved):

PR #306, opened 2024-09-09
feat: guide user to call a contract

AlexD10S:
PR that parses the contract metadata and guides the user in interacting with the smart contract

```
┌   Pop CLI : Call a contract
│
◇  Where is your project located?
│  ./
│
◇  Paste the on-chain contract address:
│  15XausWjFLBBFLDXUSBRfSfZk25warm4wZRV4ZxhZbfvjrJm
│
◇  Select the message to call:
│  get 
│
◇  Where is your contract deployed?
│  wss://rpc1.paseo.popnetwork.xyz
│
◇  Signer calling the contract:
│  //Alice
│
◓  Calling the contract...                                                                                                                                            ⚙  Result: Ok(false)
│  
▲  Your call has not been executed.
│  
◆  Do you want to do another call?
│  ○ Yes  / ● No 
└  
```
Apologies for the large PR; it grew bigger as we needed to test user interactions in the `pop-cli` crate. 
This required adding more code for testing purposes, including changes to the `CLI module` and the introduction of a new `init_tests` file which contains reusable methods for various tests. 

Ideally, these changes should have been introduced in https://github.com/r0gue-io/pop-cli/pull/315, but they were included here due to the necessity of testing the current implementation.

Another note, for testing purposes it has introduced a `testing.json` file. This file is the metadata of a modified `flipper` contract, which includes an extra message, to effectively test argument parsing and payable functions.
```
/// A message for testing, flips the value of the stored `bool` with `new_value`
/// and is payable
#[ink(message)]
#[ink(payable)]
pub fn specific_flip(&mut self, new_value: bool) {
    self.value = new_value;
}
```


brunopgalvao (commented):
Overall it LGTM. Some feedback:
```
┌   Pop CLI : Call a contract
│
◇  Where is your project located?
│  ./
│
◇  Paste the on-chain contract address:
│  5DYs7UGBm2LuX4ryvyqfksozNAW5V47tPbGiVgnjYWCZ29bt
│
└  Unable to fetch contract metadata.
Error: Anyhow error: Cannot infer the root project id
```
It would be nice if “error” was not repeated. Something like this:
```┌   Pop CLI : Call a contract
│
◇  Where is your project located?
│  ./
│
◇  Paste the on-chain contract address:
│  5DYs7UGBm2LuX4ryvyqfksozNAW5V47tPbGiVgnjYWCZ29bt
│
└  Unable to fetch contract metadata.
Error: Cannot infer the root project id
```
Same with this:
```
◇  Do you want to execute the call? (Selecting 'No' will perform a dry run)
│  Yes 
│
Error: Rpc error: RPC error: Error when opening the TCP socket: Connection refused (os error 61)

Caused by:
    RPC error: Error when opening the TCP socket: Connection refused (os error 61)
```
Change to this:
```
◇  Do you want to execute the call? (Selecting 'No' will perform a dry run)
│  Yes 
│
Error when opening the TCP socket: Connection refused (os error 61)

Caused by:
    RPC error: Error when opening the TCP socket: Connection refused (os error 61)
```

##


```
◇  Where is your project located?
│  ./flipper
```
Autocomplete when typing the contract's location would be useful.

##

Looking up RPC URL's can be cumbersome. It would be nice to be able to select commonly used RPC URL's such as Pop Network or localhost.

So this:
```
◇  Where is your contract deployed?
│  wss://rpc1.paseo.popnetwork.xyz
```
Could be something like this:
```
◆  Where is your contract deployed?
│  ● Pop Network TestNet (wss://rpc1.paseo.popnetwork.xyz)
│  ○ Local on Port 9944 (ws://localhost:9944) ○ Custom (Specify)
```

##

Change this:
```
◆  Do you want to do another call?
│  ○ Yes  / ● No 
```

To something like this:
```
◆  Do you want to do another call?
│  ● No
│  ○ Yes, with the same contract
│  ○ Yes, with another contract on the same chain
```
If this change will not be incorporated then at least don’t clear the screen when I say “yes” to call another contract so that I can copy the address and chain URL.
Regardless, the default for `pop call contract` should be to never clear the screen when it is ran because I find it useful to have a “log” of calls that I have done either for copying info to use again or simply to remember the order of calls that I have done and with which parameters.

##

Change this:
```
◆  Select the message to call:
│  ● flip ( A message that can be called on instantiated contracts.  This one flips the value of the stored `bool` from `true`  to `false` and vice versa.)
│  ○ get 
└  
```
I would imagine these Rust Doc comments can easily be 4+ lines for more sophisticated contract.
For example:
```rust
/// Use the new `call_v2` host function via the call builder to forward calls to
/// the other contract, initially calling `flip` and then `get` to return the
/// result.
///
/// This demonstrates how to set the new weight and storage limit parameters via
/// the call builder API.
```
- https://github.com/use-ink/ink-examples/blob/682605fb39f4d9a2d69a48ffc0ce042602386ad9/cross-contract-calls/lib.rs#L75C1-L80C34

Therefore I would have it be displayed underneath the function call making sure to preserve the same formatting that was presented in the Rust Doc comment e.g.:
```
◆  Select the message to call:
│  ● flip_and_get_invoke_v2_with_limits
|  
│   Use the new `call_v2` host function via the call builder to forward calls to
│   the other contract, initially calling `flip` and then `get` to return the
│   result.
│   
│   This demonstrates how to set the new weight and storage limit parameters via
│   the call builder API.
|  
│  ○ get 
└  
```

##

```
◇  Signer calling the contract:
│  //Alice
```
For a custom account, I was not sure what to input so I pasted my account’s address:
```
◇  Signer calling the contract:
│  16hkVKMehgVjTP7AsUvhU8rsYQhK9JDirPWWhpAwHaFTHRfc
```
```
Error: Failed to create keypair from URI: Cannot parse phrase: mnemonic has an invalid word count: 1. Word count must be 12, 15, 18, 21, or 24
```
It was not intuitive to me that it was expecting a mnemonic phrase like so:
```◇  Signer calling the contract:
│  car pocket team home penalty leg forget attend draft width blush lettuce
```

AlexD10S:
Nice catch on the error messages. 
I’ve improved the error message you mentioned, but error handling needs a proper refactor in all the project. Since it could become a larger task for this PR, I’ve opened a new issue to address it separately: https://github.com/r0gue-io/pop-cli/issues/309

Great point about placing the message docs below the message name I’ve fixed that:
```
◆  Select the message to call:
│  ○ flip
 
│  ● get
 ( Simply returns the current value of our `bool`.)
│  ○ specific_flip
 
└ 
```
Same for clearing the screen when prompting for another call. I liked your idea of reusing the contract information, but I reduced it to two options instead of the three you suggested. If want to call another contract, the user should start again:
```
◆  Do you want to do another call using the existing smart contract?
│  ○ Yes  / ● No 
```

Autocomplete for the path and a list of addresses for me are nice-to-have features we can add in the future, feel free to create issues if you think they are needed.

As for the signer, I’m not sure how to improve that right now, but we’ll be working on a better process for signing transactions soon, so hopefully that will make things clearer.

Another piece of feedback from @peterwht that has been implemented is showing the user the call that will be executed after all the values are prompted:
```
⚙  pop call contract --path ./ --contract 5GumB5XDPMT7i1JAFTApxhEneWqRyLaBGdJb8CVXngb4ngTX --message get --url ws://localhost:9944/ --suri //Alice
│  
◐  Calling the contract...                                                                                                                                            ⚙  Result: Ok(false)
```
The option to skip the `dry-run` and submit the call directly is a new feature. I’d prefer to address this in a separate PR. I’ve opened an issue for it: https://github.com/r0gue-io/pop-cli/issues/310.

brunopgalvao (approved):

AlexD10S:
> Some thoughts;
> 
> * Improving the amount of clicks required to call a contract for testing:
> 
> 1. when I was deploying my contract and testing it I don't want to think about the gas limit or proof size limit. So if we can call contracts with a flag that prevents always having to specify that it would be really nice and decreases the amount of clicks.
> 2. The last question; whether or not to call the contract again, is unnecessary maybe, just immediately ask for which message to call again and if the user wants to stop it can simply do ctrl c. Again, the fewer clicks the better the devex tmo.
> 3. The question whether or not to dry-run the call also seems unnecessary for testing. So perhaps a flag `dev` will prevent the weight limit question described in **1.** and also this dry run question
> 
> * Improving the functionality and usefulness of the pop cli:
>   It would be amazing if we could return the value of the message, this is decoding it and printing it out in the console. Then the dev really doesn't have to use the contracts ui anymore. This is the successful return value, error type, even returning the event would be a requirement here I think.
> * Potential bugs:
>   I couldn't deploy a new contract while using `--salt`.

I like your idea to improving DevEx with less clicks. I’ve added the `dev` flag which for now skips the point 1 and 3. This provides a good starting point, and we can continue iterating from here @brunopgalvao  thoughts?.

Regarding the --salt flag it works for me. Not it has to be an hex encoded value. I tried: `--salt 0x6644a2ae6564ea712b23389008cf9e4fe635cd7d4e9b04dd4733eb614c4112ce` (generated in https://ui.use.ink/instantiate) and `--salt $(date +%s)` (example form ink! docs https://use.ink/4.x/basics/cross-contract-calling#basiccontractref-walkthrough)

crates/pop-contracts/src/init_tests.rs
```diff
@@ -0,0 +1,29 @@
+// SPDX-License-Identifier: GPL-3.0
+
+use crate::{create_smart_contract, Contract};
+use anyhow::Result;
+use std::{fs, path::PathBuf};
+
+pub fn generate_smart_contract_test_environment() -> Result<tempfile::TempDir> {
```
evilrobot-01, line 7:
```suggestion
/// Generates a smart contract test environment.
pub fn new_environment() -> Result<tempfile::TempDir> {
```

crates/pop-contracts/src/init_tests.rs
```diff
@@ -0,0 +1,29 @@
...
+	let temp_dir = tempfile::tempdir().expect("Could not create temp dir");
+	let temp_contract_dir = temp_dir.path().join("testing");
+	fs::create_dir(&temp_contract_dir)?;
+	create_smart_contract("testing", temp_contract_dir.as_path(), &Contract::Standard)?;
+	Ok(temp_dir)
+}
+
+// Function that mocks the build process generating the contract artifacts.
```
evilrobot-01, line 15: Doc comment for public function. Please see all clippy warnings in this PR of the same nature.

crates/pop-contracts/src/init_tests.rs
```diff
@@ -0,0 +1,29 @@
+// SPDX-License-Identifier: GPL-3.0
+
+use crate::{create_smart_contract, Contract};
+use anyhow::Result;
+use std::{fs, path::PathBuf};
+
+pub fn generate_smart_contract_test_environment() -> Result<tempfile::TempDir> {
+	let temp_dir = tempfile::tempdir().expect("Could not create temp dir");
+	let temp_contract_dir = temp_dir.path().join("testing");
+	fs::create_dir(&temp_contract_dir)?;
```
evilrobot-01, line 10: `use fs::{create_dir, copy}`

crates/pop-contracts/src/init_tests.rs
```diff
@@ -0,0 +1,29 @@
...
+	create_smart_contract("testing", temp_contract_dir.as_path(), &Contract::Standard)?;
+	Ok(temp_dir)
+}
+
+// Function that mocks the build process generating the contract artifacts.
+pub fn mock_build_process(
+	temp_contract_dir: PathBuf,
```
evilrobot-01, line 17: "In public interfaces, you usually don’t want to use Path or PathBuf directly, but rather a generic P: AsRef<Path> or P: Into<PathBuf>. That way the caller can pass in Path, PathBuf, &str or String."

https://nick.groenen.me/notes/rust-path-vs-pathbuf/

crates/pop-contracts/src/lib.rs
```diff
@@ -4,6 +4,7 @@
 mod build;
 mod call;
 mod errors;
+mod init_tests;
```
evilrobot-01, line 7:
```suggestion
mod testing;
```
Reads better and is more future proof - we can place anything related to testing contracts within this module and present a better public api to the crate. All these exported functions at the root are not sustainable as the crate grows. A consumer can use `pop_contracts::testing::new_environment(..)` and `pop_contracts::testing::mock_build_process(..)`

crates/pop-contracts/src/init_tests.rs
```diff
@@ -0,0 +1,29 @@
...
+
+use crate::{create_smart_contract, Contract};
+use anyhow::Result;
+use std::{fs, path::PathBuf};
+
+pub fn generate_smart_contract_test_environment() -> Result<tempfile::TempDir> {
+	let temp_dir = tempfile::tempdir().expect("Could not create temp dir");
+	let temp_contract_dir = temp_dir.path().join("testing");
+	fs::create_dir(&temp_contract_dir)?;
+	create_smart_contract("testing", temp_contract_dir.as_path(), &Contract::Standard)?;
```
evilrobot-01, line 11: Should "testing" not be a parameter so that this crate api is more usable?

evilrobot-01 (changes requested):
High level review only. I think addressing the suggested changes around the call contract command can greatly improve the UX and the simplify the implementation.

A more flexible experience provides a greater experience. With very little refactoring, I can partially specify args on the command line e.g. 'pop call contract --contract xx' and the call UI can be used if message is ommitted.

That means I can enter stuff into the command which I dont want to repeat over and over, and then allow the terminal history to bring it up and to get me back to the bits I want to specify.

I feel this should be addressed prior to this being merged as it improves the UX significantly.

crates/pop-contracts/src/init_tests.rs
AlexD10S: You are right.

crates/pop-contracts/src/init_tests.rs
AlexD10S: Nice. Is a better approach  to use a single generic parameter
```rs
pub fn mock_build_process<P>(temp_contract_dir: P, contract_file: P, metadata_file: P) -> Result<()>
where
	P: AsRef<Path>,
{
```
or one for each?
```rs
pub fn mock_build_process<P1, P2, P3>(
    temp_contract_dir: P1,
    contract_file: P2,
    metadata_file: P3,
) -> Result<()> 
where
    P1: AsRef<Path>,
    P2: AsRef<Path>,
    P3: AsRef<Path>,
{
```

crates/pop-contracts/tests/files/testing.contract
```diff
@@ -1 +1 @@
...
...ssageResult"]},"docs":[" A message that can be called on instantiated contracts."," This one flips the value of the stored `bool` from `true`"," to `false` and vice versa."],"default":false},{"label":"get","selector":"0x2f865bd9","mutates":false,"payable":false,"args":[],"returnType":{"type":5,"displayName":["ink","MessageResult"]},"docs":[" Simply returns the current value of our `bool`."],"default":false},{"label":"specific_flip","selector":"0x6c0f1df7","mutates":true,"payable":true,"args":[{"label":"new_value","type":{"type":0,"displayName":["bool"]}}],"returnType":{"type":2,"displayName":["ink","MessageResult"]},"docs":[" A message for testing, flips the value of the stored `bool` with `new_value`"," and is payable"],"default":false}],"events":[],"docs":[],"lang_error":{"type":4,"displayName":["ink","LangError"]},"environment":{"accountId":{"type":6,"displayName":["AccountId"]},"balance":{"type":9,"displayName":["Balance"]},"hash":{"type":10,"displayName":["Hash"]},"timestamp":{"type":11,"displayName":["Timestamp"]},"blockNumber":{"type":12,"displayName":["BlockNumber"]},"chainExtension":{"type":13,"displayName":["ChainExtension"]},"maxEventTopics":4,"staticBufferSize":16384}}}
```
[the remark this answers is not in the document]
AlexD10S: Testing. Allows testing `pop call contract` and `pop up contract` without compiling (https://github.com/r0gue-io/pop-cli/blob/feat-call-ui-contracts/crates/pop-contracts/src/init_tests.rs#L16). Uses a flipper contract with custom messages to test features like prompt when contract is payable and argument parsing.

crates/pop-contracts/tests/files/testing.json
```diff
@@ -0,0 +1,453 @@
+{
```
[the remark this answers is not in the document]
AlexD10S: Same as above for testing.  In this case mainly parsing the contract metadata

crates/pop-cli/src/commands/call/contract.rs
```diff
@@ -53,84 +52,752 @@ pub struct CallContractCommand {
...
+		full_message
+	}
+
+	/// Set up the config call.
+	async fn set_up_call_config(
+		&self,
+		cli: &mut impl cli::traits::Cli,
+	) -> anyhow::Result<CallContractCommand> {
+		cli.intro("Call a contract")?;
+		let call_config = if self.contract.is_none() {
```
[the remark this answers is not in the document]
AlexD10S: Updated, now we don’t need to add the address every time.

crates/pop-contracts/src/init_tests.rs
AlexD10S: I'll keep it simple for now with one.

crates/pop-cli/src/commands/call/contract.rs
```diff
@@ -53,84 +52,752 @@ pub struct CallContractCommand {
...
+			}
+		} else {
+			self.clone()
+		};
+		Ok(call_config)
+	}
+}
+
+/// Guide the user to call the contract.
+async fn guide_user_to_call_contract(
```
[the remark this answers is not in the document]
AlexD10S: For the command.field.is_none() in case of `url`, `suri` and `value` I check is not the default value to prompt for it again, because we have a default_value for using the command line

crates/pop-contracts/src/init_tests.rs
evilrobot-01: One, but you can probably also use `impl AsRef<Path>` to avoid the need for an explicit generic.

evilrobot-01:
Running on the command on a fresh project, no build, results in the following:
image

AlexD10S:
> Running on the command on a fresh project, no build, results in the following: image

Good catch! We can apply the same logic as the `pop up contract` building it if it hasn't been built already.

crates/pop-contracts/src/call/metadata.rs
```diff
@@ -0,0 +1,102 @@
...
+	pub docs: String,
+	/// If the message is the default for off-chain consumers (e.g UIs).
+	pub default: bool,
+}
+
+/// Extracts a list of smart contract messages parsing the metadata file.
+///
+/// # Arguments
+/// * `path` -  Location path of the project.
+pub fn get_messages(path: &Path) -> Result<Vec<Message>, Error> {
```
al3mart, line 38: we are expecting path to a manifest, but it would work just fine with a path to the contract bundle. `my_contract.contract`

That way I might not even have to know about the source of the contract, but still interact with it.

al3mart (commented):
Right now we assume that we are receiving a path to a contract project to load the metadata, and if not present build the contract to obtain the metadata.

We only need the metadata in order to call a contract, so it would be handy having both, an option for any user to point to their project and build the contract if necessary, or passing the location of the metadata directly to `pop`. Maybe the user only has access to the metadata.

crates/pop-contracts/src/call/metadata.rs
AlexD10S: Solved in https://github.com/r0gue-io/pop-cli/pull/347

AlexD10S:
> > Good catch! We can apply the same logic as the `pop up contract` building it if it hasn't been built already.
> 
> A note on this one, @AlexD10S . Right now if we execute `pop call contract` on a contract that hasn't being built, and no metadata is provided either, the build starts right away which is great, but right after users are prompted for an address, as if the contract had been deployed already. At that point of the flow there's not much more users can do than killing the process.

Interesting. After some discussions the best approach seems to be prompting the user after build the contract if the contract has already been deployed. There's a valid case where the contract hasn't been built locally, but the user still wants to interact with an already deployed contract, and the local build is only a way to get the artifacts. 
If the user confirms, we continue; if they decline, we display a message prompting them to use pop up contract to deploy the contract. 
Implementation in the last commit https://github.com/r0gue-io/pop-cli/pull/306/commits/d71f35e33d7fe80b6a647e11e26c4e5227984a40

crates/pop-cli/src/commands/call/contract.rs
```diff
@@ -1,71 +1,349 @@
...
+			full_message.push_str(" --dry_run");
+		}
+		full_message
+	}
 
-		let call_exec = set_up_call(CallOpts {
+	/// Ensures the contract is ready by checking if it has been built. If not built, it builds the
+	/// contract. After building, prompts the user to confirm if the contract has already been
+	/// deployed.
+	async fn ensure_contract_ready(&self, cli: &mut impl Cli) -> Result<()> {
```
al3mart, line 133: I believe it would be nice doing a small refactor to separate the logic checking if the contract has been built from the one prompting the user if the contract has been deployed into separate functions.
`ensure_contract_ready` is not that clear to me, while if I see `ensure_contract_built` , I certainly know what to expect when reading through this code.

I appreciate that this was concretely included for this fix.

al3mart (changes requested):
After all the improvements and addressing all the feedback the call contract flow looks pretty nice!

I believe there's one new function that could be divided in two. But aside from that all is looking great. Thanks!

crates/pop-contracts/src/call.rs
```diff
@@ -225,11 +250,9 @@ mod tests {
 			suri: "//Alice".to_string(),
 			execute: false,
 		};
-		let call = set_up_call(call_opts).await;
-		assert!(call.is_err());
-		let error = call.err().unwrap();
-		assert_eq!(error.root_cause().to_string(), "Failed to find any contract artifacts in target directory. \nRun `cargo contract build --release` to generate the artifacts.");
-
+		assert!(
+			matches!(set_up_call(call_opts).await, Err(Error::AnyhowError(message)) if message.root_cause().to_string() == "Failed to find any contract artifacts in target directory. \nRun `cargo contract build --release` to generate the artifacts.")
```
al3mart, line 254: In the future we should ideally prompt to run builds with pop.

AlexD10S: It makes sense. To keep it consistent we should do the same for `pop up contract`, so probably better be in a separate PR: https://github.com/r0gue-io/pop-cli/issues/356

crates/pop-cli/src/commands/call/contract.rs
AlexD10S: Nice discussion around this refactor. Improved it in https://github.com/r0gue-io/pop-cli/pull/306/commits/b0f3e8b28b930a66353faf79166276ba48b9ec88

crates/pop-contracts/src/call.rs
al3mart: :100:

crates/pop-cli/src/commands/call/contract.rs
al3mart: Thanks for the quick action! I'll drop a review shortly :pray:

al3mart (approved):
All is looking great! :heart:

crates/pop-contracts/src/call.rs
```diff
@@ -51,24 +52,43 @@ pub struct CallOpts {
...
+				.url(call_opts.url.clone())
+				.done()
+		},
+	};
 
 	let value: BalanceVariant<<DefaultEnvironment as Environment>::Balance> =
 		parse_balance(&call_opts.value)?;
 
 	let contract: <DefaultConfig as Config>::AccountId = parse_account(&call_opts.contract)?;
+	// Process the argument values input by the user.
```
evilrobot-01, line 81:
```suggestion
	// Process the provided argument values.
```
From the perspective of a library crate, there is no guarantee that it has been input by a user. It could be read from a file as an example.

evilrobot-01 (changes requested):
Whilst there are a lot of comments, most of them are minor. Once addressed I think we should be ready to merge.

crates/pop-cli/src/commands/call/contract.rs
```diff
@@ -1,71 +1,355 @@
...
+		}
+		Ok(())
+	}
+
+	/// Checks whether building the contract is required
+	fn is_contract_build_required(&self) -> bool {
+		self.path
+			.clone()
+			.map(|p| p.is_dir() && !has_contract_been_built(Some(&p)))
+			.unwrap_or_default()
```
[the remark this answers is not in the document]
AlexD10S: At the point is called yes.  I'm not sure if it makes complete sense, as it performs the same function, but for safety, we can pass this value as a parameter: https://github.com/r0gue-io/pop-cli/blob/feat-call-ui-contracts/crates/pop-cli/src/commands/call/contract.rs#L193

crates/pop-contracts/src/utils/metadata.rs
```diff
@@ -0,0 +1,533 @@
...
+		.find(|msg| msg.label == message)
+		.ok_or_else(|| Error::InvalidMessageName(message.to_string()))
+}
+
+/// Extracts a list of smart contract contructors parsing the metadata file.
+///
+/// # Arguments
+/// * `path` -  Location path of the project.
+pub fn get_constructors(path: &Path) -> Result<Vec<ContractFunction>, Error> {
+	let cargo_toml_path = match path.ends_with("Cargo.toml") {
```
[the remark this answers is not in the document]
AlexD10S: Done in this commit https://github.com/r0gue-io/pop-cli/pull/306/commits/5cb68598a76e2f56b17f85e50efed2b5740d010d

crates/pop-cli/src/common/build.rs
```diff
@@ -0,0 +1,49 @@
...
+
+use pop_common::manifest::from_path;
+use std::path::Path;
+/// Checks if a contract has been built by verifying the existence of the build directory and the
+/// <name>.contract file.
+///
+/// # Arguments
+/// * `path` - An optional path to the project directory. If no path is provided, the current
+///   directory is used.
+pub fn has_contract_been_built(path: Option<&Path>) -> bool {
```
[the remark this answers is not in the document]
AlexD10S: Done in this commit https://github.com/r0gue-io/pop-cli/pull/306/commits/410aeea84a2765dbff8b709f111c2a921a969048

evilrobot-01 (approved):
Great stuff! Thank you for your patience on these last few improvements!

Issue #270, opened 2024-07-28
Contract Deployment Failed


Blockcowboy2:
Note: Using contract template.

 🚀 Network launched successfully - ctrl-c to terminate
│  ⛓️ rococo-local
│       alice:
│         portal: https://polkadot.js.org/apps/?rpc=ws://127.0.0.1:**51867**#/explorer
│         logs: tail -f /var/folders/cl/cptzdfls60jf6z6pcrq3dbkm0000gn/T/zombie-db27a062-3320-4b30-9301-aa682a711f9f/alice/alice.log
│       bob:
│         portal: https://polkadot.js.org/apps/?rpc=ws://127.0.0.1:51871#/explorer
│         logs: tail -f /var/folders/cl/cptzdfls60jf6z6pcrq3dbkm0000gn/T/zombie-db27a062-3320-4b30-9301-aa682a711f9f/bob/bob.log
│  ⛓️ local_testnet: 2000
│       collator-01:
│         portal: https://polkadot.js.org/apps/?rpc=ws://127.0.0.1:51875#/explorer
│         logs: tail -f /var/folders/cl/cptzdfls60jf6z6pcrq3dbkm0000gn/T/zombie-db27a062-3320-4b30-9301-aa682a711f9f/collator-01/collator-01.log

flipper % pop up contract --constructor new --args "false" --suri //Alice --url ws://127.0.0.1:**51867**

NOTE: contract has not yet been built.
│  
┌   Pop CLI : Building a contract
│
 [==] Checking clippy linting rules
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.20s
 [==] Building cargo project
    Finished `release` profile [optimized] target(s) in 0.13s
◆  
│  The contract was built in RELEASE mode.
│  
│  Your contract artifacts are ready. You can find them in:
│  /Users/freeman/Polkadot/flipper/target/ink
│  
│    - flipper.contract (code + metadata)
│    - flipper.wasm (the contract's code)
│    - flipper.json (the contract's metadata)
│  
┌   Pop CLI : Deploy a smart contract
│
Warning: This chain does not yet support checking for compatibility of your contract types (https://use.ink/faq#type-comparison).
▲  Anyhow error: Rpc error: RPC error: ErrorObject { code: ServerError(4003), message: "Client error: Execution failed: Other: Exported method ContractsApi_instantiate is not found", data: None }
│
└  **Deployment failed.**

AlexD10S:
[...]

PR #307, opened 2024-09-09
chore: bump cargo-contract and subxt versions

AlexD10S:
This PR upgrades the `subxt` and `subxt-signer` library from version 0.35.3 to 0.37.0 and `contract-build` and `contract-extrinsic` to 5.0.0-alpha where `subxt` has been updated.
This upgrade is necessary because version 0.37 of subxt includes support for the signed extension CheckMetadataHash.

Closes the bug reported https://github.com/r0gue-io/pop-cli/issues/270

To test it you can use contract template in this branch https://github.com/r0gue-io/contracts-parachain/pull/8 or Pop Network testnet.


al3mart (approved):
Awesome!

PR #311, opened 2024-09-13
refactor: improve conatiner workspace instantiation

al3mart:
Refactor and improvements on the logic for instantiate and populate the workspace Cargo of container chains.


crates/pop-common/src/manifest.rs
```diff
@@ -91,6 +92,89 @@ pub fn add_crate_to_workspace(workspace_toml: &Path, crate_path: &Path) -> anyho
...
+
+	let updated_dependencies: Vec<_> = dependencies
+		.into_iter()
+		.filter_map(|k| {
+			if let Some(dependency) = source_manifest_workspace.dependencies.get(&k) {
+				if let Some(d) = dependency.detail() {
+					let mut detail = d.clone();
+					if d.path.is_some() {
+						detail.path = None;
+						detail.git = Some("https://github.com/moondance-labs/tanssi".to_string());
```
AlexD10S, line 144: Since this is a public library on crates.io intended for anyone to integrate, it shouldn't contain hardcoded values like this. Instead, the value should be passed as a parameter, and have any hardcoded values only in the `pop-cli` crate.

crates/pop-common/src/manifest.rs
```diff
@@ -91,6 +92,89 @@ pub fn add_crate_to_workspace(workspace_toml: &Path, crate_path: &Path) -> anyho
...
+			}
+		})
+		.collect();
+
+	target_manifest_workspace.dependencies.extend(updated_dependencies);
+
+	// Update the local runtime dependency.
+	let local_runtime = target_manifest_workspace
+		.dependencies
+		.get_mut(&format!("container-chain-template-{}-runtime", template))
```
AlexD10S, line 167: Same as above

crates/pop-parachains/templates/container/Cargo.templ
```diff
@@ -32,415 +32,8 @@ type_complexity = { level = "allow", priority = 2 }
 unsafe-code = { level = "deny", priority = 1 }
 
 [workspace.dependencies]
-
 container-chain-template-^^template^^-runtime = { path = "runtime", default-features = false }
```
AlexD10S, line 35: Just a question: Does it make sense to keep the `Cargo.templ`, or is simply injecting this line sufficient?

AlexD10S (changes requested):
Looks much cleaner the Cargo now!!
I left a couple of comments to fix

crates/pop-parachains/templates/container/Cargo.templ
al3mart: It provides some lints for clippy and the workspace,  the package author and repo as well as the members of the workspace.
But This is all something that we could source.

crates/pop-common/src/manifest.rs
al3mart: That makes complete sense.

al3mart: This has now been addressed in my last commit.

crates/pop-common/src/manifest.rs
al3mart: Addressed in the last commit.

crates/pop-parachains/templates/container/Cargo.templ
al3mart: I have logged this issue https://github.com/r0gue-io/pop-cli/issues/313 so  we can come back to it and remove any dependency with the askama Cargo template.

AlexD10S (approved):
Cool! Only thing missing is tests that can be tackled in a separate PR

PR #315, opened 2024-09-19
test: use module CLI in all commands and add unit test in pop-cli crate

AlexD10S:
WIP


crates/pop-cli/src/commands/new/parachain.rs
```diff
@@ -247,64 +263,36 @@ fn is_template_supported(provider: &Provider, template: &Parachain) -> Result<()
...
-
 fn get_customization_value(
 	template: &Parachain,
 	symbol: Option<String>,
 	decimals: Option<u8>,
 	initial_endowment: Option<String>,
+	cli: &mut impl cli::traits::Cli,
 ) -> Result<Config> {
-	if !matches!(template, Parachain::Standard) &&
+	if Provider::Pop.provides(template) &&
```
AlexD10S, line 273: Fix needed. Customization values are available for all Pop templates

PR #324, opened 2024-10-19
refactor(contracts): default to riscv release build

evilrobot-01:
Should enable riscv contract building.

Note that I have not been able to test due to struggling to get the `rve-nightly` toolchain installed on my system.

### DIFF
--- a/crates/pop-cli/src/commands/build/mod.rs
+++ b/crates/pop-cli/src/commands/build/mod.rs
@@ -62,8 +62,7 @@ impl Command {
 		#[cfg(feature = "contract")]
 		if pop_contracts::is_supported(args.path.as_deref())? {
 			// All commands originating from root command are valid
-			BuildContractCommand { path: args.path, release: args.release, valid: true }
-				.execute()?;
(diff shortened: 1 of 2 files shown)


AlexD10S (approved):
The build works with this change. 
However in order to the branch to be useful we need to update `pop new contract` to generate the contract with:
```
ink = { git = "https://github.com/r0gue-io/ink ", branch = "sub0", default-features = false }
polkavm-derive = "0.11.0"
```
The e2e test `contract_lifecycle` is failing because of that.

AlexD10S:
> The build works with this change. However to make the branch useful, we need to update the `pop new contract` feature to generate the contract with the following:
> 
> ```
> ink = { git = "https://github.com/r0gue-io/ink ", branch = "sub0", default-features = false }
> polkavm-derive = "0.11.0"
> ```
> 
> The e2e test `contract_lifecycle` is failing because of that.

Resolved the issue with a change in `cargo-contract`. It can be merged now. 
To fix the `contract_lifecycle` test we will need to update `substrate-contracts-node` to use `pallet_revive` (not for now).

Issue #291, opened 2024-08-22
feat: show contract-node's node logs for debugging purposes


brunopgalvao:
It is often the case that when I am debugging ink! smart contracts, I use [ink::env::debug_println!](https://docs.rs/ink_env/4.0.0/ink_env/macro.debug_println.html) which shows up in the contract-node's node logs. Currently Pop CLI does not offer a way to see my node logs which is a deal-breaker for ink! smart contract development.

Ask: Need a way to access to contract-node’s node logs for debugging.

Also, please make sure that when contract node is spun up with the debug flags:
e.g. `substrate-contracts-node -lerror,runtime::contracts=debug`

- https://use.ink/basics/contract-debugging

Looking at node logs is also useful for debugging certain types of errors.

AlexD10S:
Closed in https://github.com/r0gue-io/pop-cli/pull/330

Issue #319, opened 2024-10-10
chore: accept strings as arguments from command line without having to use `Some()`


brunopgalvao:
When instantiating a contract that expects `String`'s as arguments, Pop CLI fails:

For example, instantiating the PSP22 contract:
```rust
#[ink(constructor)]
pub fn new(
    supply: u128,
    name: Option<String>,
    symbol: Option<String>,
    decimals: u8,
) -> Self {
    let (data, events) = PSP22Data::new(supply, Self::env().caller()); // (2)
    let contract = Self {
        data,
        name,
        symbol,
        decimals,
    };
    contract.emit_events(events);
    contract
}
```

```
pop up contract --constructor new --args 10000000, "AWESOME", "AWE", 10
```
The following error is emitted:
```
┌   Pop CLI : Deploy a smart contract
│
■  An error occurred instantiating the contract: No variant 'AWESOME' found
```

For it to work, I need to wrap the strings in `Some()` with `''` around so that `zsh` does not complain:
```
pop up contract --constructor new --args 10000000, 'Some("AWESOME")', 'Some("AWE")', 10
```

brunopgalvao:
So yes, the user should just pass in arguments without Rust syntax, so if the argument is `None` then have it be empty e.g. 

This:
```
pop up contract --constructor new --args 10000000, 'None', 'Some("AWE")', 10
```
Would be:
```
pop up contract --constructor new --args 10000000, , "AWE", 10
```

All the following variations should be okay:
```
pop up contract --constructor new --args 10000000, "", "AWE", 10
```
OR
```
pop up contract --constructor new --args supply=10000000, name=, symbol="AWE", decimals=10
```
OR
```
pop up contract --constructor new --args supply=10000000, name="", symbol=AWE, decimals="10"
```
Pop CLI should (attempt) to parse into the appropriate type.

brunopgalvao:
https://github.com/r0gue-io/pop-cli/pull/330 solves this issue

Issue #290, opened 2024-08-22
feat: show contract's `code_hash` upon deployment


brunopgalvao:
When working with contracts, it is often useful to have the contract's `code_hash`.

Ask: Have `pop up contract` also output the contract's `code_hash`.

Currently, when I deploy the contract using the `pop up contract` command, the contract's account id is outputted but not the `code_hash` and depending on my development flow, I need to go to PolkadotJs Apps and look up the contract's `code_hash` based on the contract's account id.

Here is the output that `cargo-contract` gives (having the events is nice too):
```sh
cargo-contract contract upload --suri //Alice -x
```
```
      Events
       Event Balances ➜ Withdraw
         who: 5GrwvaEF5zXb26Fz9rcQpDWS57CtERHpNehXCPcNoHGKutQY
         amount: 2.376554381mUNIT
       Event Contracts ➜ CodeStored
         code_hash: 0xc5f8f1dd30008a47574c85e4213c2e9fd883fecba5f862858e0bfd68863f6155
         deposit_held: 312.56mUNIT
         uploader: 5GrwvaEF5zXb26Fz9rcQpDWS57CtERHpNehXCPcNoHGKutQY
       Event TransactionPayment ➜ TransactionFeePaid
         who: 5GrwvaEF5zXb26Fz9rcQpDWS57CtERHpNehXCPcNoHGKutQY
         actual_fee: 2.376554381mUNIT
         tip: 0UNIT
       Event System ➜ ExtrinsicSuccess
         dispatch_info: DispatchInfo { weight: Weight { ref_time: 2379752023, proof_size: 6085 }, class: Normal, pays_fee: Yes }

   Code hash 0xc5f8f1dd30008a47574c85e4213c2e9fd883fecba5f862858e0bfd68863f6155
```

```sh
cargo-contract contract instantiate --suri //Alice --args true -x
```
```
 Dry-running new (skip with --skip-dry-run)
    Success! Gas required estimated at Weight(ref_time: 146425616, proof_size: 16689)
Confirm transaction details: (skip with --skip-confirm or -y)
 Constructor new
        Args true
   Gas limit Weight(ref_time: 146425616, proof_size: 16689)
Submit? (Y/n): Y

      Events
       Event Balances ➜ Withdraw
         who: 5GrwvaEF5zXb26Fz9rcQpDWS57CtERHpNehXCPcNoHGKutQY
         amount: 3.381873816mUNIT
       Event Contracts ➜ CodeStored
[...]

PR #330, opened 2024-10-31
fix: improve contract experience

AlexD10S:
- [x] Partially addresses https://github.com/r0gue-io/pop-cli/issues/290. The `code_hash` is now displayed when running `pop up contract` for both upload only (`-u`) and upload + instantiate scenarios.

For instantiating an already uploaded contract, `code_hash` is currently not displayed, as retrieving it would require parsing events or querying the value on-chain.

- [X] Additionally, this PR closes https://github.com/r0gue-io/pop-cli/issues/291

It adds `-lerror,runtime::contracts=debug` when running `substrate-contracts-node` to allow debugging messages (the contract has to be build without --release flag).

### DIFF
--- a/crates/pop-cli/src/commands/up/contract.rs
+++ b/crates/pop-cli/src/commands/up/contract.rs
@@ -224,12 +224,36 @@ impl UpContractCommand {
 		if !self.dry_run {
 			let spinner = spinner();
 			spinner.start("Uploading and instantiating the contract...");
-			let contract_address =
-				instantiate_smart_contract(instantiate_exec, weight_limit).await?;
+			let contract_info = instantiate_smart_contract(instantiate_exec, weight_limit).await?;
 			spinner.stop(format!(
-				"Contract deployed and instantiated: The Contract Address is {:?}",
-				contract_address
+				"Contract deployed and instantiated:\n{}",
+				style(format!(
+					"{}\n{}",
+					style(format!(
+						"{} The Contract Address is {:?}",
+						console::Emoji("●", ">"),
+						contract_info.address
+					))
+					.dim()
+					.to_string(),
+					contract_info
+						.code_hash
+						.map(|hash| style(format!(
+							"{} The Contract Code Hash is {:?}",
+							console::Emoji("●", ">"),
+							hash
+						))
+						.dim()
+						.to_string())
+						.unwrap_or_default(),
+				))
+				.dim()
 			));
+			// spinner.stop(format!(
+			// 	"Contract deployed and instantiated:\nThe Contract Address is {:?}\nThe Contract
+			// Code Hash is {:?}", 	contract_info.address,
+			// 	contract_info.code_hash
+			// ));
 			Self::terminate_node(process)?;
 			Cli.outro(COMPLETE)?;
 		}
--- a/crates/pop-cli/src/commands/up/contract.rs
+++ b/crates/pop-cli/src/commands/up/contract.rs
@@ -230,16 +230,15 @@ impl UpContractCommand {
 				style(format!(
 					"{}\n{}",
 					style(format!(
-						"{} The Contract Address is {:?}",
+						"{} The contract address is {:?}",
 						console::Emoji("●", ">"),
 						contract_info.address
 					))
-					.dim()
-					.to_string(),
+					.dim(),
 					contract_info
 						.code_hash
 						.map(|hash| style(format!(
-							"{} The Contract Code Hash is {:?}",
+							"{} The contract code hash is {:?}",
 							console::Emoji("●", ">"),
 							hash
 						))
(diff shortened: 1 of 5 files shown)


crates/pop-contracts/src/up.rs
```diff
@@ -166,12 +171,18 @@ pub async fn dry_run_upload(
...
 	gas_limit: Weight,
-) -> anyhow::Result<String, Error> {
+) -> anyhow::Result<ContractInfo, Error> {
 	let instantiate_result = instantiate_exec
 		.instantiate(Some(gas_limit))
 		.await
 		.map_err(|error_variant| Error::InstantiateContractError(format!("{:?}", error_variant)))?;
-	Ok(instantiate_result.contract_address.to_string())
+	// If is upload + instantiate, return the code hash.
+	let hash = if let Some(code_hash) = instantiate_result.code_hash {
```
evilrobot-01, line 180: Address clippy warning.

crates/pop-cli/src/commands/up/contract.rs
```diff
@@ -224,11 +224,30 @@ impl UpContractCommand {
...
+				"Contract deployed and instantiated:\n{}",
+				style(format!(
+					"{}\n{}",
+					style(format!(
+						"{} The Contract Address is {:?}",
+						console::Emoji("●", ">"),
+						contract_info.address
+					))
+					.dim()
+					.to_string(),
```
evilrobot-01, line 238: Address clippy warning

crates/pop-cli/src/commands/up/contract.rs
```diff
@@ -224,11 +224,30 @@ impl UpContractCommand {
...
-				instantiate_smart_contract(instantiate_exec, weight_limit).await?;
+			let contract_info = instantiate_smart_contract(instantiate_exec, weight_limit).await?;
 			spinner.stop(format!(
-				"Contract deployed and instantiated: The Contract Address is {:?}",
-				contract_address
+				"Contract deployed and instantiated:\n{}",
+				style(format!(
+					"{}\n{}",
+					style(format!(
+						"{} The Contract Address is {:?}",
```
evilrobot-01, line 233:
```suggestion
						"{} The contract address is {:?}",
```

crates/pop-cli/src/commands/up/contract.rs
```diff
@@ -224,11 +224,30 @@ impl UpContractCommand {
...
+						console::Emoji("●", ">"),
+						contract_info.address
+					))
+					.dim()
+					.to_string(),
+					contract_info
+						.code_hash
+						.map(|hash| style(format!(
+							"{} The Contract Code Hash is {:?}",
```
evilrobot-01, line 242:
```suggestion
							"{} The contract code hash is {:?}",
```

evilrobot-01 (changes requested):
Looks good, various minor improvements and clippy warnings to address and then I'm good to approve.

AlexD10S:
> For some reason the logs are empty. Is this the correct log path?

Exactly. If the contract has already been uploaded, retrieving the code_hash is a bit more complex. I left it out for now, but we can handle it later by parsing events or querying on-chain.

As for the logs, they are working. 
Keep in mind that the contract must be built without the `--release` flag to display debug logs (see [Ink! Debugging Guide](https://use.ink/basics/contract-debugging/#how-do-i-print-to-the-terminal-console-from-ink)). Also, note that the proposed `-lerror,runtime::contracts=debug` flag only shows error and debug output. If we need more verbosity, I can switch it to `-linfo,runtime::contracts=debug` to include all info logs.

al3mart (approved):
I see the feedback has been included.
I don't have much more to point out rather than a small nit. But definitely not blocking the merge.

image

The emojis are not colored like  the rest of the prompt is :P
Aside from that, I would assume that a user that can instantiate some code has access to the code and could obtain the hash easily. Though, I agree that being consistent and showing it every time would be the best experience.

crates/pop-contracts/src/up.rs
```diff
@@ -157,6 +157,14 @@ pub async fn dry_run_upload(
 	}
 }
 
+///Type to represent information about a deployed smart contract.
```
evilrobot-01, line 160:
```suggestion
/// Type to represent information about a deployed smart contract.
```

evilrobot-01 (approved):

brunopgalvao:
> Also, note that the proposed `-lerror,runtime::contracts=debug` flag only shows error and debug output. If we need more verbosity, I can switch it to `-linfo,runtime::contracts=debug` to include all info logs.

I think more verbosity would nice.

brunopgalvao (approved):

AlexD10S:
> > Also, note that the proposed `-lerror,runtime::contracts=debug` flag only shows error and debug output. If we need more verbosity, I can switch it to `-linfo,runtime::contracts=debug` to include all info logs.
> 
> I think more verbosity would nice.

Agree. Makes sense to have `info`

brunopgalvao (approved):

brunopgalvao:
closes https://github.com/r0gue-io/pop-cli/issues/319

Issue #305, opened 2024-09-08
chore: output a live chain spec


brunopgalvao:
```sh
pop build spec --release --id 4024 --type live --relay paseo --protocol-id my_parachain
```
Renders the following json:
```json
{
  "name": "Local Testnet",
  "id": "local_testnet",
  "chainType": "Live",
  "bootNodes": [
    "/ip4/127.0.0.1/tcp/30333/p2p/12D3KooWHArK64vzm3UFQMkjA9rjfnATbkDX61FQjtBAMygYd9RT"
  ],
  "telemetryEndpoints": null,
  "protocolId": "my_parachain",
  "properties": {
    "ss58Format": 42,
    "tokenDecimals": 12,
    "tokenSymbol": "UNIT"
  },
  "relay_chain": "paseo",
  "para_id": 4024,
  "codeSubstitutes": {},
  "genesis": {
    "runtimeGenesis": {
      "code": "....",
      "patch": {
        "balances": {
          "balances": [
            [
              "5GrwvaEF5zXb26Fz9rcQpDWS57CtERHpNehXCPcNoHGKutQY",
              1152921504606846976
            ],
            [
              "5FHneW46xGXgs5mUiveU4sbTyGBzmstUspZC92UhjJM694ty",
              1152921504606846976
            ],
            [
              "5FLSigC9HGRKVhB9FiEo4Y3koPsNmBmLJbpXg2mp1hXcS59Y",
              1152921504606846976
            ],
            [
              "5DAAnrj7VHTznn2AWBemMuyBwZWs6FNFjdyVXUeYum3PTXFy",
              1152921504606846976
            ],
            [
              "5HGjWAeFDfFCWPsjFQdVV2Msvz2XtMktvgocEZcCj68kUMaw",
              1152921504606846976
            ],
            [
              "5CiPPseXPECbkjWCa6MnjNokrgYjMqmKndv2rSnekmSK2DjL",
              1152921504606846976
            ],
            [
              "5GNJqTPyNqANBkUVMN1LPPrxXnFouWXoe2wNSmmEoLctxiZY",
              1152921504606846976
            ],
            [
              "5HpG9w8EBLe5XCrbczpwq5TSXvedjrBGCwqxK1iQ7qUsSWFc",
              1152921504606846976
            ],
            [
              "5Ck5SLSHYac6WFt5UZRSsdJjwmpSZq85fd5TRNAdZQVzEAPT",
              1152921504606846976
            ],
            [
[...]

PR #331, opened 2024-11-01
fix: build spec experience

AlexD10S:
Fixes https://github.com/r0gue-io/pop-cli/issues/305

### Changes

1. Add the flag `--chain` to specify the chain specification. Is to add this flag: https://github.com/paritytech/polkadot-sdk/blob/master/substrate/client/cli/src/params/shared_params.rs#L32 when running the `build-spec` command of the binary.
To generate `pop-node`  chain_spec mainnet for example can be done with:
`pop build spec --release --id 4024 --type live --relay paseo  --chain mainnet`
 (https://github.com/r0gue-io/pop-node/blob/main/node/src/command.rs#L73)

2. As pointed in the issue the `--default-bootnode` was a bit confusing because by default was set to true and to don't  include localhost as the default bootnode has to be specified with  `--default-bootnode false`.
Set by default to false and included if specified `--default-bootnode`.
The other approach is to change the name to `--disable-default-bootnode` and set it by default to false unless the flag is included.

### DIFF
--- a/crates/pop-cli/src/commands/build/spec.rs
+++ b/crates/pop-cli/src/commands/build/spec.rs
@@ -206,8 +206,7 @@ impl BuildSpecCommand {
 			sleep(Duration::from_secs(3))
 		}
 
-		let spinner = cliclack::spinner();
-		spinner.start("Generating chain specification...");
+		cli.info("Generating chain specification...")?;
 
 		// Create output path if needed
 		let mut output_path = self.output_file.unwrap_or_else(|| PathBuf::from("./"));
@@ -241,7 +240,7 @@ impl BuildSpecCommand {
 		};
 
 		// Generate plain spec.
-		spinner.set_message("Generating plain chain specification...");
+		cli.info("Generating plain chain specification...")?;
 		let mut generated_files = vec![];
 		generate_plain_chain_spec(
 			&binary_path,
@@ -268,7 +267,7 @@ impl BuildSpecCommand {
 		chain_spec.to_file(&plain_chain_spec)?;
 
 		// Generate raw spec.
-		spinner.set_message("Generating raw chain specification...");
+		cli.info("Generating raw chain specification...")?;
 		let spec_name = plain_chain_spec
 			.file_name()
 			.and_then(|s| s.to_str())
@@ -284,7 +283,7 @@ impl BuildSpecCommand {
 
 		// Generate genesis artifacts.
 		if self.genesis_code {
-			spinner.set_message("Generating genesis code...");
+			cli.info("Generating genesis code...")?;
 			let wasm_file_name = format!("para-{}.wasm", para_id);
 			let wasm_file = export_wasm_file(&binary_path, &raw_chain_spec, &wasm_file_name)?;
 			generated_files
@@ -292,7 +291,7 @@ impl BuildSpecCommand {
 		}
 
 		if self.genesis_state {
-			spinner.set_message("Generating genesis state...");
+			cli.info("Generating genesis state...")?;
 			let genesis_file_name = format!("para-{}-genesis-state", para_id);
 			let genesis_state_file =
 				generate_genesis_state_file(&binary_path, &raw_chain_spec, &genesis_file_name)?;
--- a/crates/pop-cli/src/commands/build/spec.rs
+++ b/crates/pop-cli/src/commands/build/spec.rs
@@ -9,7 +9,7 @@ use crate::{
 	style::style,
 };
 use clap::{Args, ValueEnum};
-use cliclack::{spinner};
+use cliclack::spinner;
 use pop_common::Profile;
 use pop_parachains::{
 	binary_path, build_parachain, export_wasm_file, generate_genesis_state_file,
@@ -536,7 +536,6 @@ fn ensure_binary_exists(
 }
 
 // Prepare the output path provided.
-
 fn prepare_output_path(output_path: impl AsRef<Path>) -> anyhow::Result<PathBuf> {
 	let mut output_path = output_path.as_ref().to_path_buf();
 	// Convert to string to check for trailing slash.
@@ -573,8 +572,6 @@ fn prepare_output_path(output_path: impl AsRef<Path>) -> anyhow::Result<PathBuf>
 	}
 	Ok(output_path)
 }
-
-
 
 #[cfg(test)]
 mod tests {
@@ -789,8 +786,9 @@ mod tests {
 			("existing_dir/", "existing_dir/chain-spec.json"),
 			("non_existing_dir", "non_existing_dir.json"),
 			("non_existing_dir/", "non_existing_dir/chain-spec.json"),
-			("some_file	", "some_file.json"),
+			("some_file", "some_file.json"),
 			("some_dir/some_file", "some_dir/some_file.json"),
+			("existing_dir/subdir/", "existing_dir/subdir/chain-spec.json"),
 		];
 		for (input_str, expected_str) in &test_cases {
 			let input_path = temp_dir_path.join(input_str);
--- a/crates/pop-cli/src/commands/build/spec.rs
+++ b/crates/pop-cli/src/commands/build/spec.rs
@@ -12,11 +12,7 @@ use pop_parachains::{
 	binary_path, build_parachain, export_wasm_file, generate_genesis_state_file,
 	generate_plain_chain_spec, generate_raw_chain_spec, is_supported, ChainSpec,
 };
-use std::{
-	env::current_dir,
-	fs::create_dir_all,
-	path::{Path, PathBuf},
-};
+use std::{env::current_dir, fs::create_dir_all, path::PathBuf};
 #[cfg(not(test))]
 use std::{thread::sleep, time::Duration};
 use strum::{EnumMessage, VariantArray};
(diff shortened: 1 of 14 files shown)


brunopgalvao:
Looking good. Some notes:

`--release` option is confusing, what does this do? It says: "For production, always build in release mode to exclude debug features", so will the `pop build spec --release` build the parachain in release mode? 

I think `--release` should be removed and the default should be so that when the user runs `pop build spec` it will build in release mode by default.

```
pop build spec --release --id 4024 --type live --relay paseo --chain mainnet 
```
For some reason this does not work, not sure if it is a pop-cli issue or a pop-node issue, also something to keep in mind, `mainnet` on pop-node uses the polkadot relay, so I guess the user would have to be aware to use the appropriate flags to generate the chain spec.

AlexD10S:
> Looking good. Some notes:
> 
> `--release` option is confusing, what does this do? It says: "For production, always build in release mode to exclude debug features", so will the `pop build spec --release` build the parachain in release mode?
> 
> I think `--release` should be removed and the default should be so that when the user runs `pop build spec` it will build in release mode by default.
> 
> ```
> pop build spec --release --id 4024 --type live --relay paseo --chain mainnet 
> ```
> 
> For some reason this does not work, not sure if it is a pop-cli issue or a pop-node issue, also something to keep in mind, `mainnet` on pop-node uses the polkadot relay, so I guess the user would have to be aware to use the appropriate flags to generate the chain spec.

You are right about the `--release` flag been confusing, similar as the problem above with `--default-bootnode` flag. I like your suggestion, I deprecated it  the `release` flag and it builds in RELEASE mode by default.

About this: `pop build spec --release --id 4024 --type live --relay paseo --chain mainnet ` not working, you probably have an old version of pop-node. See https://github.com/r0gue-io/pop-node/blob/main/node/src/command.rs#L73
You can check if your build has the mainnet available running:` ./target/release/pop-node build-spec --chain mainnet `

crates/pop-cli/src/commands/build/spec.rs
```diff
@@ -134,18 +134,22 @@ pub struct BuildSpecCommand {
 	/// [default: ./chain-spec.json].
 	#[arg(short = 'o', long = "output")]
 	pub(crate) output_file: Option<PathBuf>,
-	/// For production, always build in release mode to exclude debug features.
+	/// [DEPRECATED] For production, always build in release mode to exclude debug features.
```
evilrobot-01, line 137: Why is this deprecated? There is nothing in the PR description and no comment here to advise why this is being done.

crates/pop-cli/src/commands/build/spec.rs
```diff
@@ -134,18 +134,22 @@ pub struct BuildSpecCommand {
...
 	#[clap(short = 'r', long, default_value = "true")]
 	pub(crate) release: bool,
 	/// Parachain ID to be used when generating the chain spec files.
 	#[arg(short = 'i', long = "id")]
 	pub(crate) id: Option<u32>,
 	/// Whether to keep localhost as a bootnode.
-	#[clap(long, default_value = "true")]
+	#[clap(long, default_value = "false")]
```
evilrobot-01, line 144:
```suggestion
	#[clap(long)]
```
Surely a default value for a bool is false?

crates/pop-cli/src/commands/build/spec.rs
```diff
@@ -195,7 +199,9 @@ impl BuildSpecCommand {
 		let para_id = self.id.unwrap_or(DEFAULT_PARA_ID);
 		// Notify user in case we need to build the parachain project.
 		if !self.release {
-			cli.warning("NOTE: this command defaults to DEBUG builds for development chain types. Please use `--release` (or simply `-r` for a release build...)")?;
+			cli.warning(
+				"NOTE: this flag is deprecated, this command defaults to a release build builds.",
```
evilrobot-01, line 203:
```suggestion
				"NOTE: this flag is deprecated, this command defaults to a release build.",
```

evilrobot-01: Some context in a comment as to why this is deprecated would be appreciated please.

evilrobot-01 (commented):
Looks good, would just appreciate some context in terms of the deprecation of release option. Why not leave the flexibility?

Would also want @al3mart to look this PR over before being merged.

AlexD10S:
> Looks good, would just appreciate some context in terms of the deprecation of release option. Why not leave the flexibility?
> 
> Would also want @al3mart to look this PR over before being merged.

My apologies for not mentioning it in the PR description. The deprecation of the `--release` flag was implemented in response to a feedback comment https://github.com/r0gue-io/pop-cli/pull/331#issuecomment-2456156539 on the flag being confusing.

The changes were made in a single [commit 37bb25d](https://github.com/r0gue-io/pop-cli/pull/331/commits/37bb25d5ff2aadd774e64614dc90f34b89c6c86a). so will be easy to revert if we decide to keep it. What are your thoughts, @al3mart? A better approach suggested?

crates/pop-cli/src/commands/build/spec.rs
```diff
@@ -134,18 +134,22 @@ pub struct BuildSpecCommand {
...
-	#[clap(long, default_value = "true")]
+	#[clap(long, default_value = "false")]
 	pub(crate) default_bootnode: bool,
 	/// Type of the chain [default: development].
 	#[arg(short = 't', long = "type", value_enum)]
 	pub(crate) chain_type: Option<ChainType>,
+	/// Specify the chain specification. It can be one of the predefined ones (dev, local) or a
+	/// custom one.
+	#[arg(short = 'c', long = "chain", default_value = "dev")]
+	pub(crate) chain: Option<String>,
```
al3mart, line 152: <3

al3mart (commented):
build spec might trigger a compilation. I feel that having the option of declaring if we want the profile to be `release` or `debug` can be convenient .

If we didn't have the flag anywhere else I think removing it might make more sense.
The flag was added mainly so the options for builds were consistent across sub commands
Maybe we can improve the doc as a way to clarify why that flag is present..

AlexD10S:
> build spec might trigger a compilation. I feel that having the option of declaring if we want the profile to be `release` or `debug` can be convenient .
> 
> If we didn't have the flag anywhere else I think removing it might make more sense. The flag was added mainly so the options for builds were consistent across sub commands Maybe we can improve the doc as a way to clarify why that flag is present..

You are right, It makes sense to keep it consistent across the comment. I removed the deprecation message and added more information in the docs to clarify that this command builds the node if it hasn't been built yet.

brunopgalvao:
> build spec might trigger a compilation. I feel that having the option of declaring if we want the profile to be `release` or `debug` can be convenient .
> 
> If we didn't have the flag anywhere else I think removing it might make more sense. The flag was added mainly so the options for builds were consistent across sub commands Maybe we can improve the doc as a way to clarify why that flag is present..

What is the difference in the outputted chain spec artifacts when using debug vs release?

brunopgalvao:
> About this: `pop build spec --release --id 4024 --type live --relay paseo --chain mainnet ` not working, you probably have an old version of pop-node. See https://github.com/r0gue-io/pop-node/blob/main/node/src/command.rs#L73
> You can check if your build has the mainnet available running:`./target/release/pop-node build-spec --chain mainnet`

Yes you are right. Since there are quite a few options for this command that are needed, it would be nice to have an error if for example `--chain mainnet` does not exist instead of the CLI doing nothing which is what it currently does.

AlexD10S:
> Yes you are right. Since there are quite a few options for this command that are needed, it would be nice to have an error if for example `--chain mainnet` does not exist instead of the CLI doing nothing which is what it currently does.

Yes, makes sense. Logs of the command `./target/release/pop-node build-spec` are hidden, maybe we should show them if there is an error: https://github.com/r0gue-io/pop-cli/issues/337

al3mart:
> > build spec might trigger a compilation. I feel that having the option of declaring if we want the profile to be `release` or `debug` can be convenient .
> > If we didn't have the flag anywhere else I think removing it might make more sense. The flag was added mainly so the options for builds were consistent across sub commands Maybe we can improve the doc as a way to clarify why that flag is present..
> 
> What is the difference in the outputted chain spec artifacts when using debug vs release? Can the runtime wasm be different?

Regarding the resulting wasm I'd say that the only possible difference is that if it is not built in release it may contain some extra debug information that can be logged. Release also tries to compile with optimization. This comes to mind - https://paritytech.github.io/polkadot-sdk/master/frame_support/derive.RuntimeDebugNoBound.html

crates/pop-cli/src/commands/build/spec.rs
```diff
@@ -195,13 +200,14 @@ impl BuildSpecCommand {
...
+			cli.warning(
+				"NOTE: this command defaults to DEBUG builds for development chain types. Please use `--release` (or simply `-r` for a release build...)",
+			)?;
 			#[cfg(not(test))]
 			sleep(Duration::from_secs(3))
 		}
 
-		let spinner = cliclack::spinner();
-		spinner.start("Generating chain specification...");
+		cli.info("Generating chain specification...")?;
```
evilrobot-01, line 210: Why all the info outputs instead of the spinner? I'd imagine its just going to flood the terminal with messages which will just be annoying after the first time.

brunopgalvao (approved):

crates/pop-cli/src/commands/build/spec.rs
AlexD10S: It was to fix the previous logs: https://github.com/r0gue-io/pop-cli/pull/334

al3mart (approved):
I think the change on default value for `--default-bootnode` makes sense. I can see how that was confusing.

Love the addition of `--chain`. Was definitely something needed.

crates/pop-cli/src/commands/build/spec.rs
```diff
@@ -206,8 +206,7 @@ impl BuildSpecCommand {
 			#[cfg(not(test))]
 			sleep(Duration::from_secs(3))
 		}
-
-		cli.info("Generating chain specification...")?;
+		let multi = multi_progress("Generating chain specification...");
```
evilrobot-01, line 209: Not sure multi has much value here - there are no actions happening in parallel which need to be shown at the same time and the output in the spinner is replaced in any case. Would simplify the code to just use a single spinner.

crates/pop-cli/src/commands/build/spec.rs
```diff
@@ -284,23 +285,25 @@ impl BuildSpecCommand {
...
+			spinner.set_message("Generating genesis state...");
 			let genesis_file_name = format!("para-{}-genesis-state", para_id);
 			let genesis_state_file =
 				generate_genesis_state_file(&binary_path, &raw_chain_spec, &genesis_file_name)?;
 			generated_files
 				.push(format!("Genesis State file exported at: {}", genesis_state_file.display()));
 		}
 
-		cli.intro("Building your chain spec".to_string())?;
+		spinner.stop("Done.");
```
evilrobot-01, line 304: If using a single spinner, would be better as 'Chain specification built successfully.' or something.

evilrobot-01 (changes requested):
I think the multi/spinner can be improved a little quite easily.

evilrobot-01:
As mentioned separately, there is no consideration for builds using the production profile, which is potentially an important use case. Whilst this may have no effect on the resulting wasm, it did mean a rebuild of the node despite me already having a production build at target/production. This obviously is a waste of time and local storage and also might be frustrating to an end user.

This might be an edge case, but I think the tooling should ideally cater to both novice and professional devs.

crates/pop-cli/src/commands/build/spec.rs
AlexD10S: Solved! It looks nicer now with the `spinner`. I had to include a `spinner.clear();` in case the node has to be built otherwise output in the terminal very messy as Ale indicated here https://github.com/r0gue-io/pop-cli/pull/334 (That was the reason to use `multi_progress`).

al3mart (approved):
LGTM.

Looking forward to merging this one!

crates/pop-cli/src/commands/build/spec.rs
```diff
@@ -127,169 +131,393 @@ pub(crate) enum RelayChain {
...
+#[derive(Args, Default)]
 pub struct BuildSpecCommand {
 	/// File name for the resulting spec. If a path is given,
 	/// the necessary directories will be created
-	/// [default: ./chain-spec.json].
 	#[arg(short = 'o', long = "output")]
 	pub(crate) output_file: Option<PathBuf>,
-	/// For production, always build in release mode to exclude debug features.
-	#[clap(short = 'r', long, default_value = "true")]
+	// TODO: remove at release ...
```
AlexD10S, line 141: 🫡

crates/pop-cli/src/commands/build/spec.rs
```diff
@@ -299,156 +527,370 @@ impl BuildSpecCommand {
 			"Need help? Learn more at {}\n",
 			style("https://learn.onpop.io").magenta().underlined()
 		))?;
-
 		Ok("spec")
 	}
+
+	// Customize a chain specification.
+	fn customize(&self) -> anyhow::Result<()> {
```
AlexD10S, line 534: Just fix the 2 warnings here and we are good to merge

Daanvdplas: Thank you, resolved

crates/pop-cli/src/commands/build/spec.rs
Daanvdplas: Added the upcoming release

Daanvdplas (approved):

al3mart (approved):
The new commits after my last approval seem to make sense. Thanks for addressing those!

PR #332, opened 2024-11-04
fix: parse user inputs for Option arguments

AlexD10S:
This PR improves the DevEx for `pop call contract` when calling contracts with optional parameters (e.g. `Option<String>`). Previously, users were required to wrap optional parameter values in `Some(...)` themselves. Now, users can simply enter the raw value, and `pop-cli` will automatically handle the wrapping, adding `Some(...)` or `None` if empty.

This logic (`generate_message_args`)  has been integrated into `pop_contracts` crate  when preparing the preprocessed data for a contract call (`set_up_call`). With the check there, both CLI users and `pop_contracts` crate consumers no longer need to manually wrap optional values in` Some(...)` or specify `None` for missing values. 

### Update
Update the logic for the missing feature for `pop up contract` when calling contracts with optional parameters.
Fixes: https://github.com/r0gue-io/pop-cli/issues/319

The main change is to move the logic for parsing metadata and processing arguments into a filed used by `call` and `up`. and includes the parsing metadata and processing arguments for constructors. 

### Testing
Most of the code changes has been to adjust testing. To adjust that the contract used for testing has been updated. Is basically the `flipper` contract with a third call to test parsing arguments:
```rs
      /// Constructor that initializes the `bool` value to the given `init_value`.
        #[ink(constructor)]
        pub fn new(init_value: bool, number: Option<u32>) -> Self {
            Self {
                value: init_value,
                number: number.unwrap_or(0),
            }
        }

        /// Constructor that initializes the `bool` value to `false`.
        ///
        /// Constructors can delegate to other constructors.
        #[ink(constructor)]
        pub fn default() -> Self {
            Self::new(Default::default(), None)
        }
        /// A message for testing, flips the value of the stored `bool` with `new_value`
        /// and is payable
        #[ink(message)]
        #[ink(payable)]
        pub fn specific_flip(&mut self, new_value: bool, number: Option<u32>) {
            self.value = new_value;
            if let Some(number) = number {
                self.number = number;
            }
        }
```


evilrobot-01:
The linked issue is about constructor args, but there is nothing specific about that in this PR, only generalised message calling. Has it been tested with constructor args?

crates/pop-contracts/src/call/metadata.rs
```diff
@@ -68,6 +81,39 @@ fn process_args(message_params: &[MessageParamSpec<PortableForm>]) -> Vec<Param>
...
 
+/// Generates a list of processed argument values for a specified contract message,
+/// wrapping each value in `Some(...)` or replacing it with `None` if the argument is optional
+///
+/// # Arguments
+/// * `path` -  Location path of the project.
+/// * `message_label` - Label of the contract message to retrieve.
+/// * `args` - Argument values provided by the user.
+pub fn generate_message_args(
+	path: Option<&Path>,
```
evilrobot-01, line 92: 'In public interfaces, you usually don’t want to use Path or PathBuf directly, but rather a generic P: AsRef<Path> or P: Into<PathBuf>. That way the caller can pass in Path, PathBuf, &str or String.'

crates/pop-contracts/src/errors.rs
```diff
@@ -44,12 +48,14 @@ pub enum Error {
...
+	SourcingError(SourcingError),
 	#[error("Failed to execute test command: {0}")]
 	TestCommand(String),
 	#[error("Unsupported platform: {os}")]
 	UnsupportedPlatform { os: &'static str },
 	#[error("{0}")]
 	UploadContractError(String),
-	#[error("Sourcing error {0}")]
-	SourcingError(SourcingError),
+	#[error("Wrong number of arguments provided. Expecting {expected}, {provided} provided")]
```
evilrobot-01, line 59:
```suggestion
	#[error("Incorrect number of arguments provided. Expecting {expected}, {provided} provided")]
```

crates/pop-contracts/src/call/metadata.rs
```diff
@@ -68,6 +81,39 @@ fn process_args(message_params: &[MessageParamSpec<PortableForm>]) -> Vec<Param>
...
 }
 
+/// Generates a list of processed argument values for a specified contract message,
+/// wrapping each value in `Some(...)` or replacing it with `None` if the argument is optional
+///
+/// # Arguments
+/// * `path` -  Location path of the project.
+/// * `message_label` - Label of the contract message to retrieve.
+/// * `args` - Argument values provided by the user.
+pub fn generate_message_args(
```
evilrobot-01, line 91:
```suggestion
pub fn process_message_args(
```
We are not generating args, we are processing those provided.

crates/pop-contracts/src/call/metadata.rs
```diff
@@ -68,6 +81,39 @@ fn process_args(message_params: &[MessageParamSpec<PortableForm>]) -> Vec<Param>
...
+	path: Option<&Path>,
+	message_label: &str,
```
evilrobot-01, line 93:
```suggestion
	message: &str,
```
Suffix is noise

crates/pop-contracts/src/call/metadata.rs
evilrobot-01, line 92: Why option? Make the caller specify so that this function is less complex.

"In public interfaces, you usually don’t want to use Path or PathBuf directly, but rather a generic P: AsRef<Path> or P: Into<PathBuf>. That way the caller can pass in Path, PathBuf, &str or String."

crates/pop-contracts/src/call/metadata.rs
```diff
@@ -68,6 +81,39 @@ fn process_args(message_params: &[MessageParamSpec<PortableForm>]) -> Vec<Param>
...
+	if args.len() != message.args.len() {
+		return Err(Error::WrongNumberArguments {
+			expected: message.args.len(),
+			provided: args.len(),
+		});
+	}
+	Ok(args
+		.into_iter()
+		.zip(&message.args)
+		.map(|(arg, param)| match (param.type_name.as_str(), arg.is_empty()) {
```
evilrobot-01, line 107: Nicely done!

crates/pop-contracts/src/call/metadata.rs
```diff
@@ -94,9 +140,74 @@ mod tests {
...
+			Err(Error::InvalidMessageName(name)) if name == "wrong_flip".to_string()));
+		let message = get_message(&temp_dir.path().join("testing"), "specific_flip")?;
+		assert_eq!(message.label, "specific_flip");
+		assert_eq!(message.docs, " A message for testing, flips the value of the stored `bool` with `new_value`  and is payable");
+		// assert parsed arguments
+		assert_eq!(message.args.len(), 2);
+		assert_eq!(message.args[0].label, "new_value".to_string());
+		assert_eq!(message.args[0].type_name, "bool".to_string());
+		assert_eq!(message.args[1].label, "number".to_string());
+		assert_eq!(message.args[1].type_name, "Option".to_string());
```
evilrobot-01, line 171: I'd like to see an Option<String> test case/parameter for completeness, only as everything is being treated as a string.

Namely - "" -> None, "pop" => Some("pop")

crates/pop-contracts/src/call/mod.rs
```diff
@@ -67,10 +67,16 @@ pub async fn set_up_call(
 		parse_balance(&call_opts.value)?;
 
 	let contract: <DefaultConfig as Config>::AccountId = parse_account(&call_opts.contract)?;
+	// Parse the argument values input by the user.
```
evilrobot-01, line 70:
```suggestion
	// Process the argument values input by the user.
```

evilrobot-01 (changes requested):
Nicely done! Mostly minor naming and parameter type improvements.

Also - the linked issue is about constructor args, but there is nothing specific about that in this PR, only generalised message calling. Has it been tested with constructor args? If not, then this PR doesnt address that issue and it should be unlinked as appropriate.

crates/pop-contracts/src/call/metadata.rs
AlexD10S: That's true. Bad naming

crates/pop-contracts/src/call/metadata.rs
AlexD10S: This is tested the `process_message_args_work` test: https://github.com/r0gue-io/pop-cli/blob/fix/option-arguments-contracts/crates/pop-contracts/src/call/metadata.rs#L195

crates/pop-contracts/src/call/metadata.rs
AlexD10S: Nice feedback, thanks

AlexD10S: Nice feedback, thanks

AlexD10S:
You're right; the logic for constructors was indeed missing.
Updated the logic to include the logic in `pop up contract` when calling contracts with optional parameters.
Fixes: https://github.com/r0gue-io/pop-cli/issues/319

The main change is to move the logic for parsing metadata and processing arguments into a filed used by `call` and `up`. and includes the parsing metadata and processing arguments for constructors.

crates/pop-contracts/src/utils/metadata.rs
```diff
@@ -0,0 +1,405 @@
...
+}
+
+/// Processes a list of argument values for a specified contract constructor,
+/// wrapping each value in `Some(...)` or replacing it with `None` if the argument is optional
+///
+/// # Arguments
+/// * `path` -  Location path of the project.
+/// * `constructor` - Label of the contract constructor to retrieve.
+/// * `args` - Argument values provided by the user.
+pub fn process_constructor_args<P>(
```
al3mart, line 185: This function does almost the same than `process_message_args`.

al3mart (commented):
`Option` types are covered for constructors now, awesome!

Doing some local testing I detected the following:

The contract constructor I have been using for this test: 
```rust
#[ink(constructor)]
        pub fn new(init_value: Option<bool>, other_value: Option<bool>) -> Self {
            Self { value: init_value.unwrap_or(false) }
        }
```
If I execute:
```bash
pop up contract --args false,false
```
The behavior is the expected, the args are wrapped an it works nicely.

Although, it seems we are counting space chars as arguments.
```bash
pop up contract --constructor new --args false, false 
// outputs
An error occurred instantiating the contract: Incorrect number of arguments provided. Expecting 2, 3 provided
```

~~I have created an issue for this. https://github.com/r0gue-io/pop-cli/issues/336~~
Removed. This is better addressed as part of this PR as pointed out.

AlexD10S:
> `Option` types are covered for constructors now, awesome!
> 
> Doing some local testing I detected the following:
> 
> The contract constructor I have been using for this test:
> 
> ```rust
> #[ink(constructor)]
>         pub fn new(init_value: Option<bool>, other_value: Option<bool>) -> Self {
>             Self { value: init_value.unwrap_or(false) }
>         }
> ```
> 
> If I execute:
> 
> ```shell
> pop up contract --args false,false
> ```
> 
> The behavior is the expected, the args are wrapped an it works nicely.
> 
> Although, it seems we are counting space chars as arguments.
> 
> ```shell
> pop up contract --constructor new --args false, false 
> // outputs
> An error occurred instantiating the contract: Incorrect number of arguments provided. Expecting 2, 3 provided
> ```
> 
> I have created an issue for this. #336

Good catch!  Seems `clap` were discussing this: https://github.com/r0gue-io/pop-cli/pull/332#pullrequestreview-2417931657
Let me look into the best approach here, it's easy to overlook that single space.

crates/pop-contracts/src/utils/metadata.rs
AlexD10S: I initially planned to leave it as-is, but I agree it's a good idea to improve it now. On it.

evilrobot-01:
> I have created an issue for this. #336

Why create an issue, this should be addressed in this PR.

AlexD10S:
The best solution I have found for this issue is to remove the "," as delimiter `value_delimiter = ','`  (as it was before), which makes space the delimiter. Commas in arguments are removed when processing arguments, allowing users to input using` , `.
_Example correct args:_
`--args false, false`
`--args false, ""`
`--args false false`
`--args false ""`
_Wrong args (without spaces):_
`--args false,false`
`--args false""`

Open to better suggestions!

evilrobot-01:
As a (lazy) user, I would want to write as little as possible, so would opt for `--args false false`. What would happen when there are two default args I want to omit: `a: u8, b: Option<u8>, c: Option<u8>, d: u8`. I dont expect `--args 0 0` would work? Fine if not, provided `--args 0, , , 0` does work.

AlexD10S:
> As a (lazy) user, I would want to write as little as possible, so would opt for `--args false false`. What would happen when there are two default args I want to omit: `a: u8, b: Option<u8>, c: Option<u8>, d: u8`. I dont expect `--args 0 0` would work? Fine if not, provided `--args 0, , , 0` does work.

In your example with parameters `a: u8, b: Option<u8>, c: Option<u8>, d: u8`, I liked the idea of allowing `--args 0 0 `to work. However, this approach could lead to confusion. For instance, with `--args 0 1 0`, it’s unclear which argument will be 1.

The format `--args 0, , , 0` works well with the current implementation and avoids ambiguity. Similarly, `--args 0, "", "", 0` is also effective. The key here is maintaining the whitespace between args, otherwise will fail.

al3mart:
I agree that separating the different args with a space is nicer.

Check this out - https://docs.rs/clap/latest/clap/struct.Arg.html#method.num_args
It gives some hints when a delimiter is not specified, and space separated values are just fine.

al3mart (approved):

crates/pop-contracts/src/utils/metadata.rs
```diff
@@ -0,0 +1,367 @@
...
+	/// The label of the parameter.
+	pub label: String,
+	/// The type name of the parameter.
+	pub type_name: String,
+}
+
+#[derive(Clone, PartialEq, Eq)]
+/// Describes a contract function.
+pub struct ContractFunction {
+	/// The label of the constructor.
```
evilrobot-01, line 21:
```suggestion
	/// The label of the function.
```

crates/pop-contracts/src/utils/metadata.rs
```diff
@@ -0,0 +1,367 @@
+// SPDX-License-Identifier: GPL-3.0
+
+use crate::errors::Error;
+use contract_extrinsics::ContractArtifacts;
+use contract_transcode::ink_metadata::MessageParamSpec;
+use scale_info::form::PortableForm;
+use std::path::Path;
+
+#[derive(Clone, PartialEq, Eq)]
```
evilrobot-01, line 9: Personal preference, but I find having the doc comments before the attributes is nicer, as it keeps the code grouped together with type declaration and it makes the doc comments the first thing you read.

crates/pop-contracts/src/utils/metadata.rs
```diff
@@ -0,0 +1,367 @@
...
+/// Describes a contract message.
```
evilrobot-01, line 10:
```suggestion
/// Describes a parameter.
```

crates/pop-contracts/src/utils/metadata.rs
```diff
@@ -0,0 +1,367 @@
...
+	/// The type name of the parameter.
+	pub type_name: String,
+}
+
+#[derive(Clone, PartialEq, Eq)]
+/// Describes a contract function.
+pub struct ContractFunction {
+	/// The label of the constructor.
+	pub label: String,
+	/// If the message accepts any `value` from the caller.
```
evilrobot-01, line 23:
```suggestion
	/// If the function accepts any `value` from the caller.
```

crates/pop-contracts/src/utils/metadata.rs
```diff
@@ -0,0 +1,367 @@
...
+	pub payable: bool,
+	/// The parameters of the deployment handler.
+	pub args: Vec<Param>,
+	/// The constructor documentation.
```
evilrobot-01, line 27:
```suggestion
	/// The function documentation.
```

crates/pop-contracts/src/utils/metadata.rs
```diff
@@ -0,0 +1,367 @@
...
+	pub docs: String,
+	/// If the constructor is the default for off-chain consumers (e.g UIs).
+	pub default: bool,
```
evilrobot-01, line 30: As this only applies to constructors, might it be better to move these field definitions into the enum variants instead?

crates/pop-contracts/src/utils/metadata.rs
```diff
@@ -0,0 +1,367 @@
...
+/// * `path` -  Location path of the project.
+pub fn get_messages(path: &Path) -> Result<Vec<ContractFunction>, Error> {
+	let cargo_toml_path = match path.ends_with("Cargo.toml") {
+		true => path.to_path_buf(),
+		false => path.join("Cargo.toml"),
+	};
+	let contract_artifacts =
+		ContractArtifacts::from_manifest_or_file(Some(&cargo_toml_path), None)?;
+	let transcoder = contract_artifacts.contract_transcoder()?;
+	let mut messages: Vec<ContractFunction> = Vec::new();
```
evilrobot-01, line 54: Can simplify to `Ok(transcoder.metadata().spec().messages().iter().map(|| ContractFunction{}).collect())`

crates/pop-contracts/src/utils/metadata.rs
```diff
@@ -0,0 +1,367 @@
...
+pub fn process_function_args<P>(
+	path: P,
+	label: &str,
+	args: Vec<String>,
+	function_type: FunctionType,
+) -> Result<Vec<String>, Error>
+where
+	P: AsRef<Path>,
+{
+	let parsed_args = args.into_iter().map(|arg| arg.replace(",", "")).collect::<Vec<String>>();
```
evilrobot-01, line 152:
```suggestion
	let parsed_args = args.into_iter().map(|arg| arg.replace(",", "")).collect::<Vec<_>>();
```
FYI - can use _ to avoid explicitly redefining type.

crates/pop-contracts/src/utils/metadata.rs
AlexD10S: Nicer! Changed in both message and contructors

crates/pop-contracts/src/utils/metadata.rs
AlexD10S: `default` exist in messages, let me change the docs. The only only applies to messages (and not `contructor`) is mutates.

evilrobot-01:
Whilst manually testing with Options I found that the UI would show the type as Option only, without the underlying type which makes valid entry impossible. https://github.com/r0gue-io/pop-cli/pull/339 uses the type registry to format the type with its full representation, which should ease entry.

An example is a complex type like Weight, where previously I had to try various combinations until I discovered the correct syntax. The linked PR should show the expected type/value/shape for data entered.

al3mart (approved):

Issue #322, opened 2024-10-16
chore: update Pop CLI to default to Paseo (not Rococo)


brunopgalvao:
Since Rococo has been officially turned off. This issue is a placeholder for replacing instances of Rococo within Pop CLI with Paseo 

For example the default generated network.toml file: 
https://github.com/r0gue-io/pop-cli/blob/2af82eb44dc4385211ee4cb3986ac149fb6dc867/crates/pop-parachains/templates/base/network.templ#L1-L2

https://github.com/search?q=repo%3Ar0gue-io%2Fpop-cli%20rococo-local&type=code

PR #333, opened 2024-11-05
chore: replace rococo to paseo name

CocDap:
Relevant issue: https://github.com/r0gue-io/pop-cli/issues/322

Two manually check on local: 

1. Run `pop new parachain my-para` . Check relay chain is paseo runtime on the polkadot js explorer  ✅ 
image

2. Run `cargo test` ❌ 
image

P/S: I am very happy to contribute for the first time to this project


AlexD10S:
Thank you for your contribution! Really cool.
The test failing in the image is because you probably have a parachain running in the port `9944` and is the one the test is using for testing.

I run the CI to run all the unit and integration tests and are failing.
It doesn't seem is something to change in your code,  the integration tests https://github.com/r0gue-io/pop-cli/actions/runs/11682223509/job/32588340601?pr=333 seems something is missing in the Paseo chain spec generator:   _"Unknown chain, only supported: paseo-dev, asset-hub-paseo-local, paseo-local, bridge-hub-paseo, people-paseo, bridge-hub-paseo-local, people-paseo-local or a json file"_

Let me investigate it.

CocDap:
Thank you for reviewing this PR. I believe I've identified the issue, which originates from the `launch_paseo_and_system_parachain` test case.

It seems that Zombienet only supports `coretime-rococo-local` and does not recognize `coretime-paseo-local`
Code: https://github.com/CocDap/pop-cli/blob/chore/replace-rococo-to-paseo/tests/networks/paseo%2Bcoretime.toml#L16
CC: @AlexD10S

crates/pop-parachains/src/up/chain_specs.rs
```diff
@@ -183,7 +183,7 @@ mod tests {
 	#[tokio::test]
 	async fn chain_spec_generator_returns_none_when_no_match() -> anyhow::Result<()> {
 		let temp_dir = tempdir()?;
-		assert_eq!(chain_spec_generator("rococo-local", None, temp_dir.path()).await?, None);
```
AlexD10S, line 186: In this case leave the `rococo-local`, is a test that specifically tests that `chain_spec_generator` returns `None`  if the value is not one of the Runtimes we support

AlexD10S (changes requested):
@CocDap I’ve rerun the integration tests after the Paseo v1.3.3 release, and it looks like the issue is resolved ,thanks for your patience!

However, the unit tests still need some adjustments. I left a comment about the `chain_spec_generator_returns_none_when_no_match` test you changed.

And the other one it will need adjustment is `binaries_works` (https://github.com/r0gue-io/pop-cli/blob/main/crates/pop-parachains/src/up/mod.rs#L1238), note that the Paseo chain spec generators are now binaries we are fetching increasing the binaries count to 6 instead of 4. Just change the following line(https://github.com/r0gue-io/pop-cli/blob/main/crates/pop-parachains/src/up/mod.rs#L1272), to:
```rs
assert_eq!(zombienet.binaries().count(), 6);
```

CocDap:
It seems that the `launch_paseo_and_system_parachain` test case continues to fail on MacOS.
CC: @AlexD10S

AlexD10S:
> It seems that the `launch_paseo_and_system_parachain` test case continues to fail on MacOS. CC: @AlexD10S

Thanks for the changes. I don't think is anything you have to change in the code, and more the chain generator of Paseo. I am going to investigate why only fails in `MacOS`.

AlexD10S (approved):

AlexD10S:
Fixed! Thanks again fro contributing @CocDap

PR #334, opened 2024-11-05
fix(spec): clean terminal output

al3mart:
The spinner was not properly handled and made the output in the terminal very messy while the build was running.
Just prompting info messages should be OK for now.

Before this change whenever build spec triggers a build we get
image

After the changes the output doesn't mix the build prompts with the build spec command prompts


AlexD10S (approved):
Nice fix!

PR #338, opened 2024-11-06
refactor: use command state

evilrobot-01:
Refactors to use the existing command state. 

- Merged `set_up_call_config` and `guide_user_to_call_contract` into a single function. 
- Also added short symbols for arguments.


AlexD10S (approved):
Great work! This has significantly improved the code's readability, and DevEx feels smoother now.

PR #339, opened 2024-11-07
refactor: format types

evilrobot-01:
Shows the full type representation, making it easier to see the entry format of parameter values.

### DIFF
--- a/crates/pop-cli/src/commands/call/contract.rs
+++ b/crates/pop-cli/src/commands/call/contract.rs
@@ -96,7 +96,8 @@ impl CallContractCommand {
 			full_message.push_str(&format!(" --message {}", message));
 		}
 		if !self.args.is_empty() {
-			full_message.push_str(&format!(" --args {}", self.args.join(",")));
+			let args: Vec<_> = self.args.iter().map(|a| format!("\"{a}\"")).collect();
+			full_message.push_str(&format!(" --args {}", args.join(", ")));
 		}
 		if self.value != DEFAULT_PAYABLE_VALUE {
 			full_message.push_str(&format!(" --value {}", self.value));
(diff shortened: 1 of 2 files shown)


crates/pop-contracts/src/utils/metadata.rs
```diff
@@ -153,26 +300,26 @@ pub fn process_function_args<P>(
 where
 	P: AsRef<Path>,
 {
-	let parsed_args = args.into_iter().map(|arg| arg.replace(",", "")).collect::<Vec<_>>();
```
evilrobot-01, line 156: Note that this breaks entering complex types, where fields are separated by commas.

crates/pop-contracts/src/utils/metadata.rs
```diff
@@ -207,7 +354,7 @@ mod tests {
 		assert_eq!(message[2].args[0].label, "new_value".to_string());
 		assert_eq!(message[2].args[0].type_name, "bool".to_string());
 		assert_eq!(message[2].args[1].label, "number".to_string());
-		assert_eq!(message[2].args[1].type_name, "Option".to_string());
+		assert_eq!(message[2].args[1].type_name, "Option<u32>: None, Some(u32)".to_string());
```
evilrobot-01, line 357: Whilst this may be unnecessary for Options, showing a list of potential variant values for enums could be very useful imo. This could potentially be refactored into a separate field.

crates/pop-cli/src/commands/call/contract.rs
```diff
@@ -96,7 +96,8 @@ impl CallContractCommand {
 			full_message.push_str(&format!(" --message {}", message));
 		}
 		if !self.args.is_empty() {
-			full_message.push_str(&format!(" --args {}", self.args.join(",")));
+			let args: Vec<_> = self.args.iter().map(|a| format!("\"{a}\"")).collect();
```
evilrobot-01, line 99: Args are always quoted for safety - complex types might have spaces or commas in their value entry and its best to wrap with "".

AlexD10S: Nice

PR #347, opened 2024-11-22
feat: allow users to specify custom contract metadata files

AlexD10S:
Allow the user to pass the location of the metadata instead of the path of the contract project. There might be cases where the user only has access to the metadata.

Mentioned in https://github.com/r0gue-io/pop-cli/pull/306#pullrequestreview-2419627832


crates/pop-cli/src/commands/call/contract.rs
```diff
@@ -156,25 +166,15 @@ impl CallContractCommand {
 		}
 
 		// Resolve path.
-		let contract_path = match self.path.as_ref() {
-			None => {
-				let path = Some(PathBuf::from("./"));
-				if has_contract_been_built(path.as_deref()) {
```
AlexD10S, line 162: This was causing a panic if you run `pop call contract` outside a contract folder.

crates/pop-cli/src/commands/call/contract.rs
```diff
@@ -119,26 +119,36 @@ impl CallContractCommand {
...
-		if !has_contract_been_built(self.path.as_deref()) {
-			// Build the contract in release mode
-			cli.warning("NOTE: contract has not yet been built.")?;
-			let spinner = spinner();
-			spinner.start("Building contract in RELEASE mode...");
-			let result = match build_smart_contract(self.path.as_deref(), true, Verbosity::Quiet) {
-				Ok(result) => result,
-				Err(e) => {
-					return Err(anyhow!(format!(
+		if let Some(path) = self.path.as_deref() {
```
evilrobot-01, line 124:
```suggestion
		let Some(path) = self.path.as_deref() else { return Ok(()) };
```
You can reduce nested ifs/complexity of the function by returning early.

crates/pop-cli/src/commands/call/contract.rs
```diff
@@ -119,26 +119,36 @@ impl CallContractCommand {
...
+			// Check if is a directory, if is a file, skip the build process
+			if path.is_dir() {
```
evilrobot-01, line 126:
```suggestion
			if !path.is_dir() || has_contract_been_built(self.path.as_deref()) { return Ok(()) }
```
As above.

crates/pop-cli/src/commands/call/contract.rs
evilrobot-01, line 124: What happens if the path is not specified - surely we should default to the current path and check as appropriate?

Something like `let path = self.path.unwrap_or(Path::new("./"))`

evilrobot-01: I see that configure sets this - perhaps we should use an `.expect()` instead then?

crates/pop-cli/src/commands/call/contract.rs
```diff
@@ -156,25 +166,15 @@ impl CallContractCommand {
...
-			},
-			Some(p) => p,
-		};
+		let contract_path = self.path.get_or_insert_with(|| {
+			let input_path: String = cli
+				.input("Where is your project or metadata file located?")
+				.placeholder("./")
+				.default_input("./")
+				.interact()
+				.unwrap();
```
evilrobot-01, line 175: Reason for the unwrap? Should we not check it and return an error, or at the very least use .expect() to provide some indication as to why it failed?

evilrobot-01 (changes requested):
Looks good, think there can be some quick improvements to the code to improve readability and reduce complexity.

As mentioned elsewhere, it seems as though any contract artifact can be used (.contract, .json, .wasm), so perhaps the terminology should be adapted accodingly?

al3mart (commented):
Used this contract to test it out - https://github.com/r0gue-io/faucet-contract

```
⚙  pop call contract --path ./metadata/faucet.contract --contract 5CwTqjdN28XfcTGsaC5VbyxP8wHAfbaKMwhhGA9bNJjaSDUJ --message drip --url wss://rpc1.paseo.popnetwork.xyz/ --suri //Alice --dry_run
│  
└  Anyhow error: Error invoking `cargo metadata` for ./metadata/faucet.contract/Cargo.toml
```

It seems that we are failing to resolve paths correctly when a location to an artifact is passed directly instead of pointing to a directory.

Maybe related with the pointers @evilrobot-01 has left.

Regarding the discussion between metadata or artifact. Knowing that the CLI should work just fine regardless of being pointed to a metadata file or the code or the bundle artifact. Maybe artifact does a better job fitting all scenarios.
We also use "artifacts" after a successful contract build.

crates/pop-contracts/src/utils/metadata.rs
```diff
@@ -42,14 +42,15 @@ pub enum FunctionType {
...
 pub fn get_messages(path: &Path) -> Result<Vec<ContractFunction>, Error> {
-	let cargo_toml_path = match path.ends_with("Cargo.toml") {
-		true => path.to_path_buf(),
-		false => path.join("Cargo.toml"),
+	let contract_artifacts = if path.is_dir() || path.ends_with("Cargo.toml") {
+		let cargo_toml_path =
+			if path.ends_with("Cargo.toml") { path.to_path_buf() } else { path.join("Cargo.toml") };
+		ContractArtifacts::from_manifest_or_file(Some(&cargo_toml_path), None)?
+	} else {
+		ContractArtifacts::from_manifest_or_file(None, Some(&path.to_path_buf()))?
```
al3mart, line 52: If I use a contract metadata location in the argument `--path`, `from_manifest_or_file` seem to still try to locate the manifest.

crates/pop-cli/src/commands/call/contract.rs
AlexD10S: It makes sense to move the call to `ensure_contract_built` after the user has input the path.

crates/pop-contracts/src/utils/metadata.rs
AlexD10S: Good catch! Fixed here https://github.com/r0gue-io/pop-cli/pull/347/commits/53f350505f6d33eadf93f505c4640b4c5350b509 and added some tests

crates/pop-cli/src/commands/call/contract.rs
```diff
@@ -842,7 +894,7 @@ mod tests {
 			dry_run: false,
 			execute: false,
 			dev_mode: false,
-		}.configure(&mut cli, false).await, Err(message) if message.to_string().contains("Unable to fetch contract metadata: Failed to find any contract artifacts in target directory.")));
```
AlexD10S, line 845: Now the `ensure_contract_built` has been moved the error: `Failed to find any contract artifacts in target directory.` would never reach. But still a valid test to see what happens if invalid artifacts in the target folder

crates/pop-cli/src/common/build.rs
```diff
@@ -14,9 +14,13 @@ pub fn has_contract_been_built(path: Option<&Path>) -> bool {
 		Ok(manifest) => manifest,
 		Err(_) => return false,
 	};
-	let contract_name = manifest.package().name();
-	project_path.join("target/ink").exists() &&
-		project_path.join(format!("target/ink/{}.contract", contract_name)).exists()
+	if manifest.package.as_ref().is_none() {
```
AlexD10S, line 17: As mentioned in this comment: https://github.com/r0gue-io/pop-cli/pull/347/files#r1854713654

crates/pop-cli/src/commands/call/contract.rs
```diff
@@ -119,8 +117,15 @@ impl CallContractCommand {
...
 	async fn ensure_contract_built(&self, cli: &mut impl Cli) -> Result<()> {
-		// Check if build exists in the specified "Contract build directory"
+		// The path is expected to be set. If it is not, exit early without attempting to build the contract.
+		let Some(path) = self.path.as_deref() else { return Ok(()) };
+		// Check if the path is a file or the build exists in the specified "Contract build
+		// directory"
+		if !path.is_dir() || has_contract_been_built(self.path.as_deref()) {
+			return Ok(());
+		}
 		if !has_contract_been_built(self.path.as_deref()) {
```
evilrobot-01, line 129: You no longer need this as it is checked on line 126, eliminating the if and indentation.

crates/pop-cli/src/commands/call/contract.rs
```diff
@@ -119,8 +117,15 @@ impl CallContractCommand {
...
 
 	/// Checks if the contract has been built; if not, builds it.
+	/// If the path is a contract artifact file, skips the build process
 	async fn ensure_contract_built(&self, cli: &mut impl Cli) -> Result<()> {
-		// Check if build exists in the specified "Contract build directory"
+		// The path is expected to be set. If it is not, exit early without attempting to build the contract.
+		let Some(path) = self.path.as_deref() else { return Ok(()) };
+		// Check if the path is a file or the build exists in the specified "Contract build
+		// directory"
+		if !path.is_dir() || has_contract_been_built(self.path.as_deref()) {
```
evilrobot-01, line 126:
```suggestion
		if path.is_file() || has_contract_been_built(self.path.as_deref()) {
```
Improves readability.

crates/pop-cli/src/commands/call/contract.rs
```diff
@@ -508,6 +508,52 @@ mod tests {
...
+		};
+		call_config.configure(&mut cli, false).await?;
+		assert_eq!(call_config.display(), format!(
+			"pop call contract --path {} --contract 15XausWjFLBBFLDXUSBRfSfZk25warm4wZRV4ZxhZbfvjrJm --message flip --gas 100 --proof_size 10 --url wss://rpc1.paseo.popnetwork.xyz/ --suri //Alice --dry_run",
+			current_dir.join("pop-contracts/tests/files/testing.contract").display().to_string(),
+		));
+		// Contract deployed on Pop Network testnet, test dry-run
+		call_config.execute_call(&mut cli, false).await?;
+
+		// From .json file
```
evilrobot-01, line 544: A .wasm file is also supported - is it quick to duplicate this code block to add a test case for .wasm?

crates/pop-cli/src/commands/call/contract.rs
```diff
@@ -828,6 +874,19 @@ mod tests {
 	#[tokio::test]
 	async fn guide_user_to_call_contract_fails_not_build() -> Result<()> {
 		let temp_dir = new_environment("testing")?;
+		let mut current_dir = env::current_dir().expect("Failed to get current directory");
+		current_dir.pop();
+		// Create invalid `.json` and `.contract` files in the mock build directory and avoid
```
evilrobot-01, line 879: Why not .wasm too?

crates/pop-cli/src/common/build.rs
evilrobot-01: Can simplify to use 

```
manifest.package.map(|p| { 
  project_path.join(format!("target/ink/{}.contract", p.name())).exists()
 })
.unwrap_or_default()
```

evilrobot-01 (changes requested):
Some suggestions for minor improvements. Also, the .wasm artifact doesnt have any consideration in the tests.

crates/pop-cli/src/commands/call/contract.rs
AlexD10S: Yes.

crates/pop-cli/src/commands/call/contract.rs
```diff
@@ -901,7 +910,26 @@ mod tests {
 			dry_run: false,
 			execute: false,
 			dev_mode: false,
-		}.configure(&mut cli, false).await, Err(message) if message.to_string().contains("Unable to fetch contract metadata")));
+		};
+		let mut cli = MockCli::new();
+		assert!(
```
evilrobot-01, line 915: For future reference, you could have put all the variants into a vec/array and looped over it to avoid all this duplicate code.

evilrobot-01 (approved):
Looks good, thanks for your patience!

al3mart (approved):
Quite nice! Thanks.

Now users can provide any contract artifact and perform their calls without problems!

PR #348, opened 2024-11-25
feat: submit extrinsic from call_data

AlexD10S:
This PR is part of the `pop call parachain` feature. It allows submitting an extrinsic by providing the call data with the flag `--call`.

Example for testing:
`pop call parachain --call 0x00000410 --url ws://localhost:9944/ --suri //Alice`

(In this example, the call data corresponds to a remark extrinsic.)

Additionally the function `find_free_port` introduced in https://github.com/r0gue-io/pop-cli/pull/360 has been moved to `pop_common` crate in order to be used in the parachain integration tests, to test the `pop call parachain` feature.

[sc-1514]


crates/pop-cli/src/commands/call/parachain.rs
```diff
@@ -188,6 +201,44 @@ impl CallParachainCommand {
...
+			return Ok(());
+		}
+		let spinner = cliclack::spinner();
+		spinner.start("Signing and submitting the extrinsic, please wait...");
+		let result = sign_and_submit_extrinsic_with_call_data(api.clone(), call_data, &suri)
+			.await
+			.map_err(|err| anyhow!("{}", format!("{err:?}")))?;
+
+		spinner.stop(format!("Extrinsic submitted successfully with hash: {:?}", result));
+		display_message("Parachain calling complete.", true, cli)?;
```
evilrobot-01, line 238: I think we should avoid using 'parachain' as the narrative is now moving to 'rollups'. Suggest we also add aliases to the command, but that can be done later. If parachain is littered everywhere, it will just make for more work in the future.

Perhaps something generic and simple like 'Call complete.'

crates/pop-cli/src/commands/call/parachain.rs
```diff
@@ -646,6 +700,27 @@ mod tests {
 		cli.verify()
 	}
 
+	#[tokio::test]
+	async fn user_cancel_send_extrinsic_from_call_data_works() -> Result<()> {
+		let api = set_up_api("wss://rpc1.paseo.popnetwork.xyz").await?;
```
evilrobot-01, line 705:
```suggestion
		let client = set_up_client("wss://rpc1.paseo.popnetwork.xyz").await?;
```
See above for justification. I assume that this is in the base PR, so happy to adjust it there instead and just rebase.

crates/pop-parachains/src/call/mod.rs
```diff
@@ -74,6 +74,47 @@ pub fn encode_call_data(
...
+	) -> Result<(), subxt::ext::subxt_core::Error> {
+		out.extend_from_slice(&self.0);
+		Ok(())
+	}
+}
+
+/// Signs and submits a given extrinsic to the blockchain.
+///
+/// # Arguments
+/// * `api` - Reference to an `OnlineClient` connected to the chain.
```
evilrobot-01, line 95:
```suggestion
/// * `client` - Reference to an `OnlineClient` connected to the chain.
```

crates/pop-parachains/src/call/mod.rs
```diff
@@ -74,6 +74,47 @@ pub fn encode_call_data(
...
+/// * `call_data` -SCALE encoded bytes representing the call data of the transaction.
```
evilrobot-01, line 96:
```suggestion
/// * `call_data` - SCALE encoded bytes representing the call data of the transaction.
```

crates/pop-parachains/src/call/mod.rs
```diff
@@ -74,6 +74,47 @@ pub fn encode_call_data(
...
+/// * `suri` - The secret URI (e.g., mnemonic or private key) for signing the extrinsic.
+pub async fn sign_and_submit_extrinsic_with_call_data(
+	api: OnlineClient<SubstrateConfig>,
```
evilrobot-01, line 99:
```suggestion
	client: OnlineClient<SubstrateConfig>,
```

crates/pop-parachains/src/errors.rs
```diff
@@ -11,6 +11,8 @@ pub enum Error {
 	AnyhowError(#[from] anyhow::Error),
 	#[error("Failed to establish a connection to the API: {0}")]
 	ApiConnectionFailure(String),
+	#[error("Failed to decode call data. {0}")]
```
evilrobot-01, line 14: Missing doc comment - lets try to ensure that we have them for any new variants so we can slowly get full doc coverage for errors.

evilrobot-01 (changes requested):
Mostly minor suggestions to improve terminology and efficiency.

Note: I did not review any of the free port stuff as I feel that should not have been included in this PR, but rather a PR to base feature branch.

crates/pop-cli/src/commands/call/parachain.rs
```diff
@@ -34,6 +35,9 @@ pub struct CallParachainCommand {
 	/// - with a password "//Alice///SECRET_PASSWORD"
 	#[arg(short, long)]
 	suri: Option<String>,
+	/// SCALE encoded bytes representing the call data of the transaction.
```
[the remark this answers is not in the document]
AlexD10S: Yes.

crates/pop-cli/src/commands/call/parachain.rs
```diff
@@ -188,6 +201,44 @@ impl CallParachainCommand {
 		}
 	}
 
+	// Sends an extrinsic to the chain using the call data.
+	async fn send_extrinsic_from_call_data(
+		&mut self,
+		api: &OnlineClient<SubstrateConfig>,
```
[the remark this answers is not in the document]
AlexD10S: The name was chosen based on its usage in Subxt ([example](https://github.com/paritytech/subxt/blob/master/subxt/examples/block_decoding_dynamic.rs#L7)). However, looking at type's comment, I agree that `client` is a better name.

I refactored all related naming across the main feature branch.

crates/pop-cli/src/commands/call/parachain.rs
```diff
@@ -47,41 +51,50 @@ impl CallParachainCommand {
...
-			// Display the configured call.
-			cli.info(call.display(&chain))?;
-			// Prepare the extrinsic.
-			let tx = match call.prepare_extrinsic(&chain.api, &mut cli).await {
-				Ok(api) => api,
-				Err(e) => {
+		// Execute the call if call_data is provided.
+		if let Some(call_data) = self.call_data.as_ref().cloned() {
+			if let Err(e) =
+				self.send_extrinsic_from_call_data(&chain.api, &call_data, &mut cli::Cli).await
```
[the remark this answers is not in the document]
AlexD10S: Same as above

crates/pop-cli/src/commands/call/parachain.rs
```diff
@@ -47,41 +51,50 @@ impl CallParachainCommand {
...
+			{
+				display_message(&e.to_string(), false, &mut cli::Cli)?;
+			}
```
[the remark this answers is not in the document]
AlexD10S: You are right, much better

crates/pop-cli/src/commands/call/parachain.rs
```diff
@@ -47,41 +51,50 @@ impl CallParachainCommand {
...
-				},
-			};
-			// Display the configured call.
-			cli.info(call.display(&chain))?;
-			// Prepare the extrinsic.
-			let tx = match call.prepare_extrinsic(&chain.api, &mut cli).await {
-				Ok(api) => api,
-				Err(e) => {
+		// Execute the call if call_data is provided.
+		if let Some(call_data) = self.call_data.as_ref().cloned() {
```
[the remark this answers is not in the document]
AlexD10S: Adjust to eliminate unnecessary clones

crates/pop-cli/src/commands/call/parachain.rs
```diff
@@ -188,6 +201,44 @@ impl CallParachainCommand {
...
+	// Sends an extrinsic to the chain using the call data.
+	async fn send_extrinsic_from_call_data(
+		&mut self,
+		api: &OnlineClient<SubstrateConfig>,
+		call_data: &str,
+		cli: &mut impl cli::traits::Cli,
+	) -> Result<()> {
+		// Resolve who is signing the extrinsic.
+		let suri = match self.clone().suri {
+			Some(suri) => suri,
```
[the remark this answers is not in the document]
evilrobot-01: Nicely done!

crates/pop-cli/src/commands/call/parachain.rs
```diff
@@ -639,6 +692,29 @@ mod tests {
 		cli.verify()
 	}
 
+	#[tokio::test]
+	async fn user_cancel_submit_extrinsic_from_call_data_works() -> Result<()> {
+		let client = set_up_client("wss://rpc1.paseo.popnetwork.xyz").await?;
+		let call_config = CallParachainCommand {
```
evilrobot-01, line 698: For future, deriving Default on the command might make all this initialisation much simpler via `CallParachainCommand { ..Default::default() }`

crates/pop-parachains/src/call/mod.rs
```diff
@@ -73,6 +73,47 @@ pub fn encode_call_data(
...
+
+/// Signs and submits a given extrinsic.
+///
+/// # Arguments
+/// * `client` - Reference to an `OnlineClient` connected to the chain.
+/// * `call_data` - SCALE encoded bytes representing the call data of the transaction.
+/// * `suri` - The secret URI (e.g., mnemonic or private key) for signing the extrinsic.
+pub async fn sign_and_submit_extrinsic_with_call_data(
+	client: OnlineClient<SubstrateConfig>,
+	call_data: &str,
```
evilrobot-01, line 99: Minor refactor: I think we should change this type to Vec<u8> and refactor `hex::decode(call_data.trim_start_matches("0x")).map_err(|e| Error::CallDataDecodingError(e.to_string()))?;` into a function. As this is a library crate, there is the very likely case that someone has some call data bytes which they want to submit - they then need to take those bytes and convert them to a hex string first, only for them to be converted back. 

In general, strings and string handling should be handled by the UI as that is the place that is receiving user input. The UI should convert it to the type exposed by the library, in this case bytes. Another implementation may use this crate to submit an extrinsic based on bytes received from a websocket connection for example.

evilrobot-01 (changes requested):
Looking great!

Just one minor refactor and then I think we are good to go.

evilrobot-01 (approved):

PR #349, opened 2024-11-25
feat: flag sudo to wrap extrinsic

AlexD10S:
This PR is part of the `pop call parachain` feature. It adds the ability to submit an extrinsic as a sudo operation using the `--sudo` flag, wrapping the selected extrinsic with sudo privileges.

_**For testing:**_
You can run `pop call parachain` interactively to select the extrinsic you want to execute and choose to run it as a sudo at the end. 
Alternatively, you can specify it directly in the command line with:
`pop call parachain --pallet System --extrinsic remark --args "0x11" --url ws://localhost:9944/ --suri //Alice --sudo`

[sc-1516]

### DIFF
--- a/crates/pop-cli/src/commands/call/parachain.rs
+++ b/crates/pop-cli/src/commands/call/parachain.rs
@@ -94,8 +94,9 @@ impl CallParachainCommand {
 				break;
 			}
 
-			if !prompt_to_repeat_call ||
-				!cli.confirm("Do you want to perform another call?")
+			if !prompt_to_repeat_call
+				|| !cli
+					.confirm("Do you want to perform another call?")
 					.initial_value(false)
 					.interact()?
 			{
@@ -158,8 +159,9 @@ impl CallParachainCommand {
 
 			// Resolve extrinsic.
 			let extrinsic = match self.extrinsic {
-				Some(ref extrinsic_name) =>
-					find_extrinsic_by_name(&chain.pallets, &pallet.name, extrinsic_name).await?,
+				Some(ref extrinsic_name) => {
+					find_extrinsic_by_name(&chain.pallets, &pallet.name, extrinsic_name).await?
+				},
 				None => {
 					let mut prompt_extrinsic = cli.select("Select the extrinsic to call:");
 					for extrinsic in &pallet.extrinsics {
@@ -204,8 +206,9 @@ impl CallParachainCommand {
 			// Resolve who is signing the extrinsic.
 			let suri = match self.suri.as_ref() {
 				Some(suri) => suri.clone(),
-				None =>
-					cli.input("Signer of the extrinsic:").default_input(DEFAULT_URI).interact()?,
+				None => {
+					cli.input("Signer of the extrinsic:").default_input(DEFAULT_URI).interact()?
+				},
 			};
 
 			return Ok(CallParachain {
@@ -232,8 +235,9 @@ impl CallParachainCommand {
 			None => &cli.input("Signer of the extrinsic:").default_input(DEFAULT_URI).interact()?,
 		};
 		cli.info(format!("Encoded call data: {}", call_data))?;
-		if !self.skip_confirm &&
-			!cli.confirm("Do you want to submit the extrinsic?")
+		if !self.skip_confirm
+			&& !cli
+				.confirm("Do you want to submit the extrinsic?")
 				.initial_value(true)
 				.interact()?
 		{
@@ -268,11 +272,11 @@ impl CallParachainCommand {
 
 	// Function to check if all required fields are specified.
 	fn requires_user_input(&self) -> bool {
-		self.pallet.is_none() ||
-			self.extrinsic.is_none() ||
-			self.args.is_empty() ||
-			self.url.is_none() ||
-			self.suri.is_none()
+		self.pallet.is_none()
+			|| self.extrinsic.is_none()
+			|| self.args.is_empty()
+			|| self.url.is_none()
+			|| self.suri.is_none()
 	}
 
 	/// Replaces file arguments with their contents, leaving other arguments unchanged.
@@ -359,8 +363,9 @@ impl CallParachain {
 		tx: DynamicPayload,
 		cli: &mut impl Cli,
 	) -> Result<()> {
-		if !self.skip_confirm &&
-			!cli.confirm("Do you want to submit the extrinsic?")
+		if !self.skip_confirm
+			&& !cli
+				.confirm("Do you want to submit the extrinsic?")
 				.initial_value(true)
 				.interact()?
 		{
--- a/crates/pop-cli/src/commands/call/parachain.rs
+++ b/crates/pop-cli/src/commands/call/parachain.rs
@@ -6,10 +6,10 @@ use crate::cli::{self, traits::*};
 use anyhow::{anyhow, Result};
 use clap::Args;
 use pop_parachains::{
-	construct_extrinsic, decode_call_data, encode_call_data, find_extrinsic_by_name,
-	find_pallet_by_name, parse_chain_metadata, set_up_client, sign_and_submit_extrinsic,
-	sign_and_submit_extrinsic_with_call_data, supported_actions, Action, DynamicPayload, Extrinsic,
-	OnlineClient, Pallet, Param, SubstrateConfig,
+	construct_extrinsic, construct_sudo_extrinsic, decode_call_data, encode_call_data,
+	find_extrinsic_by_name, find_pallet_by_name, parse_chain_metadata, set_up_client,
+	sign_and_submit_extrinsic, sign_and_submit_extrinsic_with_call_data, supported_actions, Action,
+	DynamicPayload, Extrinsic, OnlineClient, Pallet, Param, SubstrateConfig,
 };
 use url::Url;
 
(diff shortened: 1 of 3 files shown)


crates/pop-cli/src/commands/call/parachain.rs
```diff
@@ -41,6 +41,9 @@ pub struct CallParachainCommand {
 	/// Automatically signs and submits the extrinsic without prompting for confirmation.
 	#[arg(short('y'), long)]
 	skip_confirm: bool,
+	/// Authenticates the sudo key and dispatches a function call with `Root` origin.
+	#[arg(name = "sudo", short = 'S', long)]
```
evilrobot-01, line 45:
```suggestion
	#[arg(short = 'S', long)]
```
Name _may_ be unnecessary based on the field name being the same.

crates/pop-cli/src/commands/call/parachain.rs
```diff
@@ -191,12 +194,21 @@ impl CallParachainCommand {
 					cli.input("Signer of the extrinsic:").default_input(DEFAULT_URI).interact()?,
 			};
 
+			// Prompt the user to confirm if they want to execute the call with sudo privileges.
+			if !self.sudo {
+				self.sudo = cli
+					.confirm("Would you like to execute this operation with sudo privileges?")
```
evilrobot-01, line 200:
```suggestion
					.confirm("Would you like to dispatch this function call with `Root` origin?")
```
Based on the parameter description above and the docs at https://github.com/paritytech/polkadot-sdk/tree/master/substrate/frame/sudo

crates/pop-cli/src/commands/call/parachain.rs
```diff
@@ -191,12 +194,21 @@ impl CallParachainCommand {
 					cli.input("Signer of the extrinsic:").default_input(DEFAULT_URI).interact()?,
 			};
 
+			// Prompt the user to confirm if they want to execute the call with sudo privileges.
```
evilrobot-01, line 197:
```suggestion
			// Prompt the user to confirm if they want to execute the call via sudo.
```

crates/pop-cli/src/commands/call/parachain.rs
evilrobot-01: Perhaps one for the future, but this prompt only makes sense if the sudo pallet is mentioned in the chain metadata. There is also an assumption here that it is named 'Sudo', when in fact any chain could have chosen a different name.

crates/pop-cli/src/commands/call/parachain.rs
```diff
@@ -299,6 +315,8 @@ impl CallParachain {
 				return Err(anyhow!("Error: {}", e));
 			},
 		};
+		// If sudo is enabled, wrap the call in a sudo call.
```
evilrobot-01, line 318:
```suggestion
		// If sudo is required, wrap the call in a sudo call.
```

crates/pop-parachains/src/call/mod.rs
```diff
@@ -36,6 +36,14 @@ pub async fn construct_extrinsic(
 	Ok(subxt::dynamic::tx(pallet_name, extrinsic_name, parsed_args))
 }
 
+/// Constructs a Sudo extrinsic.
+///
+/// # Arguments
+/// * `tx`: The transaction to be executed with sudo privileges.
```
evilrobot-01, line 42: Terminology is off, 'sudo privileges' doesnt sound right. See command param description and above comment for more info.

crates/pop-parachains/src/call/mod.rs
```diff
@@ -166,4 +174,21 @@ mod tests {
 		assert_eq!(encode_call_data(&api, &extrinsic)?, "0x00000411");
 		Ok(())
 	}
+
+	#[tokio::test]
+	async fn construct_sudo_extrinsic_works() -> Result<()> {
+		let extrinsic = construct_extrinsic(
+			"Balances",
+			"transfer_allow_death",
```
evilrobot-01, line 182: Whilst it probably doesnt matter, especially in terms of the test, it would be ideal to use one of the force_ dispatchables which actually requires root origin.

evilrobot-01 (changes requested):
Looks good, just some more cosmetic suggestions for improvement, especially around terminology.

al3mart (commented):
Looking good!

I don't have much more to add on top of what Frank has already pointed out.
Agreed a better testing call will be one requiring root origin.

Happy to approve once the feedback has been addressed.

crates/pop-parachains/src/call/mod.rs
AlexD10S: Changed to `* `tx`: The transaction payload representing the function call to be dispatched with `Root` privileges.`

crates/pop-parachains/src/call/mod.rs
AlexD10S: In this specific test it doesn't really matter because is only constructing the extrinsic,  but changed.

crates/pop-cli/src/commands/call/parachain.rs
AlexD10S: Good catch.
It makes sense to check if Sudo exist before prompt the user, added in this commit https://github.com/r0gue-io/pop-cli/pull/349/commits/935744ac60b0838d24f53362d40c34437e2f2bb5

However it is making the assumption that it is named 'Sudo', but not sure how we can improve this.

crates/pop-cli/src/commands/call/parachain.rs
```diff
@@ -240,11 +248,34 @@ impl CallParachainCommand {
...
+					self.sudo = cli
+						.confirm(
+							"Would you like to dispatch this function call with `Root` origin?",
+						)
+						.initial_value(false)
+						.interact()?;
+				},
+			Err(_) =>
+				if self.sudo {
+					cli.warning("NOTE: sudo extrinsic is not supported by the chain. Ignoring `--sudo` flag.")?;
```
evilrobot-01, line 266:
```suggestion
					cli.warning("NOTE: sudo is not supported by the chain. Ignoring `--sudo` flag.")?;
```

crates/pop-cli/src/commands/call/parachain.rs
```diff
@@ -285,6 +316,9 @@ struct CallParachain {
 	suri: String,
 	/// Whether to automatically sign and submit the extrinsic without prompting for confirmation.
 	skip_confirm: bool,
+	/// Whether to authenticate with the sudo key and dispatch the function call with `Root`
```
evilrobot-01, line 319:
```suggestion
	/// Whether to dispatch the function call with `Root` origin.
```
Internal struct so a simpler comment is fine imo.

crates/pop-cli/src/commands/call/parachain.rs
```diff
@@ -628,6 +672,7 @@ mod tests {
 		assert_eq!(call_parachain.extrinsic.name, "place_order_allow_death");
 		assert_eq!(call_parachain.args, ["10000".to_string(), "2000".to_string()].to_vec());
 		assert_eq!(call_parachain.suri, "//Bob");
+		assert!(!call_config.sudo);
```
evilrobot-01, line 675:
```suggestion
		assert!(!call_parachain.sudo);
```

crates/pop-cli/src/commands/call/parachain.rs
evilrobot-01: Likewise. I like that you check the dispatchable name too!

crates/pop-cli/src/commands/call/parachain.rs
AlexD10S: Good catch, solved in https://github.com/r0gue-io/pop-cli/pull/349/commits/924662d302ac63a1e6d64a36d3849cf90fa06250

crates/pop-cli/src/commands/call/parachain.rs
AlexD10S: Improved in https://github.com/r0gue-io/pop-cli/pull/349/commits/620252fefff47e47f786f3ba4f5dbbfeba00bd74

crates/pop-cli/src/commands/call/parachain.rs
AlexD10S: Improved in https://github.com/r0gue-io/pop-cli/pull/349/commits/620252fefff47e47f786f3ba4f5dbbfeba00bd74

crates/pop-cli/src/commands/call/parachain.rs
```diff
@@ -240,11 +248,34 @@ impl CallParachainCommand {
 		Ok(())
 	}
 
+	/// Checks if the chain has the Sudo pallet and prompt the user to confirm if they want to
+	/// execute the call via sudo.
+	async fn configure_sudo(&mut self, chain: &Chain, cli: &mut impl Cli) -> Result<()> {
```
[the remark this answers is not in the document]
AlexD10S: It's a nice-to-have check, and I don't see it as a blocker either. If you can get a PR for this, I'll gladly review and merge it.

al3mart (approved):
No requests from my side here. It looks great!

An idea for future work, maybe when we tackle wrapped calls.
Might be interesting to prompt the call data of the wrapped calls too, not only the one that is going to be dispatched. But needs future consideration.

AlexD10S:
> An idea for future work, maybe when we tackle wrapped calls. Might be interesting to prompt the call data of the wrapped calls too, not only the one that is going to be dispatched. But needs future consideration.

The idea is to prompt the user with the `call_data` in case they want to submit it using another source or validate it before submit it. Both actions can already be performed using the wrapped call `call_data`.

However, if you think this could be interesting for any specific use case, it’s an easy fix. I’d just need to add a line here: [code link](https://github.com/r0gue-io/pop-cli/blob/feat-call-extrinsic-sudo-wrap/crates/pop-cli/src/commands/call/parachain.rs#L366) before construct the sudo call.

PR #352, opened 2024-11-26
bump zombienet version to `v0.2.18`

pepoviola:
Fix `fn nodes()` visibility.

Thanks!


AlexD10S:
Thanks for that @pepoviola 
We use `cargo-deny` https://embarkstudios.github.io/cargo-deny/cli/check.html  to ensure that the dependencies we add are free of vulnerabilities. However, it seems that the check is currently failing in the CI pipeline: https://embarkstudios.github.io/cargo-deny/cli/check.html.

You can run it locally with `cargo-deny check`

AlexD10S:
There seems to be a version mismatch with `subxt`. We’ll need to update the versions in some dependencies of `pop-cli` before we can use the latest version of zombienet.

```
subxt-signer = { version = "0.38.0", features = ["subxt", "sr25519"] }
subxt = "0.38.0"
sp-core = "32.0.0"
sp-weights = "31.0.0"
contract-build = "5.0.0"
contract-extrinsics = "5.0.0"
contract-transcode = "5.0.0"
scale-info = { version = "2.11.4", default-features = false, features = ["derive"] }
scale-value = { version = "0.17.0", default-features = false, features = ["from-string", "parser-ss58"] }
```

AlexD10S (approved):
Thank you for your patience in fixing everything and for the amazing `Zombienet-SDK`.

PR #354, opened 2024-11-27
feat: production build

Daanvdplas:
Adds the build profile to pop cli

### DIFF
--- a/crates/pop-cli/src/commands/build/mod.rs
+++ b/crates/pop-cli/src/commands/build/mod.rs
@@ -30,7 +30,7 @@ pub(crate) struct BuildArgs {
 	#[arg(short = 'p', long)]
 	pub(crate) package: Option<String>,
 	/// For production, always build in release mode to exclude debug features.
-	#[clap(short = 'r', long, conflicts_with = "profile")]
+	#[clap(short, long, conflicts_with = "profile")]
 	pub(crate) release: bool,
 	/// Build profile [default: debug].
 	#[clap(long, value_enum)]
@@ -145,7 +145,7 @@ mod tests {
 		for package in [None, Some(name.to_string())] {
 			for release in [true, false] {
 				for profile in Profile::VARIANTS {
-					let profile = if release { Profile::Release } else { Profile::Debug };
+					let profile = if release { Profile::Release } else { profile.clone() };
 					let project = if package.is_some() { "package" } else { "project" };
 					let mut cli = MockCli::new()
 						.expect_intro(format!("Building your {project}"))
--- a/crates/pop-cli/src/commands/build/mod.rs
+++ b/crates/pop-cli/src/commands/build/mod.rs
@@ -8,6 +8,7 @@ use duct::cmd;
 use std::path::PathBuf;
 #[cfg(feature = "parachain")]
 use {parachain::BuildParachainCommand, spec::BuildSpecCommand};
+use pop_common::Profile;
 
 #[cfg(feature = "contract")]
 pub(crate) mod contract;
@@ -28,9 +29,9 @@ pub(crate) struct BuildArgs {
 	/// The package to be built.
 	#[arg(short = 'p', long)]
 	pub(crate) package: Option<String>,
-	/// For production, always build in release mode to exclude debug features.
-	#[clap(short, long)]
-	pub(crate) release: bool,
+	/// Build profile [default: debug].
+	#[clap(long, value_enum)]
+	pub(crate) profile: Option<Profile>,
 	/// Parachain ID to be used when generating the chain spec files.
 	#[arg(short = 'i', long = "id")]
 	#[cfg(feature = "parachain")]
@@ -62,7 +63,11 @@ impl Command {
 		#[cfg(feature = "contract")]
 		if pop_contracts::is_supported(args.path.as_deref())? {
 			// All commands originating from root command are valid
-			BuildContractCommand { path: args.path, release: args.release, valid: true }
+			let release = match args.profile.unwrap_or(Profile::Debug) {
+				Profile::Debug => false,
+				_ => true,
+			};
+			BuildContractCommand { path: args.path, release, valid: true }
 				.execute()?;
 			return Ok("contract");
 		}
@@ -74,7 +79,7 @@ impl Command {
 			BuildParachainCommand {
 				path: args.path,
 				package: args.package,
-				release: args.release,
+				profile: args.profile,
 				id: args.id,
 				valid: true,
 			}
@@ -101,13 +106,15 @@ impl Command {
 			_args.push("--package");
 			_args.push(package)
 		}
-		if args.release {
+		let profile = args.profile.unwrap_or(Profile::Debug);
+		if profile == Profile::Release {
 			_args.push("--release");
+		} else if profile == Profile::Production {
+			_args.push("--profile=production");
 		}
 		cmd("cargo", _args).dir(args.path.unwrap_or_else(|| "./".into())).run()?;
 
-		let mode = if args.release { "RELEASE" } else { "DEBUG" };
-		cli.info(format!("The {project} was built in {mode} mode."))?;
+		cli.info(format!("The {project} was built in {} mode.", profile))?;
 		cli.outro("Build completed successfully!")?;
 		Ok(project)
 	}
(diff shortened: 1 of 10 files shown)


crates/pop-cli/src/commands/build/mod.rs
```diff
@@ -101,46 +106,69 @@ impl Command {
...
 #[cfg(test)]
 mod tests {
+	use std::fs;
+	use std::fs::write;
+	use std::path::Path;
 	use super::*;
 	use cli::MockCli;
+	use strum::VariantArray;
+
+	fn add_production_profile(project: &Path) -> anyhow::Result<()> {
```
Daanvdplas, line 132: Happy to hear where I can place this function as it now exists in three places.

AlexD10S: I had the same question in the `pop call contract`. The solution was to have a public function in the crate `pop_contracts` under a testing module that can be then used by the` pop-cli` https://github.com/r0gue-io/pop-cli/pull/306/files#diff-1735e49d7426a2bb8a539485425778554b43ef1ca455b26bd8138b69c69b18aeR27

crates/pop-cli/src/commands/build/mod.rs
```diff
@@ -28,9 +29,9 @@ pub(crate) struct BuildArgs {
 	/// The package to be built.
 	#[arg(short = 'p', long)]
 	pub(crate) package: Option<String>,
-	/// For production, always build in release mode to exclude debug features.
-	#[clap(short, long)]
-	pub(crate) release: bool,
+	/// Build profile [default: debug].
+	#[clap(long, value_enum)]
+	pub(crate) profile: Option<Profile>,
```
al3mart, line 34: https://github.com/r0gue-io/pop-cli/blob/fix/build-spec/crates/pop-cli/tests/parachain.rs#L13

Should be updated accordingly to this change.

crates/pop-cli/src/commands/build/spec.rs
```diff
@@ -40,20 +38,20 @@ pub(crate) enum ChainType {
 	// A development chain that runs mainly on one node.
 	#[default]
 	#[strum(
-		serialize = "Development",
+		serialize = "development",
```
al3mart, line 41: Needs the capitalization. Otherwise the command seems to complete successfully.

crates/pop-cli/src/commands/build/spec.rs
```diff
@@ -40,20 +38,20 @@ pub(crate) enum ChainType {
...
 		message = "Development",
 		detailed_message = "Meant for a development chain that runs mainly on one node."
 	)]
 	Development,
 	// A local chain that runs locally on multiple nodes for testing purposes.
 	#[strum(
-		serialize = "Local",
+		serialize = "local",
```
al3mart, line 48: Same as above

crates/pop-cli/src/commands/build/spec.rs
```diff
@@ -40,20 +38,20 @@ pub(crate) enum ChainType {
...
 		message = "Local",
 		detailed_message = "Meant for a local chain that runs locally on multiple nodes for testing purposes."
 	)]
 	Local,
 	// A live chain.
-	#[strum(serialize = "Live", message = "Live", detailed_message = "Meant for a live chain.")]
+	#[strum(serialize = "live", message = "Live", detailed_message = "Meant for a live chain.")]
```
al3mart, line 54: Same as above.

crates/pop-cli/src/commands/build/mod.rs
```diff
@@ -28,9 +29,9 @@ pub(crate) struct BuildArgs {
 	/// The package to be built.
 	#[arg(short = 'p', long)]
 	pub(crate) package: Option<String>,
-	/// For production, always build in release mode to exclude debug features.
-	#[clap(short, long)]
-	pub(crate) release: bool,
```
al3mart, line 33: This is a breaking change.
Users that were running with `pop build --release` won't be able to just use this newer version. Imagine scripts.

We should consider a deprecation message and maybe an alias from `--release` to `--profile=release` during at least one more release of the CLI.

al3mart (changes requested):
Way cleaner this way :+1: :

My biggest concern is that by removing the option `--release` it forces us to include a deprecation message, an alias or something so users or some warning in the release notes.

Looks good. Left a few nits.

crates/pop-cli/src/commands/build/mod.rs
AlexD10S: If we see it can be used in contracts then Ale's suggestion is the best approach.

crates/pop-cli/src/commands/build/mod.rs
```diff
@@ -62,7 +63,11 @@ impl Command {
 		#[cfg(feature = "contract")]
 		if pop_contracts::is_supported(args.path.as_deref())? {
 			// All commands originating from root command are valid
-			BuildContractCommand { path: args.path, release: args.release, valid: true }
+			let release = match args.profile.unwrap_or(Profile::Debug) {
```
AlexD10S, line 66: Do we want `Debug` as default? before it was `--release` the default value: https://github.com/r0gue-io/pop-cli/blob/main/crates/pop-cli/src/commands/build/spec.rs#L138

AlexD10S (changes requested):
The only change requested is based on Ale's feedback about keeping the `--release` flag as deprecated. 
Here’s how you can handle it:
1- Update the help message for the flag: https://github.com/r0gue-io/pop-cli/blob/main/crates/pop-cli/src/commands/test/contract.rs#L18
2- If the flag is used, display a warning to inform the user: https://github.com/r0gue-io/pop-cli/blob/main/crates/pop-cli/src/commands/test/contract.rs#L50
3- You can set the Profile to `release` if the flag was specified.

crates/pop-cli/src/commands/build/mod.rs
Daanvdplas: It will be used for parachains and builds, not for smart contracts as far as I'm aware. However, I see no reason not to place it in common.

@AlexD10S your suggestion doesn't use any feature flag? This means it is added to the compilation time of the users which doesn't seem ideal. I suggest we come up with a feature name that we can start to use in the repository

crates/pop-cli/src/commands/build/spec.rs
Daanvdplas: I don't understand. I changed this so that the user doesn't have to write "Development" but "development"

crates/pop-cli/src/commands/build/mod.rs
Daanvdplas: This is for smart contracts

AlexD10S: You are right, my mistake

crates/pop-cli/src/commands/build/mod.rs
Daanvdplas: That is a great suggestion, I will not add it to the templates as default, something we can consider in the future

crates/pop-cli/src/commands/build/spec.rs
al3mart: Yeah, so I'm not really familiar with the details of strum here. 
But the reason of the failure is that the the node or spec builder requires this to be "Development". So the pain point is when this variant is serialized into the spec.

Although, even if we have `serialize = "Development"` users will be able to use `--type development` just fine. It's a bit confusing.

evilrobot-01: Isnt there an ignore case flag on strum? https://docs.rs/strum/latest/strum/additional_attributes/index.html

crates/pop-cli/src/commands/build/mod.rs
evilrobot-01: An alias is ideal - the build command should have some familarity with cargo, allowing -r, --release or --profile release

evilrobot-01: It should not be deprecrated in my view.

AlexD10S: Not a bad idea to use the alias. 
 A suggestion would be to add the [conflicts_with](https://docs.rs/clap/latest/clap/struct.Arg.html#method.conflicts_with) rule to prevent users from using both the `--release `and `--profile=production` flags at the same time.
```
#[clap(short, long, conflicts_with="profile")]
pub(crate) release: bool,
```

al3mart:
I'll do the rebase here

crates/pop-cli/src/commands/build/mod.rs
Daanvdplas: I agree that it should not be deprecated for the normal build command and put it back in

crates/pop-cli/src/commands/build/mod.rs
```diff
@@ -62,19 +66,26 @@ impl Command {
...
+			let profile = match args.profile {
+				Some(profile) => profile,
+				None => args.release.into(),
+			};
 			// All commands originating from root command are valid
 			BuildParachainCommand {
 				path: args.path,
 				package: args.package,
-				release: args.release,
+				profile: Some(profile),
```
Daanvdplas, line 88: The `Option` requirement should be removed when removing the deprecated parachain build and contract command.

crates/pop-cli/src/commands/build/spec.rs
al3mart: This seems resolved now. Is there anything else you want to do here, @Daanvdplas ?

al3mart (approved):
All the feedback has been now addressed.

crates/pop-cli/src/commands/build/mod.rs
```diff
@@ -29,8 +30,11 @@ pub(crate) struct BuildArgs {
 	#[arg(short = 'p', long)]
 	pub(crate) package: Option<String>,
 	/// For production, always build in release mode to exclude debug features.
-	#[clap(short, long)]
+	#[clap(short = 'r', long, conflicts_with = "profile")]
```
AlexD10S, line 33: You caa use only `short`, it will already use `r`:` #[clap(short, long, conflicts_with = "profile")]`

crates/pop-common/src/manifest.rs
```diff
@@ -97,6 +98,46 @@ pub fn add_crate_to_workspace(workspace_toml: &Path, crate_path: &Path) -> anyho
 	Ok(())
 }
 
+use std::fs;
```
AlexD10S, line 101: This is not used. Run `cargo +nightly clippy` and fix the warnings like this one.

AlexD10S (commented):
Thank you for addressing all the changes! I’ve just left a couple of minor comments.

AlexD10S (approved):

crates/pop-common/src/manifest.rs
```diff
@@ -419,4 +460,43 @@ mod tests {
...
+            "#,
+		);
+
+		let binding = test_builder.workspace.expect("Workspace should exist");
+		let project_path = binding.path();
+		let cargo_toml_path = project_path.join("Cargo.toml");
+
+		// Call the function to add the production profile
+		let result = add_production_profile(project_path);
+		println!("{:?}", result);
```
AlexD10S, line 478: Remove `println` instruction

crates/pop-cli/src/commands/build/mod.rs
Daanvdplas: Thank you!

PR #355, opened 2024-11-27
refactor: structs

Daanvdplas:
During the implementation of the build spec I refactored the code by adding a struct which has all the defined call configurations. I applied it to the call parachain call too as it greatly improves the code's flow imo.


crates/pop-cli/src/commands/call/parachain.rs
```diff
@@ -6,207 +6,222 @@ use clap::Args;
...
-	#[clap(short('y'), long)]
+	#[arg(short('y'), long)]
 	skip_confirm: bool,
 }
 
 impl CallParachainCommand {
 	/// Executes the command.
 	pub(crate) async fn execute(mut self) -> Result<()> {
-		// Check if message specified via command line argument.
-		let prompt_to_repeat_call = self.extrinsic.is_none();
```
AlexD10S, line 45: This change makes sense. However, if the user has provided all the necessary fields, I don't think we should prompt them to repeat the call; this behavior should be reserved for the interactive UI.

For instance, if I execute the following extrinsic:
`pop call parachain --pallet Nfts --extrinsic mint --args "" --url ws://127.0.0.1:9944/ --suri //Alice,`
at the end, the UI continues to wait for a response, which feels unnecessary in this context (Think of a script using it).

We could address this by introducing a flag like `skip_confirm`, or by extending the logic we had before but verifying that all required fields are set instead of only the `extrinsic`. For example:
```rs
let prompt_to_repeat_call = self.pallet.is_none() 
    && self.extrinsic.is_none() 
    && self.args.is_some() 
    && self.url.is_some() 
    && self.suri.is_some();
```

AlexD10S: I'll add it

AlexD10S (approved):

PR #357, opened 2024-11-29
chore: refactor prepare_output_path

al3mart:
Refactor of `prepare_output_path` function to handle both directories and files passed as the output file / directory for the resulting specs.

Resolving conversation in: https://github.com/r0gue-io/pop-cli/pull/351#issuecomment-2506623640

### DIFF
--- a/crates/pop-cli/src/commands/build/spec.rs
+++ b/crates/pop-cli/src/commands/build/spec.rs
@@ -9,7 +9,7 @@ use crate::{
 	style::style,
 };
 use clap::{Args, ValueEnum};
-use cliclack::{spinner};
+use cliclack::spinner;
 use pop_common::Profile;
 use pop_parachains::{
 	binary_path, build_parachain, export_wasm_file, generate_genesis_state_file,
@@ -536,7 +536,6 @@ fn ensure_binary_exists(
 }
 
 // Prepare the output path provided.
-
 fn prepare_output_path(output_path: impl AsRef<Path>) -> anyhow::Result<PathBuf> {
 	let mut output_path = output_path.as_ref().to_path_buf();
 	// Convert to string to check for trailing slash.
@@ -573,8 +572,6 @@ fn prepare_output_path(output_path: impl AsRef<Path>) -> anyhow::Result<PathBuf>
 	}
 	Ok(output_path)
 }
-
-
 
 #[cfg(test)]
 mod tests {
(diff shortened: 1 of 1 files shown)


crates/pop-cli/src/commands/build/spec.rs
```diff
@@ -536,32 +536,46 @@ fn ensure_binary_exists(
 }
 
 // Prepare the output path provided.
+
```
AlexD10S, line 539: Remove

crates/pop-cli/src/commands/build/spec.rs
```diff
@@ -536,32 +536,46 @@ fn ensure_binary_exists(
...
+	// After modifications, create the parent directory if it doesn't exist.
+	if let Some(parent_dir) = output_path.parent() {
+		if !parent_dir.exists() {
+			create_dir_all(parent_dir)?;
 		}
 	}
 	Ok(output_path)
 }
 
+
```
AlexD10S, line 577: Remove

crates/pop-cli/src/commands/build/spec.rs
```diff
@@ -769,31 +783,33 @@ mod tests {
...
-			let result = prepare_output_path(&non_existing_dir)?;
-			// Expected path: non_existing_dir/chain-spec.json
-			let expected_path = non_existing_dir.join(DEFAULT_SPEC_NAME);
+		let test_cases = [
+			("chain-spec.json", "chain-spec.json"),
+			("existing_dir", "existing_dir.json"),
+			("existing_dir/", "existing_dir/chain-spec.json"),
+			("non_existing_dir", "non_existing_dir.json"),
+			("non_existing_dir/", "non_existing_dir/chain-spec.json"),
+			("some_file	", "some_file.json"),
```
AlexD10S, line 792: This specific case fails when I run the test.
The fix is only to remove the white space: ("some_file", "some_file.json"),

crates/pop-cli/src/commands/build/spec.rs
```diff
@@ -769,31 +783,33 @@ mod tests {
...
-			assert_eq!(result, expected_path);
-		}
-
-		// Non-existing directory Path.
-		for dir in ["non_existing_dir", "non_existing_dir/", "non_existing_dir_json"] {
-			let non_existing_dir = temp_dir_path.join(dir);
-			let result = prepare_output_path(&non_existing_dir)?;
-			// Expected path: non_existing_dir/chain-spec.json
-			let expected_path = non_existing_dir.join(DEFAULT_SPEC_NAME);
+		let test_cases = [
```
AlexD10S, line 786: Might be a good idea to throw in a more complex path, like `("existing_dir/subdir/", "existing_dir/subdir/chain-spec.json")`. I tried it out and it works, but it’d add a bit more coverage.

AlexD10S (changes requested):
Nice! Thanks for improving this. The only important change is to fix the test, but is an easy fix.

crates/pop-cli/src/commands/build/spec.rs
al3mart: Sorry I might have added that by accident when changing tabs. Thanks!

crates/pop-cli/src/commands/build/spec.rs
al3mart: Included in https://github.com/r0gue-io/pop-cli/pull/357/commits/b1969c0d820fbc30a1c9ab7ceaa60126307dca53

Daanvdplas (approved):

AlexD10S (approved):
Great!

Issue #350, opened 2024-11-25
chore: support specify contract path input with or without `-p` flag


AlexD10S:
Currently, contract commands use the `-p` flag to specify the path to a contract or artifact file.  The experience can be improved if `pop-cli` supports both options: using the `-p` flag and specifying the path directly without it (subcommand). 

Example:
```
pop call contract -p ../contract.json 
```
can simply be 
```
pop call contract ../contract.json
```

This improvement should be consistent across all contract-related commands, including:
- `pop build`
- `pop up contract`
- `pop call contract`

ndkazu:
Hello, I started working on this issue and confirmed that the improvement is working as expected for `pop build`.
`pop call contract` & some additional tests are still needed: Work in progress.

ndkazu:
@AlexD10S , I made a cleaner version of my first suggestion (modifications are minimal, and limited to pop build ...). I am still looking for other (less invasive?) routes, but based on my previous comment, let me know if it makes sense to propagate this method to pop up contract & pop call contract.

ndkazu:
Can be closed

AlexD10S:
Completed in https://github.com/r0gue-io/pop-cli/pull/361

PR #361, opened 2024-12-03
chore: support specify contract path input with or without -p flag

ndkazu:
## Description
This PR is addressing  issue #350. 
Allow `pop-cli` to  supports two ways to  to specify the path to a contract or artifact file: using the `-p` flag or specifying the path directly without it:
```
pop call contract -p ../contract.json 
```
and
```
pop call contract ../contract.json
```
should both be possible.
## Implementation
Two arguments are used to specify the path: One Optional, and one positional:
```
#[arg(long,required = false)]
	pub(crate) path: Option<PathBuf>,
	#[arg(value_name = "PATH",required = false)]
	pub(crate) path1: Option<PathBuf>,
```
The following contract-related commands will be upgraded
- [x]     pop build
- [x]     pop up contract
- [x]     pop call contract


AlexD10S:
Thanks for your interest in contributing to `pop-cli`! 
I’ve reviewed your approach, and maintaining a second argument `path1` might not be the cleanest way to solve this. 
You might want to explore using positional arguments, as discussed here [clap-rs/clap#2260](https://github.com/clap-rs/clap/discussions/2260), or consider the use of subcommands (https://docs.rs/clap/latest/clap/trait.Subcommand.html) instead.

ndkazu:
> Thanks for your interest in contributing to `pop-cli`! I’ve reviewed your approach, and maintaining a second argument `path1` might not be the cleanest way to solve this. You might want to explore using positional arguments, as discussed here [clap-rs/clap#2260](https://github.com/clap-rs/clap/discussions/2260), or consider the use of subcommands (https://docs.rs/clap/latest/clap/trait.Subcommand.html) instead.

Hello, I looked into it, and it seems to me that what we're trying to do is closer to [this](https://github.com/clap-rs/clap/discussions/1818), and [this](https://github.com/clap-rs/clap/issues/1820).
and from what I understood from these discussions involving the Clap devTeam, using two arguments is the way to go(maybe not the way I did it though...).

ndkazu:
@AlexD10S , I made a cleaner version of my first suggestion (modifications are limited to `pop build ...`). I am still investigating other routes, but based on my previous remark, let me know if it makes sense to propagate this method to `pop up contract` & `pop call contract`.

AlexD10S:
> @AlexD10S , I made a cleaner version of my first suggestion (modifications are minimal, and limited to `pop build ...`). I am still looking for other (less invasive?) routes, but based on my previous comment, let me know if it makes sense to propagate this method to `pop up contract` & `pop call contract`.

Thanks! Running the CI to check all tests pass, and I'll do a proper review

ndkazu:
> > @AlexD10S , I made a cleaner version of my first suggestion (modifications are minimal, and limited to `pop build ...`). I am still looking for other (less invasive?) routes, but based on my previous comment, let me know if it makes sense to propagate this method to `pop up contract` & `pop call contract`.
> 
> Thanks! Running the CI to check all tests pass, and I'll do a proper review

I saw that the CLI was failing, and did a `cargo fmt`.

ndkazu:
Any update? should I propagate the modification to `pop up contract` & `pop call contract` ?

evilrobot-01:
> Any update? should I propagate the modification to `pop up contract` & `pop call contract` ?

Sorry for the delay, we have been readying a release. @AlexD10S could you provide some feedback on next steps here please?

AlexD10S (changes requested):
Thank you for sharing these resources! They were very helpful in understanding the best approach. While I'm not thrilled about it, it seems that keeping the two arguments is the most practical solution, great job.

A couple of quick changes:

If you run `pop` in your parachain or contract folder, there’s no need to specify the path. That's why we changed the argument to Option.
You could remove `required_unless_present` and probably makes sense to use `index = 1`.

It might also be helpful to add `conflicts_with = "path"` so that users don’t accidentally specify both positional and named paths, like this:
```
pop build ../my_contract --path ../my_contract_2
```

Here's an example of how the argument definition could look:
```
/// Directory path without flag for your project [default: current directory]
#[arg(value_name = "PATH", index = 1, conflicts_with = "path")]
pub(crate) path_pos: Option<PathBuf>,
```

Finally, with this optional path in mind, the code where you determine project_path can be simplified to something like this:

```
 let project_path = if let Some(path_pos) = args.path_pos {
        Some(path_pos) // Use positional path if present
    } else {
        args.path // Otherwise, use the named path
    };
```

Great work so far, and thank you for your patience while I got around to reviewing this! Apologies for the delay.

crates/pop-cli/src/commands/call/contract.rs
```diff
@@ -148,11 +158,21 @@ impl CallContractCommand {
 
 	/// If the contract has not been built, build it in release mode.
 	async fn ensure_contract_built(&self, cli: &mut impl Cli) -> Result<()> {
+		let path_flag = self.path.clone();
```
AlexD10S, line 161: This code is now being used in four different place, it makes sense to refactor this and create a method within the `pop-cli/common` module. Also will make it easier to test only this funcionality.

crates/pop-cli/src/commands/call/contract.rs
```diff
@@ -109,7 +112,14 @@ impl CallContractCommand {
 
 	fn display(&self) -> String {
 		let mut full_message = "pop call contract".to_string();
-		if let Some(path) = &self.path {
+		let path_flag = self.path.clone();
```
AlexD10S, line 115: This function is only to print the command in the screen to the user.
Is probably easier only:
```rs
if let Some(path) = &self.path {
  full_message.push_str(&format!(" --path {}", path.display()));
}
if let Some(path_pos) = &self.path_pos {
  full_message.push_str(&format!("{}", path_pos.display()));
}
```

crates/pop-cli/src/commands/call/contract.rs
```diff
@@ -148,11 +158,21 @@ impl CallContractCommand {
...
+		} else {
+			path_flag.as_ref() // Otherwise, use the named path
+		};
 		// Build the contract in release mode
 		cli.warning("NOTE: contract has not yet been built.")?;
 		let spinner = spinner();
 		spinner.start("Building contract in RELEASE mode...");
-		let result = match build_smart_contract(self.path.as_deref(), true, Verbosity::Quiet) {
+		let result = match build_smart_contract(
+			project_path.as_deref().map(|v| &**v),
```
AlexD10S, line 172: This feels unnecesary, maybe you can change the code above to get an Option<Path>:
```rs
let path_flag = args.path.clone();
let project_path = if let Some(ref path_pos) = args.path_pos {
	Some(path_pos.as_path()) // Use positional path if present
} else {
	path_flag.as_deref() // Otherwise, use the named path
};

#[cfg(feature = "contract")]
if pop_contracts::is_supported(project_path)? 
```

crates/pop-cli/src/commands/call/contract.rs
```diff
@@ -564,6 +605,7 @@ mod tests {
 		// Contract deployed on Pop Network testnet, test get
 		CallContractCommand {
 			path: Some(temp_dir.path().join("testing")),
+			path_pos: Some(temp_dir.path().join("testing")),
```
AlexD10S, line 608: Should not be None? We already specified `path`, no need to specify `path_pos`.  This is probably what is making the unit tests fail, you can test it locally with: `cargo test --lib --bins`.

crates/pop-cli/src/commands/up/contract.rs
```diff
@@ -92,13 +95,23 @@ impl UpContractCommand {
...
 
+		let path_flag = self.path.clone();
+		let project_path = if let Some(ref path_pos) = self.path_pos {
+			Some(path_pos) // Use positional path if present
+		} else {
+			path_flag.as_ref() // Otherwise, use the named path
+		};
 		// Check if build exists in the specified "Contract build directory"
-		if !has_contract_been_built(self.path.as_deref()) {
+		if !has_contract_been_built(project_path.as_deref().map(|v| &**v)) {
```
AlexD10S, line 105: Same as a comment above.

crates/pop-cli/tests/contract.rs
```diff
@@ -67,13 +67,24 @@ async fn contract_lifecycle() -> Result<()> {
 		.success();
 	assert!(temp_dir.join("test_contract").exists());
 
-	// pop build --path ./test_contract --release
+	/*	// pop build --path ./test_contract --release
```
AlexD10S, line 70: I'd keep this code and remove the one below, test it with the `--path` flag in this specific case

crates/pop-cli/tests/contract.rs
```diff
@@ -67,13 +67,24 @@ async fn contract_lifecycle() -> Result<()> {
...
+	// pop build ./test_contract --release
 	Command::cargo_bin("pop")
 		.unwrap()
 		.current_dir(&temp_dir)
-		.args(&["build", "--path", "./test_contract", "--release"])
+		.args(&["build", "./test_contract", "--release"])
 		.assert()
 		.success();
+
+	println!("Contract built!!");
```
AlexD10S, line 86: Remove

crates/pop-cli/tests/contract.rs
```diff
@@ -88,13 +99,28 @@ async fn contract_lifecycle() -> Result<()> {
...
+			"contract",
+			"--path",
+			"./test_contract",
+			"--upload-only",
+			"--url",
+			default_endpoint,
+		])
 		.assert()
 		.success();
+	/*// pop up contract --upload-only
```
AlexD10S, line 117: Remove commented code

AlexD10S (changes requested):
Good work!
I left some comment mainly to refactor repetitive code.
The important thing to fix are the unit tests that are failing. I think it can be because you specify both `path` and `path_pos` in the tests and then the result is not the expected.
Review the tests and adjust it. I'd probably leave `path` and `path_pos` to None in most of the tests  except one where use `path_pos` or  have a specific one to test the `path_pos` functionality

You can test them locally with: `cargo test --lib --bins`.

AlexD10S (changes requested):
Thanks for the fixes! I noticed the CI is failing because the get_project_path function, used for both contracts and parachain, was moved to a file that only works under the contracts feature, see: https://github.com/r0gue-io/pop-cli/blob/main/crates/pop-cli/src/common/mod.rs#L3

you can test it with:  `cargo check --no-default-features --features parachain`

You can fix this by either removing the line above or, better yet, moving the function to a shared file under common (maybe a new build.rs). 
Also, please add comments to the method.

ndkazu:
@AlexD10S , I see that all CI checks are passing: what's next?

evilrobot-01:
Alex is OOO today, so expect a response when he is back tomorrow.

AlexD10S (approved):
Great job! Everything looks excellent. Thank you for your patience and effort in addressing all the requested changes.

PR #362, opened 2024-12-04
feat(wallet-integration): server and API

peterwht:
### Overview
Implements a server and API in Pop CLI that enables wallet integrations for secure signing:
1. Pop CLI -> launch UI
2. UI performs GET from API -> Pop CLI returns tx data from memory
3. UI opens wallet, wallet signs TX [frontend not created yet]
4. UI performs POST to API with signed payload

`WalletIntegrationManager` is used to launch the frontend and API. It accepts encoded call data to pass to the frontend.

**All tests are not implemented.** I will wait for high level review before finishing tests.

### API endpoints:
- `/` -> access frontend
- GET `/data` for getting the data needed for signing
- POST `/submit` handles the submission of the signed payload. Stored for later use in Pop CLI for on-chain submission.
- POST `/error` to report an error message to Pop CLI
- POST `/terminate` to terminate the server

Two new dependencies were added for the server:
- axum, built by the tokio-rs organization and is used for creating the web-API. Uses tokio. Relatively lightweight.
- tower_http, recommended for use with axum. Tower serves the static frontend.

```
# GET
curl -X GET http://127.0.0.1:3000/payload

# POST
curl -X POST http://127.0.0.1:3000/submit \
             -H "Content-Type: application/json" \
             -d '"0x1234"'
```


peterwht:
pop cli wallet integration server and launching [sc-1717]

crates/pop-wallet-integration/src/lib.rs
```diff
@@ -0,0 +1,161 @@
...
+		// used to signal shutdown.
+		let (tx, rx) = oneshot::channel();
+
+		// shared state between routes. Will be used to store the signed payload.
+		let state =
+			Arc::new(Mutex::new(StateHandler { shutdown_tx: Some(tx), signed_payload: None }));
+
+		// will shutdown when the signed payload is received
+		let app = Router::new()
+			// cloning Arcs is cheap
```
al3mart, line 144: on line [76](https://github.com/r0gue-io/pop-cli/pull/362/commits/173ec1b707f2fb764921b088fb5502e6cc6919c8#diff-15f82683f5b143fff44ed297f03becc696d7983f7abe1727f94382979a2fff70R76) you mention that cloning the field data can be expensive while being an `Arc` as well.

The `Data` struct should never be cloned in any case, right ?

crates/pop-wallet-integration/src/lib.rs
```diff
@@ -0,0 +1,209 @@
...
+		args: Vec<String>,
+		value: String,
+		gas_limit: Option<u64>,
+		proof_size: Option<u64>,
+		salt: Option<Vec<u8>>,
+	}
+
+	/// Arguments for calling a contract.
+	#[derive(Serialize, Debug, PartialEq)]
+	pub struct CallArgs {
```
AlexD10S, line 57: For the data we pass to contracts, can we construct the `call_data` in `pop-cli` in the same way as we do for parachain extrinsics. This is a feature we didn’t have before, but for `pop call parachain` we are constructing the extrinsic (see https://github.com/r0gue-io/pop-cli/blob/feat-call-ui-extrinsics-limited/crates/pop-parachains/src/call/mod.rs#L30). Before open the UI we have all the fields required to construct the Contracts(or Revive)::call  extrinsic so we can get the `call_data` and have the same behaviour as Parachain extrinsics?

With this approach, we will still need to modify how the `dry_run` is performed in the `pop-cli`

crates/pop-wallet-integration/src/lib.rs
```diff
@@ -0,0 +1,161 @@
...
+	chain_rpc: String,
+	data_type: DataType,
+}
+
+/// The type of transaction with specific data for signing (contract or parachain).
+#[derive(Serialize)]
+pub enum DataType {
+	/// Parachain call, where Vec<u8> is the encoded call data.
+	Parachain(Vec<u8>),
+	Contract(ContractArgs),
```
al3mart, line 22: We should be able to benefit from the work done around `pop call` and `pop up` to intake the payloads

crates/pop-wallet-integration/src/lib.rs
```diff
@@ -0,0 +1,209 @@
...
+	frontend_path: PathBuf,
+	// cloning can be expensive (e.g. contract code stored in-memory)
+	data: Arc<Data>,
+	signed_payload: Option<String>,
+}
+
+impl WalletIntegrationManager {
+	/// - frontend_path: Path to the wallet-integration frontend.
+	/// - data: Data to be sent to the frontend for signing.
+	pub fn new(frontend_path: PathBuf, data: Data) -> Self {
```
evilrobot-01, line 128: Why a path? My assumption was that the UI would be bundled with the binary, whereas this implies it has to be written to some local path first. Could it not just serve the static content from a string, or is the problem the additional assets (images/css)? I would have assumed those would be better server from a CDN to keep the binary size down.

crates/pop-wallet-integration/src/lib.rs
```diff
@@ -0,0 +1,209 @@
...
+			Arc::new(Mutex::new(StateHandler { shutdown_tx: Some(tx), signed_payload: None }));
+
+		// will shutdown when the signed payload is received
+		let app = Router::new()
+			// cloning Arcs is cheap
+			.route("/data", get(routes::get_data_handler).with_state(self.data.clone()))
+			.route("/submit", post(routes::handle_submit).with_state(state.clone()))
+			.nest_service("/", ServeDir::new(self.frontend_path.clone()));
+
+		let listener = tokio::net::TcpListener::bind("127.0.0.1:3000").await.unwrap();
```
evilrobot-01, line 149: Consider port is in use and an unwrap reminder.

crates/pop-wallet-integration/src/lib.rs
```diff
@@ -0,0 +1,209 @@
...
+
+		// shared state between routes. Will be used to store the signed payload.
+		let state =
+			Arc::new(Mutex::new(StateHandler { shutdown_tx: Some(tx), signed_payload: None }));
+
+		// will shutdown when the signed payload is received
+		let app = Router::new()
+			// cloning Arcs is cheap
+			.route("/data", get(routes::get_data_handler).with_state(self.data.clone()))
+			.route("/submit", post(routes::handle_submit).with_state(state.clone()))
```
evilrobot-01, line 146: What about routes for error reporting, cancellations from the wallet?

crates/pop-wallet-integration/src/lib.rs
```diff
@@ -0,0 +1,209 @@
...
+				}}})
+			);
+		}
+
+		// TODO: after high level review, complete serialization tests.
+	}
+}
+use crate::contracts::ContractArgs;
+
+struct StateHandler {
```
evilrobot-01, line 112: Comments would be appreciated to better understand the usages. My understanding is that a message is sent to the channel to shutdown the server once successfully signed.

evilrobot-01 (commented):
Looks good, mostly just a few comments to confirm my understanding.

Might be useful to confirm there are no obvious hurdles as the data size is increased as early as possible - e.g. can a runtime upgrade be instantiated, the wasm included in the data passed to the server on startup and then served to the wallet via the route. I would imagine it would be fine, but might be worth checking sooner than later.

PS - whilst routes are easier to work with, might there be any benefits in using websockets instead of http?

crates/pop-wallet-integration/src/lib.rs
peterwht: Recommendation is to source the frontend from a GH release, similar to what we do with nodes. So, the plan is to have a path to the pop cache for the frontend.

crates/pop-wallet-integration/src/lib.rs
peterwht: good point. Maybe can consider doing something similar to zombienet.

crates/pop-wallet-integration/src/lib.rs
peterwht: however, will attempt to get everything into one string still -- a good idea

evilrobot-01: It would be nice to ship with something that works out of the box, even if just a fallback. The GH release does provide a means to push updates if the UI needs bumps (e.g. security patches), but it does arguably add complexity in caching and checking for new versions.

crates/pop-wallet-integration/src/lib.rs
evilrobot-01: But if its being submitted by the CLI via subxt, then as long as its signed does it matter? Guess it depends on whether is the UI showing inclusion/finalization or the CLI. Or both?!

peterwht: Preference is Pop CLI only.

crates/pop-wallet-integration/src/lib.rs
```diff
@@ -0,0 +1,209 @@
...
+/// The type of transaction with specific data for signing (contract or parachain).
+#[derive(Serialize, Debug, PartialEq)]
+pub enum DataType {
+	/// Parachain call, where Vec<u8> is the encoded call data.
+	Parachain(Vec<u8>),
+	Contract(ContractArgs),
+}
+
+/// Contract specific data variations.
+pub mod contracts {
```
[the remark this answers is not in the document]
peterwht: removed -- PAPI can decode the call data for us allowing dry-runs to still work even if we only pass encoded call data.

crates/pop-wallet-integration/src/lib.rs
peterwht: https://github.com/r0gue-io/pop-cli/pull/362/commits/cfe41f00110093a32bd2006efafad443b11d0085

crates/pop-wallet-integration/src/lib.rs
```diff
@@ -0,0 +1,161 @@
...
+		let (tx, rx) = oneshot::channel();
+
+		// shared state between routes. Will be used to store the signed payload.
+		let state =
+			Arc::new(Mutex::new(StateHandler { shutdown_tx: Some(tx), signed_payload: None }));
+
+		// will shutdown when the signed payload is received
+		let app = Router::new()
+			// cloning Arcs is cheap
+			.route("/data", get(routes::get_data_handler).with_state(self.data.clone()))
```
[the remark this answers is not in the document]
peterwht: https://github.com/r0gue-io/pop-cli/pull/362/commits/cfe41f00110093a32bd2006efafad443b11d0085

crates/pop-wallet-integration/src/lib.rs
peterwht: https://github.com/r0gue-io/pop-cli/pull/362/commits/cfe41f00110093a32bd2006efafad443b11d0085

crates/pop-wallet-integration/Cargo.toml
```diff
@@ -0,0 +1,14 @@
+[package]
+name = "pop-wallet-integration"
```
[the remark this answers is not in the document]
peterwht: To be honest, not much additional thought was given besides following existing architecture. I am sure there are things that this could be extended into in terms of (potentially) hosting the backend as an API, extending to other types of signing, etc.

But, the maintenance burden is valid. I will move to a sub-module within the pop-cli crate. 

Feel free to mark as resolved if satisfied. cc @AlexD10S

crates/pop-wallet-integration/src/lib.rs
```diff
@@ -0,0 +1,209 @@
...
+/// Data to be sent to frontend for signing.
+#[derive(Serialize, Debug)]
+pub struct Data {
+	chain_rpc: String,
+	data_type: DataType,
+}
+
+/// The type of transaction with specific data for signing (contract or parachain).
+#[derive(Serialize, Debug, PartialEq)]
+pub enum DataType {
```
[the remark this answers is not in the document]
peterwht: https://github.com/r0gue-io/pop-cli/pull/362/commits/cfe41f00110093a32bd2006efafad443b11d0085 -> `TransactionData`

crates/pop-wallet-integration/src/lib.rs
peterwht: Added a simple frontend trait that serves a custom route. Allows us to easily plug in different frontend sources (baked-in string, cache, etc.).

Please see: https://github.com/r0gue-io/pop-cli/pull/362/commits/0edfbad00685263f61e690b1bc0e21434352fc83

crates/pop-wallet-integration/src/lib.rs
peterwht: Thinking out loud. What types of errors might there be in the frontend?
- Wallet cancellation -> problem for the frontend.
- Invalid call data -> theoretically should not happen if pop-cli validates. However, still worthwhile to report to CLI.
- Unable to connect to RPC -> report to CLI
- Dry-run failure

peterwht:
> PS - whilst routes are easier to work with, might there be any benefits in using websockets instead of http?

For the current CLI <> frontend communication, I feel a websocket will be overkill and have too much complexity. Theoretically, in this first version we should have the following communications:
- (serve files)
- Frontend: GET payload FROM server
- Report error TO server
- POST signed payload TO server

This is rather infrequent. I would consider a websocket for more advanced versions where the server is needing to report continuous information (events, block inclusion, etc.) to the frontend. First version frontend has very clear expectations in when it needs to call the API (action-based, no polling needed).

crates/pop-wallet-integration/src/lib.rs
peterwht: no longer needed

crates/pop-wallet-integration/Cargo.toml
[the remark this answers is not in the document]
peterwht: Please see https://github.com/r0gue-io/pop-cli/pull/362/commits/8cffac1ebd896e63db3a5eb08d399505001d5939. Resolve if happy.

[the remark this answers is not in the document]
AlexD10S: Joining the discussion a bit late here, but I like the last change you made. 
If we believe other projects can use this functionality, we can place it in a separate crate, but for now is fine in pop-cli crate.

crates/pop-wallet-integration/src/lib.rs
```diff
@@ -0,0 +1,209 @@
...
+	/// Parachain call, where Vec<u8> is the encoded call data.
+	Parachain(Vec<u8>),
```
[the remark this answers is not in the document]
peterwht: see https://github.com/r0gue-io/pop-cli/pull/362/commits/81d5068df728d529e8b682b5c49ad7c8caf855e8#diff-437a9f52ab531ef3be20b5017014887e96f67be0d5a7963e98ef0f9b3e548b2fR369

crates/pop-wallet-integration/src/lib.rs
peterwht: https://github.com/r0gue-io/pop-cli/pull/362/commits/54837ae29c45a89e28247c0b0bf30b2efa8c1a3d

crates/pop-cli/src/wallet_integration.rs
```diff
@@ -0,0 +1,510 @@
...
+		let addr = format!("http://{}", wim.addr);
+		let actual_payload = reqwest::get(&format!("{}/payload", addr))
+			.await
+			.expect("Failed to get payload")
+			.json::<TransactionData>()
+			.await
+			.expect("Failed to parse payload");
+
+		assert_eq!(actual_payload.chain_rpc, "localhost:9944");
+		assert_eq!(actual_payload.call_data, vec![1, 2, 3]);
```
Daanvdplas, line 278:
```suggestion
		assert_eq!(actual_payload.chain_rpc, expected_payload.chain_rpc);
		assert_eq!(actual_payload.call_data, expected_payload.call_data);
```

Daanvdplas (commented):
Had a brief look at the code to get myself familiar with the workings for the integration with the parachain onboarding. Leaving a few nitpicks that I came across. 

Great job btw, must have been a cool feature to work on!

al3mart (approved):
+1 on the requested doc styling. Other than that, I'm happy with the status of the PR.

crates/pop-cli/src/wallet_integration.rs
```diff
@@ -0,0 +1,509 @@
...
+			Ok(())
+		});
+
+		Self { state, rpc_url: rpc.to_string(), task_handle }
+	}
+
+	/// Signals the wallet integration server to shut down.
+	pub async fn terminate(&mut self) {
+		if let Some(shutdown_tx) = self.state.lock().await.shutdown_tx.take() {
+			let _ = shutdown_tx.send(());
```
AlexD10S, line 115:
```suggestion
			shutdown_tx.send(())?;
```

crates/pop-cli/src/wallet_integration.rs
```diff
@@ -0,0 +1,509 @@
...
+		State(state): State<Arc<Mutex<StateHandler>>>,
+		Json(payload): Json<String>,
+	) -> Json<serde_json::Value> {
+		let mut state = state.lock().await;
+		state.signed_payload = Some(payload);
+
+		// Signal shutdown.
+		// Using WalletIntegrationManager::terminate() introduces unnecessary complexity.
+		if let Some(shutdown_tx) = state.shutdown_tx.take() {
+			let _ = shutdown_tx.send(());
```
AlexD10S, line 181:
```suggestion
			shutdown_tx.send(())?;
```

crates/pop-cli/src/wallet_integration.rs
```diff
@@ -0,0 +1,509 @@
...
+	) {
+		let mut state = state.lock().await;
+		state.error = Some(error);
+	}
+
+	/// Allows the server to be terminated from the frontend.
+	pub(super) async fn terminate_handler(State(state): State<Arc<Mutex<StateHandler>>>) {
+		let mut state = state.lock().await;
+		if let Some(shutdown_tx) = state.shutdown_tx.take() {
+			let _ = shutdown_tx.send(());
```
AlexD10S, line 201:
```suggestion
			shutdown_tx.send(())?;
```

AlexD10S (approved):
Looks good! 
Just a couple of small nitpicks. Also run `cargo +nightly clippy`, I know most of the warnings are temporal until we connect the rest of functionality, but some I believe can cleaned.

PR #316, opened 2024-09-24
feat: guide user to call a parachain

AlexD10S:
This PR introduces the `pop call parachain` command, providing guidance for users to interact with a chain. The approach to achieving a great DevEx is to start with a list of specific use cases that we believe will be most useful for developers,  such as creating an asset, nfts, and other interesting developer focused actions, while also allowing them to access a broader range of extrinsics.

```
┌   Pop CLI : Call a parachain
│
◇    Which chain would you like to interact with?
│  ws://localhost:9944/
│
◆  What would you like to do?
│  ○ Transfer balance 
│  ○ Create an asset 
│  ○ Mint an asset 
│  ○ Create an NFT collection 
│  ○ Mint an NFT 
│  ● All (Explore all pallets and extrinsics)
└  
```

`pop-cli` also offer a the command-line interface for direct execution. For example to mint an NFT directly: `pop call parachain --pallet Nfts --extrinsic mint --args "1" "1" "Id(5FHneW46xGXgs5mUiveU4sbTyGBzmstUspZC92UhjJM694ty)" "None()" --url ws://localhost:9944/ --suri //Alice`

####  Additional PRs:
To simplify the review process, some functionality has been split out from this main PR:
- [x] Submit extrinsic indicating the `call_data`: https://github.com/r0gue-io/pop-cli/pull/348
- [x] Wrap the selected call as a `sudo` call: https://github.com/r0gue-io/pop-cli/pull/349
- [x]  Parse files when the argument values are very big https://github.com/r0gue-io/pop-cli/pull/363 
There is a limitation on the input size in the CLI, which may cause the process to fail if the input string is too large. For instance, this issue can occur when using the `register` extrinsic:
```
◇  What would you like to do?
│  Register a parachain ID with genesis state and code 
│
◇  Enter the value for the parameter: id
│  2000
│
◇  Enter the value for the parameter: genesis_head
│  0x000000000000000000000000000000000000000000000000000000000000000000fc3a2a9c95d7e6cfc871d72111eb7af5c30da67f1b8ef35ff4eb346c9483201e03170a2e7597b7b7e3d84c05391d139a62b157e78786d8c082f29dcf4c11131400
│
◆  Enter the value for the parameter: validation_code
│  Type required: [u8]
└  
```

####  Review Considerations

- Some extrinsics are not supported yet due to the complexity of input the parameters. Example: A `sudo` call where another call has to be inputted by a command line: 
```
- ◇  Select the pallet to call:
│  Sudo 
│
[...]

PR #363, opened 2024-12-05
feat: parse files when the argument values are very big

AlexD10S:
This PR is part of the `pop call parachain `feature. 
It improves handling the parameters with type `TypeDef::Sequence`  similar to the approach used in PolkadotJS.

1. Properly parsing user-provided values automatically encoding the value into a hexadecimal format. (Based on this comment https://github.com/r0gue-io/pop-cli/pull/316#discussion_r1874191647)
2.  Allowing users to provide large values through a file, similar to the approach used in PolkadotJS, instead of requiring direct string input:

Captura de pantalla 2024-12-05 a las 20 59 12

**For testing**
Run `pop call parachain`, input a relay chain url and select the `Register a parachain ID with genesis state and code` option.
```
◇  What would you like to do?
│  Register a parachain ID with genesis state and code 
│
◇  Enter the value for the parameter: id
│  2000
│
◇  The value for `genesis_head` might be too large to enter. Provide a file instead?
│  Yes 
│
◇  Enter the file path for the parameter `genesis_head`:
│  ../pop-node/para-2000-genesis-state
│
◇  The value for `validation_code` might be too large to enter. Provide a file instead?
│  Yes 
│
◇  Enter the file path for the parameter `validation_code`:
│  ../pop-node/para-2000.wasm
```

[sc-1507]

### DIFF
--- a/crates/pop-cli/src/commands/call/parachain.rs
+++ b/crates/pop-cli/src/commands/call/parachain.rs
@@ -1,5 +1,7 @@
 // SPDX-License-Identifier: GPL-3.0
 
+use std::path::Path;
+
 use crate::cli::{self, traits::*};
 use anyhow::{anyhow, Result};
 use clap::Args;
@@ -75,8 +77,9 @@ impl CallParachainCommand {
 				break;
 			}
 
-			if !prompt_to_repeat_call ||
-				!cli.confirm("Do you want to perform another call?")
+			if !prompt_to_repeat_call
+				|| !cli
+					.confirm("Do you want to perform another call?")
 					.initial_value(false)
 					.interact()?
 			{
@@ -139,8 +142,9 @@ impl CallParachainCommand {
 
 			// Resolve extrinsic.
 			let extrinsic = match self.extrinsic {
-				Some(ref extrinsic_name) =>
-					find_extrinsic_by_name(&chain.pallets, &pallet.name, extrinsic_name).await?,
+				Some(ref extrinsic_name) => {
+					find_extrinsic_by_name(&chain.pallets, &pallet.name, extrinsic_name).await?
+				},
 				None => {
 					let mut prompt_extrinsic = cli.select("Select the extrinsic to call:");
 					for extrinsic in &pallet.extrinsics {
@@ -177,8 +181,9 @@ impl CallParachainCommand {
 			// Resolve who is signing the extrinsic.
 			let suri = match self.suri.as_ref() {
 				Some(suri) => suri.clone(),
-				None =>
-					cli.input("Signer of the extrinsic:").default_input(DEFAULT_URI).interact()?,
+				None => {
+					cli.input("Signer of the extrinsic:").default_input(DEFAULT_URI).interact()?
+				},
 			};
 
 			return Ok(CallParachain {
@@ -200,11 +205,11 @@ impl CallParachainCommand {
 
 	// Function to check if all required fields are specified.
 	fn requires_user_input(&self) -> bool {
-		self.pallet.is_none() ||
-			self.extrinsic.is_none() ||
-			self.args.is_empty() ||
-			self.url.is_none() ||
-			self.suri.is_none()
+		self.pallet.is_none()
+			|| self.extrinsic.is_none()
+			|| self.args.is_empty()
+			|| self.url.is_none()
+			|| self.suri.is_none()
 	}
 
 	/// Replaces file arguments with their contents, leaving other arguments unchanged.
@@ -287,8 +292,9 @@ impl CallParachain {
 		tx: DynamicPayload,
 		cli: &mut impl Cli,
 	) -> Result<()> {
-		if !self.skip_confirm &&
-			!cli.confirm("Do you want to submit the extrinsic?")
+		if !self.skip_confirm
+			&& !cli
+				.confirm("Do you want to submit the extrinsic?")
 				.initial_value(true)
 				.interact()?
 		{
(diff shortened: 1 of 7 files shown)


crates/pop-cli/src/commands/call/parachain.rs
```diff
@@ -362,6 +380,30 @@ fn get_param_value(
 	}
 }
 
+// Prompt for the value when is a primitive.
+fn prompt_for_sequence_param(cli: &mut impl Cli, param: &Param) -> Result<String> {
+	if cli
+		.confirm(format!(
+			"The value for `{}` might be too large to enter. Provide a file instead?",
```
al3mart, line 387: Let's maybe ask directly for a file instead of asking if the user wants to paste in the terminal. I think both of these, the code and the state, will be always better managed in files.

AlexD10S: My approach was similar to PolkadotJS, using a simple input example like the `remark` extrinsic where you probably won't need a File. However, your suggestion likely makes more sense, as it helps avoid potential input issues for users.

crates/pop-parachains/src/call/mod.rs
```diff
@@ -28,11 +28,11 @@ pub async fn set_up_client(url: &str) -> Result<OnlineClient<SubstrateConfig>, E
 /// * `args` - A vector of string arguments to be passed to the extrinsic.
 pub async fn construct_extrinsic(
 	pallet_name: &str,
-	extrinsic_name: &str,
+	extrinsic: &Extrinsic,
 	args: Vec<String>,
 ) -> Result<DynamicPayload, Error> {
-	let parsed_args: Vec<Value> = metadata::parse_extrinsic_arguments(args).await?;
-	Ok(subxt::dynamic::tx(pallet_name, extrinsic_name, parsed_args))
+	let parsed_args: Vec<Value> = metadata::parse_extrinsic_arguments(&extrinsic, args).await?;
```
evilrobot-01, line 34: Clippy warning - please kindly self-review on GH, after CI successful, before requesting reviews.

crates/pop-parachains/src/call/mod.rs
```diff
@@ -92,11 +92,16 @@ mod tests {
 
 	#[tokio::test]
 	async fn construct_extrinsic_works() -> Result<()> {
+		let client = set_up_client("wss://rpc1.paseo.popnetwork.xyz").await?;
+		let pallets = parse_chain_metadata(&client).await?;
```
evilrobot-01, line 96: For future - it would probably be better to place this behind a lazy static or similar so that it is done once per test **run** rather than for every test, reducing the number of metadata parse requests.

evilrobot-01 (changes requested):
Looks good, some suggestions for relatively easy improvements.

crates/pop-cli/src/commands/call/parachain.rs
AlexD10S: Based on both opinions and to minimize unnecessary prompts, I removed the confirmation step. Now, the user input can be either a file path or a value, and `pop-cli` will handle the value:
```rs
cli
        .input(format!(
            "The value for `{}` might be too large to enter. You may enter the path to a file instead.",
            param_name
        ))
        .placeholder(&format!(
            "Enter a value of type {} or provide a file path (e.g., /path/to/your/file.json)",
            param_type
        ))
        .interact()?;
```
Refactored in this PR: https://github.com/r0gue-io/pop-cli/pull/363/commits/f39900e7f0791e61b0ad289333e01c54abe48bf2

crates/pop-parachains/src/call/metadata/mod.rs
```diff
@@ -228,6 +241,24 @@ mod tests {
 
 	#[tokio::test]
 	async fn parse_extrinsic_arguments_works() -> Result<()> {
+		/// Helper function to create a `Param` with default values.
+		fn create_param(
```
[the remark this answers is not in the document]
AlexD10S: Removed. Not needed anymore with the use of `Default`

crates/pop-parachains/src/call/metadata/params.rs
```diff
@@ -20,6 +20,8 @@ pub struct Param {
 	pub is_tuple: bool,
 	/// Indicates if the parameter is a Variant.
 	pub is_variant: bool,
+	/// Indicates if the parameter is a Sequence.
+	pub is_sequence: bool,
```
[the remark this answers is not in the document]
AlexD10S: Introduced in https://github.com/r0gue-io/pop-cli/pull/363/commits/f2a0f76f1a4847b5eb294efc2c82c296e05eb697. because it solves the comment to improve tests: https://github.com/r0gue-io/pop-cli/pull/363#discussion_r1875829087 

But I am going to prepare another PR to refactor all the rest of usages.

crates/pop-parachains/src/call/mod.rs
```diff
@@ -92,11 +92,16 @@ mod tests {
 
 	#[tokio::test]
 	async fn construct_extrinsic_works() -> Result<()> {
+		let client = set_up_client("wss://rpc1.paseo.popnetwork.xyz").await?;
```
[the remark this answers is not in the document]
AlexD10S: Noted.

crates/pop-parachains/src/call/mod.rs
AlexD10S: This is great feedback for improving the tests. Noting it down for a future task.

AlexD10S:
The last commit https://github.com/r0gue-io/pop-cli/pull/363/commits/423eb3b26e659aa6cc8da622aa677bdc29ad03c8 updates a library to address a vulnerability detected by the CI during the `cargo-deny check` run:  https://github.com/r0gue-io/pop-cli/actions/runs/12236673594/job/34130654485?pr=363

evilrobot-01 (approved):
Looks good.

Note that I havent tested manually, which I wanted to do on the mean feat PR instead.

PR #365, opened 2024-12-09
feat(wallet integration): pop up contract

peterwht:
Integrates the wallet integration into pop up contract. Is based on top of https://github.com/r0gue-io/pop-cli/pull/362

A new flag was added `--use-wallet`, that activates the wallet integration. The flow is (for --use-wallet):
1.  generate the call data, 
2. launch the frontend for signing that data. 
3. Once the data is received back to Pop CLI, the wallet integration will close.
4. Pop CLI will take the call data and use Subxt to submit it

## Things to Note (as of december 8th):
- unit tests still need completion
- We are using functionality in cargo-contract that had to be modified to generate the call-data. We may want to reimplement this locally.
- cargo contract changes here: https://github.com/use-ink/cargo-contract/tree/peter/chore-make-types-pub
- need to check subxt return display


crates/pop-contracts/src/up.rs
```diff
@@ -93,9 +102,95 @@ pub async fn set_up_upload(
...
+	let signer = create_signer("//Alice")?;
+	let extrinsic_opts =
+		ExtrinsicOptsBuilder::<DefaultConfig, DefaultEnvironment, Keypair>::new(signer)
+			.manifest_path(Some(manifest_path))
+			.done();
+	let artifacts = extrinsic_opts.contract_artifacts()?;
+
+	let artifacts_path = artifacts.artifact_path().to_path_buf();
+	let code = artifacts.code.ok_or_else(|| {
+		anyhow::anyhow!("Contract code not found from artifact file {}", artifacts_path.display())
```
AlexD10S, line 176: In the crates return a custom Error instead of `anyhow`

crates/pop-contracts/src/up.rs
```diff
@@ -93,9 +102,95 @@ pub async fn set_up_upload(
...
+		contract_extrinsics::upload::Determinism::Enforced,
+	);
+
+	// TODO: review placement
+	let rpc_client = subxt::backend::rpc::RpcClient::from_url(up_opts.url.as_str()).await?;
+	let client = subxt::OnlineClient::<SubstrateConfig>::from_rpc_client(rpc_client).await?;
+
+	let call_data = upload_code.build();
+	let mut encoded_data = Vec::<u8>::new();
+	call_data.encode_call_data_to(&client.metadata(), &mut encoded_data)?;
```
AlexD10S, line 125: I'll take a look on replicate this behaviour in `pop-cli` without need of `contract-extrinsics = { git = "https://github.com/use-ink/cargo-contract", branch = "peter/chore-make-types-pub" }`

crates/pop-cli/src/commands/up/contract.rs
```diff
@@ -312,6 +341,54 @@ impl UpContractCommand {
...
+	async fn wait_for_signature(&self, call_data: Vec<u8>) -> anyhow::Result<Option<String>> {
+		// TODO: to be addressed in future PR. Should not use FromDir (or local path).
+		let ui = FrontendFromDir::new(PathBuf::from(
+			"/Users/peter/dev/r0gue/react-teleport-example/dist",
+		));
+
+		let transaction_data = TransactionData::new(self.url.to_string(), call_data);
+		// starts server
+		let mut wallet = WalletIntegrationManager::new(ui, transaction_data);
+		log::step(format!("Wallet signing portal started at http://{}", wallet.addr))?;
```
AlexD10S, line 372: I get an unknown field error  here:
`log::step(format!("Wallet signing portal started at http://{}", wallet.addr))?; `

AlexD10S (commented):
Just a quick high level review, will do a more complete one tomorrow!

crates/pop-contracts/src/up.rs
peterwht: thanks for the pointer!

crates/pop-contracts/src/up.rs
```diff
@@ -93,9 +102,150 @@ pub async fn set_up_upload(
...
+	let instantiated = events
+		.find_first::<ContractInstantiated<subxt::config::substrate::AccountId32>>()?
+		.ok_or_else(|| {
+			Error::InstantiateContractError("Failed to find Instantiated event".to_string())
+		})?;
+
+	Ok(InstantiateExecResult { events, code_hash, contract_address: instantiated.contract })
+}
+
+pub async fn submit_signed_payload(
```
AlexD10S, line 214: There’s potential duplicate code here, but we can refactor it once everything is finalized. Moving this into `pop_common`: https://github.com/r0gue-io/pop-cli/blob/main/crates/pop-parachains/src/call/mod.rs#L51, would allow us to reuse it. The difference is returning events, which is a work in progress here: https://github.com/r0gue-io/pop-cli/pull/372

crates/pop-contracts/src/up.rs
```diff
@@ -93,9 +102,150 @@ pub async fn set_up_upload(
...
+	use subxt::{
+		error::{RpcError, TransactionError},
+		tx::TxStatus,
+	};
+
+	let mut tx = extrinsic.submit_and_watch().await?;
+
+	while let Some(status) = tx.next().await {
+		match status? {
+			TxStatus::InBestBlock(tx_in_block) | TxStatus::InFinalizedBlock(tx_in_block) => {
```
AlexD10S, line 236: Just to be aware for `pop call chain` we wait until is finalized. 

Does it make sense to use the same approach  as `cargo_contracts`?. Based on the comment `cargo_contract` has here: https://github.com/use-ink/cargo-contract/blob/68691b9b6cdb7c6ec52ea441b3dc31fcb1ce08e0/crates/extrinsics/src/lib.rs#L171 this code is specific for `substrate-contracts-node`, but are there any potential use cases for secure signing on the testing chain `substrate-contracts-node`?

PR #367, opened 2024-12-09
refactor: various fixes and improvements

evilrobot-01:
Various fixes and improvements to `pop call parachain`.

### DIFF
--- a/crates/pop-parachains/src/call/metadata/mod.rs
+++ b/crates/pop-parachains/src/call/metadata/mod.rs
@@ -56,9 +56,7 @@ impl Display for Extrinsic {
 /// * `client`: The client to interact with the chain.
 ///
 /// NOTE: pallets are ordered by their index within the runtime by default.
-pub async fn parse_chain_metadata(
-	client: &OnlineClient<SubstrateConfig>,
-) -> Result<Vec<Pallet>, Error> {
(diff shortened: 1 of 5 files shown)


crates/pop-parachains/src/call/metadata/mod.rs
```diff
@@ -14,6 +14,8 @@ pub mod params;
 pub struct Pallet {
 	/// The name of the pallet.
 	pub name: String,
+	/// The index of the pallet within the runtime.
+	pub index: u8,
```
evilrobot-01, line 18: Added to enable sorting by name or index.

crates/pop-parachains/src/call/metadata/mod.rs
```diff
@@ -31,6 +33,8 @@ impl Display for Pallet {
 pub struct Extrinsic {
 	/// The name of the extrinsic.
 	pub name: String,
+	/// The index of the extrinsic within the pallet.
+	pub index: u8,
```
evilrobot-01, line 37: Added to enable sorting by name or index.

AlexD10S (approved):
Looks great! Thanks for cleaning up unnecessary code and improving the readability of the pallet and extrinsics.

PR #368, opened 2024-12-09
refactor: terminology

evilrobot-01:
Refactors to disambiguate between dispatchable functions and extrinsics.


AlexD10S (approved):
Based on the definitions provided in the resource you shared: [Polkadot's documentation on account abstraction](https://wiki.polkadot.network/docs/learn-account-abstraction#origin-abstraction-in-polkadot), it technically makes sense to rename it to _dispatchable function_ instead of `extrinsic`.

PR #373, opened 2024-12-10
refactor: rename parachain with chain as the primary command and retain parachain as an alias

AlexD10S:
Simple renaming to use `chain` instead of `parachain` as the primary command for `pop call chain` but retainining `parachain` as an alias.

Example help output:
```
Call a chain or a smart contract

Usage: pop call
       pop call <COMMAND>

Commands:
  chain     Call a chain [aliases: parachain]
  contract  Call a contract
  help      Print this message or the help of the given subcommand(s)
```

### DIFF
--- a/crates/pop-cli/src/commands/call/mod.rs
+++ b/crates/pop-cli/src/commands/call/mod.rs
@@ -15,13 +15,13 @@ pub(crate) struct CallArgs {
 	pub command: Command,
 }
 
-/// Call a smart contract.
+/// Call a chain or a smart contract.
 #[derive(Subcommand)]
 pub(crate) enum Command {
-	/// Call a parachain.
+	/// Call a chain
 	#[cfg(feature = "parachain")]
-	#[clap(alias = "p")]
-	Parachain(parachain::CallParachainCommand),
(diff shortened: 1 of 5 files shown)


al3mart (changes requested):
Any reason for not renaming the struct `CallParachain`  ?
It seems we are not using it outside of this module, so I'd say let's apply the same naming change to it :)

evilrobot-01:
> Any reason for not renaming the struct `CallParachain`? 

I had wanted to change that to `Call`, so would be happy to see it renamed. As you suggest, its an internal type so the name doesnt matter too much to me, apart from shifting away from Parachain.

crates/pop-cli/src/commands/call/mod.rs
```diff
@@ -2,10 +2,10 @@
 
 use clap::{Args, Subcommand};
 
+#[cfg(feature = "parachain")]
```
evilrobot-01, line 5: We may need to remove this conditional feature in the future.

evilrobot-01 (approved):
LGTM. It would be good to address the point Ale made, but as its internal I dont consider it a blocker right now so up to you.

al3mart (approved):
Looking good. Thanks!

PR #375, opened 2024-12-11
refactor: events

Daanvdplas:


crates/pop-common/src/lib.rs
```diff
@@ -73,6 +73,12 @@ pub fn find_free_port() -> u16 {
 		.port()
 }
 
+pub mod call {
```
AlexD10S, line 76: Nice! I like you use `pop_common` crate for this

crates/pop-parachains/src/call/mod.rs
```diff
@@ -50,6 +51,7 @@ pub fn construct_sudo_extrinsic(xt: DynamicPayload) -> Result<DynamicPayload, Er
 /// * `suri` - The secret URI (e.g., mnemonic or private key) for signing the extrinsic.
 pub async fn sign_and_submit_extrinsic(
 	client: &OnlineClient<SubstrateConfig>,
+	url: &url::Url,
```
AlexD10S, line 54: Does make sense to pass &`chain` instead? that already contains `client` and `url` in it?

Is just a small comment, up to you. But whatever approach you use, update the comment of the function.

Daanvdplas: The problem is that the pop-parachain crate then has to take a dependency of the pop-cli crate. Or I move the `Chain` struct to the `pop-parachains` crate but it would only be used for that parameter.

AlexD10S: Ah I see. Leave it like this then.
`pop_parachain` crate can be use but other services that doesn't need to know what the struct Chain is.

PR #377, opened 2024-12-12
feat(up-contract): serve HTML from string, fix deployment message on error, cors, and other misc. improvements

peterwht:
frontend from: https://github.com/r0gue-io/wallet-signing-portal

based on: https://github.com/r0gue-io/pop-cli/pull/365


AlexD10S (commented):
Changes look good. 
Probably frontend feedback better here than in the frontend repository:

The flow looks great, nice work! It would be even better with the frontend a bit more polished. 
My main feedback is about the buttons: `Terminate` appears bigger than `Sign`, which can be confusing. I suggest either moving `Terminate` below as a text-only button or emphasizing `Sign` as the primary button to make its importance clearer.

I know you're already aware, but just to have it written down so we don't forget: the `dry_run` functionality still needs to be implemented.

peterwht:
superseded by https://github.com/r0gue-io/pop-cli/pull/380

PR #374, opened 2024-12-11
test(wallet-integration): unit tests for pop up contract

al3mart:
This PR covers the addition of unit tests for the wallet integration with `pop up contracts`.

~~`fn wait_for_signature` is tested as part of an integration test in `pop-cli/tests/contracts.rs`.~~
The integration test the https server handling the signature of payloads has been included in the test `contract_lifecycle` in `pop-cli/tests/contracts.rs`.

Retrieving the correct upload and instantiation call data is tested in `pop-cli/src/commands/up/contract.rs` in `get_upload_call_data_works` and `get_instantiate_call_data_works` tests.

[sc-1887]

PR #378, opened 2024-12-12
feat: wallet integration call contract

AlexD10S:
The tests failing or ignored are from the main feat branch and is being tackled in this PR: https://github.com/r0gue-io/pop-cli/pull/374

The `dry-run` in the frontend is a pending feature is being work on another PR.

[sc-1719]


crates/pop-cli/src/commands/call/contract.rs
```diff
@@ -115,14 +120,19 @@ impl CallContractCommand {
 			full_message.push_str(&format!(" --gas {}", gas_limit));
 		}
 		if let Some(proof_size) = self.proof_size {
-			full_message.push_str(&format!(" --proof_size {}", proof_size));
+			full_message.push_str(&format!(" --proof-size {}", proof_size));
```
AlexD10S, line 123: No related to this feature but a very easy fix, display correctly the flags: `--proof-size`

peterwht (commented):
Looking really good! The flow feels nice.

Only question is about the dry run using `//Alice` even if use wallet was selected.

AlexD10S:
> Looking really good! The flow feels nice.
> 
> Only question is about the dry run using `//Alice` even if use wallet was selected.

Good catch, is a bug, it should trigger the browser too. Fixed https://github.com/r0gue-io/pop-cli/pull/378/commits/98f4de2b4c2b6a239f49207d82cdece88bd7c484

crates/pop-contracts/src/call.rs
```diff
@@ -173,6 +176,53 @@ pub async fn call_smart_contract(
...
+/// Generates the payload for executing a smart contract call.
+///
+/// # Arguments
+/// * `call_exec` - A struct containing the details of the contract call.
+/// * `gas_limit` - The maximum amount of gas allocated for executing the contract call.
+pub fn get_call_payload(
+	call_exec: &CallExec<DefaultConfig, DefaultEnvironment, Keypair>,
+	gas_limit: Weight,
+) -> anyhow::Result<Vec<u8>> {
+	let storage_deposit_limit: Option<u128> = None;
```
peterwht, line 212: any reason not to use `call_exec.opts.storage_deposit_limit`?

peterwht (commented):
Great job! Pushed a couple small tweaks.

The biggest thing -- which I didn't have the time to fix tonight -- is that we don't need the signer when someone is performing a get. We can simply use the RPC.

So, in the case it's a read-only function, then we don't care if --suri or --use-wallet is provided.

crates/pop-contracts/src/call.rs
AlexD10S: You are right, it makes more sense. Are you changing it in the `pop up contract` too? https://github.com/r0gue-io/pop-cli/pull/365/files#diff-1352e29eed34e9a7d6dadab47b3cba66c1838625bbc8533fdf9a5246bc8c1717R138

AlexD10S: https://github.com/r0gue-io/pop-cli/pull/378/commits/55367b00f420ca4461977eea16115961cd0c52c2

AlexD10S:
> Great job! Pushed a couple small tweaks.
> 
> The biggest thing -- which I didn't have the time to fix tonight -- is that we don't need the signer when someone is performing a get. We can simply use the RPC.
> 
> So, in the case it's a read-only function, then we don't care if --suri or --use-wallet is provided.

Good point. Skip `suri` prompt for read-only calls in https://github.com/r0gue-io/pop-cli/pull/378/commits/81738f4f15003e126a73f7fdbb7a487f582d2476

Also fixed when `--use-wallet` is specified using the command line in a read-only call by showing a warning and disabling the flag to prevent unnecessary signing (https://github.com/r0gue-io/pop-cli/pull/378/commits/6327d7e2f4c5d1a1b61fa75db1f4e86d539ead2c).  Example:
```
pop call contract --path ../my_contract --contract 5F3brxNL5hrCbzUerT9KMy2neVzycH3nQ5D9UeGPgUatJArE --message get --url ws://localhost:9944/ --use-wallet
```

peterwht (approved):
LGTM! Thank you!

PR #385, opened 2024-12-16
chore: set msrv

evilrobot-01:
subxt 0.38 uses features only available from Rust 1.81: https://github.com/paritytech/subxt/commit/7d1002192e5b69947052ae8754f4e85d43d242df


AlexD10S (approved):
Thank you for taking care of this!

Issue #353, opened 2024-11-27
refactor: use proper #[arg] in all the clap arguments 


AlexD10S:
In the code, we currently use both `#[arg] `and `#[clap] `to define arguments, for example:
```rs
        #[arg(short = 'o', long = "output")]
	pub(crate) output_file: Option<PathBuf>,
	/// For production, always build in release mode to exclude debug features.
	#[clap(short = 'r', long, default_value = "true")]
	pub(crate) release: bool, 
```
In https://github.com/r0gue-io/pop-cli/blob/main/crates/pop-cli/src/commands/build/spec.rs#L135C2-L139C27

Eeview all commands and update the argument definitions to consistently use `#[arg]`, as it aligns with the examples provided in the clap library documentation: https://github.com/clap-rs/clap

PR #386, opened 2024-12-16
refactor: ensure short args consistency

evilrobot-01:
Aligns short arg values used across commands for consistency.

### DIFF
--- a/crates/pop-cli/src/commands/up/contract.rs
+++ b/crates/pop-cli/src/commands/up/contract.rs
@@ -30,49 +30,49 @@ const FAILED: &str = "🚫 Deployment failed.";
 #[derive(Args, Clone)]
 pub struct UpContractCommand {
 	/// Path to the contract build directory.
-	#[arg(short = 'p', long)]
+	#[arg(short, long)]
 	path: Option<PathBuf>,
 	/// The name of the contract constructor to call.
-	#[clap(name = "constructor", long, default_value = "new")]
+	#[clap(short, long, default_value = "new")]
 	constructor: String,
 	/// The constructor arguments, encoded as strings.
-	#[clap(long, num_args = 0..,)]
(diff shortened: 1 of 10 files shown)


crates/pop-cli/src/commands/up/contract.rs
```diff
@@ -30,49 +30,49 @@ const FAILED: &str = "🚫 Deployment failed.";
...
-	#[clap(name = "suri", long, short, default_value = "//Alice")]
+	#[clap(short, long, default_value = "//Alice")]
 	suri: String,
 	/// Perform a dry-run via RPC to estimate the gas usage. This does not submit a transaction.
-	#[clap(long)]
+	#[clap(short = 'D', long)]
 	dry_run: bool,
 	/// Uploads the contract only, without instantiation.
-	#[clap(short('u'), long)]
+	#[clap(short = 'U', long)]
```
evilrobot-01, line 71: Note: this is a (minor) breaking change, required so that -u is consistent for url across calling and uploading contracts. I figure upload only is probably used rarely and therefor `may` be safe to make this change now, to resolve the inconsistencies across commands which is much more frustrating in my view.

For example, I just want to do `pop up contract -u my_rpc ...` and then `pop call contract -u my_rpc ...`

AlexD10S (approved):
Looks good! 
It would be a good idea to use this refactor to address https://github.com/r0gue-io/pop-cli/issues/353 and ensure all commands are consistent by using `arg` instead of `clap`. However, not blocking to me, this can also be handled in a separate PR.

evilrobot-01:
> It would be a good idea to use this refactor to address #353 and ensure all commands are consistent by using `arg` instead of `clap`. However, not blocking to me, this can also be handled in a separate PR.

I'd rather spend the time on the hrmp PR at this current time.

PR #387, opened 2024-12-16
feat(call-parachain): add remark action

evilrobot-01:
Adds an action to make a remark, useful for testing. Also refactors the tests to simplify and fixes some incorrect language.

### DIFF
--- a/crates/pop-parachains/src/call/metadata/action.rs
+++ b/crates/pop-parachains/src/call/metadata/action.rs
@@ -43,19 +43,19 @@ pub enum Action {
 		props(Pallet = "Assets")
 	)]
 	MintAsset,
-	/// Create an NFT collection.
+	/// Create a NFT collection.
 	#[strum(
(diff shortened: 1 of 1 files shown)


crates/pop-parachains/src/call/metadata/action.rs
```diff
@@ -43,19 +43,19 @@ pub enum Action {
 		props(Pallet = "Assets")
 	)]
 	MintAsset,
-	/// Create an NFT collection.
+	/// Create a NFT collection.
```
AlexD10S, line 46: I guess 'a NFT'  is grammatically correct, but I’ve seen 'an NFT' being used frequently. For example: [Polkadot Wiki](https://wiki.polkadot.network/docs/learn-nft-projects#rmrk).

AlexD10S (approved):
Nice! I like to see this list grow based on user feedback.

PR #393, opened 2024-12-18
refactor(up): bump fallback versions

evilrobot-01:
Bumps the fallback versions of the relay, parachains and chain specs.

### DIFF
--- a/crates/pop-parachains/src/up/mod.rs
+++ b/crates/pop-parachains/src/up/mod.rs
@@ -660,6 +660,8 @@ mod tests {
 	use std::{env::current_dir, fs::File, io::Write};
 	use tempfile::tempdir;
 
+	pub(crate) const VERSION: &str = "stable2409";
+
 	mod zombienet {
 		use super::*;
 		use pop_common::Status;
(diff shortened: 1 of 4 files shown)


peterwht:
not sure if still good to review, but out of curiosity, any reason not to kickoff another release in our polkadot fork to use 2412?

evilrobot-01:
> not sure if still good to review, but out of curiosity, any reason not to kickoff another release in our polkadot fork to use 2412?

These are only the fallback versions in case it cant connect to GH - e.g. rate limited. It should still check for a newer version if it exists. No reason not to kick off the stable2412 release, will give it a go.

crates/pop-parachains/src/up/mod.rs
```diff
@@ -907,12 +906,12 @@ id = 1000
 chain = "asset-hub-paseo-local"
 "#
 			)?;
-			let system_parachain_version = "v1.12.0";
+			let system_parachain_version = "stable2407";
```
peterwht, line 909: Is there a reason this is `2407` instead of `2409?

peterwht (approved):
Left one comment. But I am going to approve as I am sure there is a reason for the 2407 and I don't want my timezone to block.

crates/pop-parachains/src/up/mod.rs
evilrobot-01: Its just testing that a different version for a system parachain is used from the relay version.

AlexD10S (approved):
Looks good!

PR #297, opened 2024-08-28
feat: update parachain templates

AlexD10S:
This PR introduces updates to the templates.

- It introduces the new Parity parachain template: [Polkadot SDK's Parachain Template](https://github.com/paritytech/polkadot-sdk-parachain-template), which replaces the existing contracts template as the default for Parity.
_Closes https://github.com/r0gue-io/pop-cli/issues/238_

- It removes the [Frontier Parachain template](https://github.com/paritytech/frontier-parachain-template) due to its lack of maintenance (no updates in the past 9 months). 
_For now, we will retain the [Contracts Parachain](https://github.com/paritytech/substrate-contracts-node), as it is actively maintained once again._

- It updates the supported versions for the OpenZeppelin templates after its release.
_The current logic for the supported versions needs to be updated: https://github.com/r0gue-io/pop-cli/issues/400_

[sc-1334]

PR #394, opened 2024-12-19
chore: release v0.6.0

AlexD10S:
- [X]  Update crate versions to 0.6.0.
- [X]  Update Changelog file. Run `git cliff --bump`, but fixed manually.
- [X]  Remove Deprecated flags or commands: 
          - Remove the subcommands `parachain` and `contract` in pop build.
          - Old feature to run e2e tests..
          - OpenZeppelin template name changed.
_Note: In this release, a new deprecated flag has been introduced. The release flag is being deprecated in favor of profile (see [source](https://github.com/r0gue-io/pop-cli/blob/main/crates/pop-cli/src/commands/build/spec.rs#L141))._
_Note 2: Parity templates will be addressed in https://github.com/r0gue-io/pop-cli/pull/297_

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,10 @@ All notable changes to this project will be documented in this file.
 - Replace rococo to paseo name (#333)
 - Replace codeowners (#388)
 
+### Build
+
+- *(deps)* Remove unused dependencies (#369)
+
 ## [0.5.0] - 2024-11-08
 
 ### 🚀 Features
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ All notable changes to this project will be documented in this file.
 - Wallet integration (#371)
 - Guide user to call a chain (#316)
 - Guide user to call a contract (#306)
-- Call chain events (#372)
+- Output events after calling a chain (#372)
 - Add remark action (#387)
 
 ### 🐛 Fixes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,33 @@
 
 All notable changes to this project will be documented in this file.
 
+## [0.6.0] - 2024-12-19
+
+### 🚀 Features
+
+- Wallet integration (#371)
+- Guide user to call a chain (#316)
+- Guide user to call a contract (#306)
+- Call chain events (#372)
+- Add remark action (#387)
+
+### 🐛 Fixes
+
+- Build spec experience (#331)
+- HRMP channels (#278)
+
+### 🚜 Refactor
+
+- Ensure short args consistency (#386)
+- Bump fallback versions (#393)
+
+### ⚙️ Miscellaneous Tasks
+
+- Bump zombienet version to `v0.2.18` (#352)
+- Set msrv (#385)
+- Replace rococo to paseo name (#333)
+- Replace codeowners (#388)
+
 ## [0.5.0] - 2024-11-08
 
 ### 🚀 Features
--- a/crates/pop-cli/src/commands/build/contract.rs
+++ b/crates/pop-cli/src/commands/build/contract.rs
@@ -4,9 +4,9 @@ use crate::cli;
 use pop_contracts::{build_smart_contract, Verbosity};
 use std::path::PathBuf;
 
-/// Represents the configuration for building a smart contract.
+/// Configuration for building a smart contract.
 pub struct BuildContract {
-	/// Path for the contract project.
+	/// Path of the contract project.
 	pub(crate) path: Option<PathBuf>,
 	/// Build profile: `true` for release mode, `false` for debug mode.
 	pub(crate) release: bool,
--- a/crates/pop-cli/src/commands/build/contract.rs
+++ b/crates/pop-cli/src/commands/build/contract.rs
@@ -4,16 +4,15 @@ use crate::cli;
 use pop_contracts::{build_smart_contract, Verbosity};
 use std::path::PathBuf;
 
-/// Command to build a smart contract with configurable path and release mode.
-pub struct BuildContractCommand {
-	/// Path for the contract project [default: current directory]
+/// Represents the configuration for building a smart contract.
+pub struct BuildContract {
+	/// Path for the contract project.
 	pub(crate) path: Option<PathBuf>,
-	/// The default compilation includes debug functionality, increasing contract size and gas
-	/// usage. For production, always build in release mode to exclude debug features.
+	/// Build profile: `true` for release mode, `false` for debug mode.
 	pub(crate) release: bool,
 }
 
-impl BuildContractCommand {
+impl BuildContract {
 	/// Executes the command.
 	pub(crate) fn execute(self) -> anyhow::Result<&'static str> {
 		self.build(&mut cli::Cli)
@@ -55,7 +54,7 @@ mod tests {
 				.expect_outro("Build completed successfully!");
 
 			assert_eq!(
-				BuildContractCommand { path: Some(path.join(name)), release }.build(&mut cli)?,
+				BuildContract { path: Some(path.join(name)), release }.build(&mut cli)?,
 				"contract"
 			);
 
--- a/crates/pop-cli/src/commands/build/contract.rs
+++ b/crates/pop-cli/src/commands/build/contract.rs
@@ -4,8 +4,6 @@ use crate::cli;
 use clap::Args;
 use pop_contracts::{build_smart_contract, Verbosity};
 use std::path::PathBuf;
-#[cfg(not(test))]
-use std::{thread::sleep, time::Duration};
 
 #[derive(Args)]
 pub struct BuildContractCommand {
@@ -16,9 +14,6 @@ pub struct BuildContractCommand {
 	/// usage. For production, always build in release mode to exclude debug features.
 	#[clap(short, long)]
 	pub(crate) release: bool,
-	// Deprecation flag, used to specify whether the deprecation warning is shown.
-	#[clap(skip)]
-	pub(crate) valid: bool,
 }
 
 impl BuildContractCommand {
@@ -33,14 +28,6 @@ impl BuildContractCommand {
 	/// * `cli` - The CLI implementation to be used.
 	fn build(self, cli: &mut impl cli::traits::Cli) -> anyhow::Result<&'static str> {
 		cli.intro("Building your contract")?;
-
-		// Show warning if specified as deprecated.
-		if !self.valid {
-			cli.warning("NOTE: this command is deprecated. Please use `pop build` (or simply `pop b`) in future...")?;
-			#[cfg(not(test))]
-			sleep(Duration::from_secs(3));
-		}
-
 		// Build contract.
 		let build_result =
 			build_smart_contract(self.path.as_deref(), self.release, Verbosity::Default)?;
@@ -66,23 +53,16 @@ mod tests {
 		create_smart_contract(name, &path.join(name), &Standard)?;
 
 		for release in [false, true] {
-			for valid in [false, true] {
-				let mut cli = MockCli::new()
-					.expect_intro("Building your contract")
-					.expect_outro("Build completed successfully!");
+			let mut cli = MockCli::new()
+				.expect_intro("Building your contract")
+				.expect_outro("Build completed successfully!");
 
-				if !valid {
-					cli = cli.expect_warning("NOTE: this command is deprecated. Please use `pop build` (or simply `pop b`) in future...");
-				}
+			assert_eq!(
+				BuildContractCommand { path: Some(path.join(name)), release }.build(&mut cli)?,
+				"contract"
+			);
 
-				assert_eq!(
-					BuildContractCommand { path: Some(path.join(name)), release, valid }
-						.build(&mut cli)?,
-					"contract"
-				);
-
-				cli.verify()?;
-			}
+			cli.verify()?;
 		}
 
 		Ok(())
--- a/crates/pop-cli/src/commands/build/contract.rs
+++ b/crates/pop-cli/src/commands/build/contract.rs
@@ -1,18 +1,15 @@
 // SPDX-License-Identifier: GPL-3.0
 
 use crate::cli;
-use clap::Args;
 use pop_contracts::{build_smart_contract, Verbosity};
 use std::path::PathBuf;
 
-#[derive(Args)]
+/// Command to build a smart contract with configurable path and release mode.
 pub struct BuildContractCommand {
 	/// Path for the contract project [default: current directory]
-	#[arg(long)]
 	pub(crate) path: Option<PathBuf>,
 	/// The default compilation includes debug functionality, increasing contract size and gas
 	/// usage. For production, always build in release mode to exclude debug features.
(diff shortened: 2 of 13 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,33 @@
...
 All notable changes to this project will be documented in this file.
 
+## [0.6.0] - 2024-12-19
+
+### 🚀 Features
+
+- Wallet integration (#371)
+- Guide user to call a chain (#316)
+- Guide user to call a contract (#306)
+- Call chain events (#372)
```
Daanvdplas, line 12:
```suggestion
- Output events after calling a chain (#372)
```

CHANGELOG.md
```diff
@@ -2,6 +2,33 @@
...
+- Add remark action (#387)
+
+### 🐛 Fixes
+
+- Build spec experience (#331)
+- HRMP channels (#278)
```
Daanvdplas, line 18: In terms of posting, this definitely deserves a post as it makes deploying chains much easier.

crates/pop-cli/src/commands/build/mod.rs
```diff
@@ -44,15 +44,6 @@ pub(crate) struct BuildArgs {
 /// Build a parachain, smart contract or Rust package.
```
Daanvdplas, line 44: Update docs

crates/pop-cli/src/commands/build/parachain.rs
```diff
@@ -42,12 +39,7 @@ impl BuildParachainCommand {
 		cli.intro(format!("Building your {project}"))?;
 
 		let profile = self.profile.unwrap_or(Profile::Debug);
```
Daanvdplas, line 41: `profile` field doesn't require a `Option` value as it is initialised in `Command.execute`. This can be considered for the other fields as well (perhaps not the package)

crates/pop-cli/src/commands/build/parachain.rs
```diff
@@ -22,9 +22,6 @@ pub struct BuildParachainCommand {
 	/// Parachain ID to be used when generating the chain spec files.
 	#[arg(short, long)]
```
Daanvdplas, line 23: Do we still need those here as it is not a command anymore. Also perhaps a name change of this type as it is not a command anymore.

crates/pop-cli/src/commands/build/contract.rs
```diff
@@ -16,9 +14,6 @@ pub struct BuildContractCommand {
 	/// usage. For production, always build in release mode to exclude debug features.
 	#[clap(short, long)]
```
Daanvdplas, line 15: Do we still need those here as it is not a command anymore. That also gives reason to change the name of the type.

crates/pop-cli/src/commands/build/parachain.rs
```diff
@@ -22,9 +22,6 @@ pub struct BuildParachainCommand {
...
 	pub(crate) id: Option<u32>,
```
Daanvdplas, line 24: Are we using this value anywhere?

crates/pop-cli/src/commands/build/contract.rs
Daanvdplas: The above `path` field can also already be initialised and thus remove the `Option`

crates/pop-cli/src/commands/build/mod.rs
Daanvdplas: Cannot point to the `BuildsArgs` struct, but do we still need the `id` field there as it is not needed for building a chain / contract and it is included in the `BuildSpecCmd`?

crates/pop-cli/src/commands/build/parachain.rs
Daanvdplas: Cannot point the `Option` fields above but some don't have to be anymore.

Daanvdplas (commented):
Was only able to provide input for the removal of the `BuildParachainCmd` and `BuildContractCmd` as I was not involved with the other deprecations.

crates/pop-cli/src/commands/build/mod.rs
AlexD10S: Good point, `id` here is not needed anymore. Removed and update docs in https://github.com/r0gue-io/pop-cli/pull/394/commits/b0b5946973c0f24ff07ef0d539ecd381b33f8fc4

crates/pop-cli/src/commands/build/parachain.rs
AlexD10S: You are right, for profile and path can Option can be removed: https://github.com/r0gue-io/pop-cli/pull/394/commits/b0b5946973c0f24ff07ef0d539ecd381b33f8fc4

crates/pop-cli/src/commands/build/parachain.rs
AlexD10S: No need for macros anymore, cleaned in https://github.com/r0gue-io/pop-cli/pull/394/commits/b0b5946973c0f24ff07ef0d539ecd381b33f8fc4

crates/pop-cli/src/commands/build/contract.rs
AlexD10S: No need for macros anymore, cleaned in https://github.com/r0gue-io/pop-cli/pull/394/commits/b0b5946973c0f24ff07ef0d539ecd381b33f8fc4

For contracts is ok to keep path as Optional as the `build_smart_contract` method in the `pop_contracts` crate is expecting an Optional value.

CHANGELOG.md
AlexD10S: Usually, the tool grabs the name from the commit (Call parachain events), but since we’re updating manually, I’ll use the suggested one.

crates/pop-cli/src/commands/build/mod.rs
```diff
@@ -35,13 +35,9 @@ pub(crate) struct BuildArgs {
...
 	#[clap(long, value_enum)]
 	pub(crate) profile: Option<Profile>,
-	/// Parachain ID to be used when generating the chain spec files.
-	#[arg(short = 'i', long = "id")]
-	#[cfg(feature = "parachain")]
-	pub(crate) id: Option<u32>,
 }
 
-/// Build a parachain, smart contract or Rust package.
+/// Build a parachain, smart contract, chain specification or Rust package.
```
Daanvdplas, line 40: This comment is incorrect as it only provides the possibility for building a chain specification. The others are not included here as they fall under the BuildArgs

crates/pop-cli/src/commands/build/contract.rs
```diff
@@ -1,18 +1,15 @@
 // SPDX-License-Identifier: GPL-3.0
 
 use crate::cli;
-use clap::Args;
 use pop_contracts::{build_smart_contract, Verbosity};
 use std::path::PathBuf;
 
-#[derive(Args)]
+/// Command to build a smart contract with configurable path and release mode.
 pub struct BuildContractCommand {
```
Daanvdplas, line 8: It is not a command anymore. Something along the lines of the following might be better:
```
/// Build contract options.
pub struct BuildContract {
```

The fields docs also need to be changed. The `release` field can be as simple as the profile docs of the `BuildParachainCmd`

crates/pop-cli/src/commands/build/parachain.rs
```diff
@@ -1,27 +1,20 @@
...
-use clap::Args;
 use pop_common::Profile;
 use pop_parachains::build_parachain;
 use std::path::PathBuf;
 #[cfg(not(test))]
 use std::{thread::sleep, time::Duration};
 
-#[derive(Args)]
+/// Command to build a parachain with configurable path, package, and profile.
 pub struct BuildParachainCommand {
```
Daanvdplas, line 11: Also not a command anymore, suggest a similar change as the BuildContractCmd

crates/pop-cli/src/commands/build/parachain.rs
```diff
@@ -1,27 +1,20 @@
...
 	/// Directory path for your project [default: current directory].
```
Daanvdplas, line 12: Default part is also no longer necessary as it was there to explain what the command would default to (same for BuildContractsCmd)

Daanvdplas (commented):
I think the two Build_Cmd types can be refactored a little bit more

crates/pop-cli/src/commands/build/mod.rs
AlexD10S: It doesn't show this in the `--help` anymore, but you are right: _Build a chain specification and its genesis artifacts._

crates/pop-cli/src/commands/build/contract.rs
AlexD10S: Thanks for the feedback, it doesn't make sense to keep the word Command in the name, or the command description as it was before. Refactored in https://github.com/r0gue-io/pop-cli/pull/394/commits/f1963e7bd59df8820b984bc5ae0aaaa95adaaee4

crates/pop-cli/src/commands/build/parachain.rs
AlexD10S: As above https://github.com/r0gue-io/pop-cli/pull/394/commits/f1963e7bd59df8820b984bc5ae0aaaa95adaaee4

crates/pop-cli/src/commands/build/parachain.rs
AlexD10S: Good point, feedback, it doesn't make sense to keep the comment as it was before, to display to the user. Refactored in https://github.com/r0gue-io/pop-cli/pull/394/commits/f1963e7bd59df8820b984bc5ae0aaaa95adaaee4

crates/pop-cli/src/commands/build/mod.rs
```diff
@@ -37,7 +37,7 @@ pub(crate) struct BuildArgs {
 	pub(crate) profile: Option<Profile>,
 }
 
-/// Build a parachain, smart contract, chain specification or Rust package.
+/// Build a chain specification and its genesis artifacts.
```
Daanvdplas, line 40: So this is technically an umbrella type for commands, now build spec being the only one.

"Subcommand for building chain artifacts" might be an oke description to go with for now

AlexD10S: Is tricky being the only one subcommand, that's why I used the same comment. But probably you are right and is better to indicate is a subcommand in the command.

AlexD10S: https://github.com/r0gue-io/pop-cli/pull/394/commits/c3a724e9482308f8a4204aadcc5e433390b78f66

Daanvdplas: Thank you!

Yeah otherwise the umbrella type should be removed but it might become useful in the future so think it is better to leave it in. It is also out of scope for now but keeping things correct and of a certain quality should always be strived for :)

crates/pop-cli/src/commands/build/contract.rs
```diff
@@ -1,27 +1,18 @@
...
 use pop_contracts::{build_smart_contract, Verbosity};
 use std::path::PathBuf;
-#[cfg(not(test))]
-use std::{thread::sleep, time::Duration};
 
-#[derive(Args)]
-pub struct BuildContractCommand {
-	/// Path for the contract project [default: current directory]
-	#[arg(long)]
+/// Represents the configuration for building a smart contract.
```
Daanvdplas, line 7:
```suggestion
/// Configuration for building a smart contract.
```

crates/pop-cli/src/commands/build/mod.rs
```diff
@@ -65,12 +52,12 @@ impl Command {
 		// If only contract feature enabled, build as contract
 		#[cfg(feature = "contract")]
 		if pop_contracts::is_supported(args.path.as_deref())? {
-			// All commands originating from root command are valid
+			// All arguments originating from root command are valid
```
Daanvdplas, line 55: I don't get this command

crates/pop-cli/src/commands/build/mod.rs
```diff
@@ -81,13 +68,11 @@ impl Command {
 				Some(profile) => profile,
 				None => args.release.into(),
 			};
-			// All commands originating from root command are valid
-			BuildParachainCommand {
-				path: args.path,
+			// All arguments originating from root command are valid.
```
Daanvdplas, line 71: Also don't get

crates/pop-cli/src/commands/build/parachain.rs
```diff
@@ -1,34 +1,24 @@
...
 use std::path::PathBuf;
 #[cfg(not(test))]
 use std::{thread::sleep, time::Duration};
 
-#[derive(Args)]
-pub struct BuildParachainCommand {
-	/// Directory path for your project [default: current directory].
-	#[arg(long)]
-	pub(crate) path: Option<PathBuf>,
+// Represents the configuration for building a parachain.
```
Daanvdplas, line 10:
```suggestion
// Configuration for building a parachain.
```

crates/pop-cli/src/commands/build/parachain.rs
```diff
@@ -1,34 +1,24 @@
...
+pub struct BuildParachain {
+	/// Directory path for your project.
```
Daanvdplas, line 12:
```suggestion
	/// Path of the parachain project.
```

crates/pop-cli/src/commands/build/contract.rs
```diff
@@ -1,27 +1,18 @@
...
+pub struct BuildContract {
+	/// Path for the contract project.
```
Daanvdplas, line 9:
```suggestion
	/// Path of the contract project.
```

crates/pop-contracts/Cargo.toml
```diff
@@ -33,8 +33,8 @@ contract-build.workspace = true
 contract-extrinsics.workspace = true
 contract-transcode.workspace = true
 scale-info.workspace = true
-# pop
-pop-common = { path = "../pop-common", version = "0.5.0" }
+#  pop
```
Daanvdplas, line 36:
```suggestion
# pop
```

Daanvdplas (approved):
Apart from some final nitpicks looking good! 

Note: I am not so familar with the openzeppelin and e2e tests so this should be reviewed by someone who is

crates/pop-cli/src/commands/build/parachain.rs
AlexD10S: This isn’t a change in this PR, so I prefer to leave the previous comment for now.

AlexD10S:
> Note: I am not so familar with the openzeppelin and e2e tests so this should be reviewed by someone who is

Thanks!
The tricky part was removing the deprecated code from the build. e2e tests and OZ old naming template is a very straightforward removal. Anyway, I’ve requested another review just to be sure.

crates/pop-cli/src/commands/build/parachain.rs
Daanvdplas: It is a very small change and makes it consistent with the other command (which is changed)

peterwht (approved):

PR #396, opened 2025-01-03
fix: check_contracts_node handles skip_confirm

al3mart:
There was a small bug where the `skip_confirm` was ignored.
Prompting the user to source `sustrate-contracts-node`.

I have added a very minimal test. It could be improved and made more robust if we were to expand `MockCli` to include checks for certain prompts not being presented in the CLI. But that is another story.

Thanks to @ltfschoen for reporting.

[sc-2094]


AlexD10S (approved):
LGTM! Thanks for fixing this

PR #398, opened 2025-01-08
build(release): update upload-artifact to v4

al3mart:
As per https://github.blog/changelog/2024-04-16-deprecation-notice-v3-of-the-artifact-actions/ `upload-artifact` action needs to be updated to use v4 instead.

Will test run our release workflow before asking for reviews.


al3mart:
It seems the action fails after the change:

https://github.com/r0gue-io/pop-cli/actions/runs/12666964995/job/35299364389

PR #403, opened 2025-02-06
feat: enable pop up without project type specification

AlexD10S:
* Enables `pop up` command to run without specifying contract or parachain, automatically detecting the project type.

* Handles the deprecation of `pop up contract` and `pop up parachain`, keeping them functional for now but marking them for future removal (Reference https://github.com/r0gue-io/pop-cli/pull/222).

* Refactors `pop up` to distinguish between deploying a parachain and launching a local network.
    * `pop up parachain` is now migrated to `pop up network`. 
    * `pop up` in a parachain template will deploy a parachain instead of launching a network.  
_**Note:**_:  Currently, running `pop up` inside a parachain project **does nothing**. The deployment logic for parachains has been removed in this commit https://github.com/r0gue-io/pop-cli/pull/403/commits/38a19a84c697987e7106f110a31719d01dfb3865 and will be implemented separately in an upcoming PR https://github.com/r0gue-io/pop-cli/pull/404. 

_**Commands for testing**_
```sh
# Deploy a contract passing path
pop up ../flipper --args false
# Deploy a contract using flag --path
pop up --path ../flipper --args false
# Deploy a contract, inside the contract folder.
pop up --args false
# Deploy a contract with deprecated command. (Should display an warning message).
pop up contract --args false
```

```sh
# Launch a local network
pop up network -f ./network.toml
# Launch a local network  with deprecated command. (Should display an warning message).

# This will be used to deploy a parachain; currently, not implemented, only shows a warning.
pop up
```

[sc-2506]


crates/pop-cli/src/commands/mod.rs
```diff
@@ -53,6 +52,16 @@ fn about_build() -> &'static str {
 	return "Build a smart contract or Rust package.";
 }
 
+/// Help message for the up command.
+fn about_up() -> &'static str {
```
chungquantin, line 56: nice 💪

crates/pop-cli/src/commands/mod.rs
```diff
@@ -101,10 +110,21 @@ impl Command {
...
-				#[cfg(feature = "parachain")]
-				up::Command::Parachain(cmd) => cmd.execute().await.map(|_| Value::Null),
-				#[cfg(feature = "contract")]
-				up::Command::Contract(cmd) => cmd.execute().await.map(|_| Value::Null),
+				None => up::Command::execute(args).await.map(|t| json!(t)),
+				Some(cmd) => match cmd {
+					#[cfg(feature = "parachain")]
+					up::Command::Network(mut cmd) => {
+						cmd.valid = true;
+						cmd.execute().await.map(|_| Value::Null)
```
chungquantin, line 118: `cmd.execute().await.map(|_| Value::Null)` is used on each case, could this be refactored to only call this after match and use the `match` statement to only mutate the `cmd`.

crates/pop-cli/src/commands/up/mod.rs
```diff
@@ -1,29 +1,136 @@
...
+				skip_confirm: false,
+				valid: false,
+			},
+			command: None,
+		};
+		let mut cli = MockCli::new();
+		assert_eq!(Command::execute_project_deployment(args.clone(), &mut cli).await?, "contract");
+		cli.verify()?;
+
+		// Rust project
```
chungquantin, line 125: Could this test be separated into `detects_contract_correctly ` and `detects_rust_project_correctly`?

crates/pop-cli/src/commands/up/contract.rs
```diff
@@ -78,36 +76,37 @@ pub struct UpContractCommand {
...
 
 impl UpContractCommand {
 	/// Executes the command.
 	pub(crate) async fn execute(mut self) -> anyhow::Result<()> {
 		Cli.intro("Deploy a smart contract")?;
-
-		let project_path = get_project_path(self.path.clone(), self.path_pos.clone());
+		// Show warning if specified as deprecated.
+		if !self.valid {
+			Cli.warning("NOTE: this command is deprecated. Please use `pop up` (or simply `pop u`) in future...")?;
```
chungquantin, line 101: Not a blocker but I feel like `DEPRECATION: Please use `pop up` (or simply `pop u`) in future...` will bring a better format for deprecation notice. Below is the example from ember CLI. Otherwise, ưe have to repeat "this command is deprecated." on every deprecated command. 

image

crates/pop-cli/src/commands/up/contract.rs
```diff
@@ -29,47 +28,46 @@ const COMPLETE: &str = "🚀 Deployment complete";
 const DEFAULT_URL: &str = "ws://localhost:9944/";
 const DEFAULT_PORT: u16 = 9944;
 const FAILED: &str = "🚫 Deployment failed.";
+const HELP_HEADER: &str = "Smart Contract deployment options";
```
chungquantin, line 31:
```suggestion
const HELP_HEADER: &str = "Smart contract deployment options";
```

crates/pop-cli/src/commands/up/mod.rs
```diff
@@ -1,29 +1,136 @@
...
-	#[clap(alias = "c")]
+	/// [DEPRECATED] Deploy a smart contract (will be removed in v0.8.0).
+	#[clap(alias = "c", hide = true)]
 	Contract(contract::UpContractCommand),
 }
+
+impl Command {
+	/// Executes the command.
+	pub(crate) async fn execute(args: UpArgs) -> anyhow::Result<&'static str> {
+		Self::execute_project_deployment(args, &mut Cli).await
```
chungquantin, line 56: Curious why do we need extra function for this?

crates/pop-cli/src/commands/mod.rs
```diff
@@ -53,6 +52,16 @@ fn about_build() -> &'static str {
 	return "Build a smart contract or Rust package.";
 }
 
+/// Help message for the up command.
+fn about_up() -> &'static str {
+	#[cfg(all(feature = "parachain", feature = "contract"))]
+	return "Deploy a smart contract or launch a local network.";
+	#[cfg(all(feature = "parachain", not(feature = "contract")))]
+	return "Launch a local network.";
+	#[cfg(all(feature = "contract", not(feature = "parachain")))]
```
chungquantin, line 61:
```suggestion
	#[cfg(all(not(feature = "parachain"), feature = "parachain"))]
```

I prefer this way to keep the feature order consistent to the above.

chungquantin (changes requested):
PR looks great! I left some comments mainly for the code improvement. Some based on my personal preferences. Not a blocker though.

crates/pop-cli/src/commands/up/mod.rs
AlexD10S: The reason is only for testing. I don’t call the `execute` function directly in the tests, but rather `execute_project_deployment`, so the mocked CLI object can be passed as a parameter.

This follows the same approach we used for ### build, e.g :https://github.com/r0gue-io/pop-cli/blob/main/crates/pop-cli/src/commands/build/contract.rs#L18

crates/pop-cli/src/commands/mod.rs
AlexD10S: Is a valid point! Is there a best practice in Rust for this? 
I followed the same format as [about_build()](https://github.com/r0gue-io/pop-cli/blob/main/crates/pop-cli/src/commands/mod.rs) above, so if we change it here, we should also update `about_build()` for consistency.

crates/pop-cli/src/commands/up/contract.rs
AlexD10S: Fixed in https://github.com/r0gue-io/pop-cli/pull/403/commits/827d1ab89b4b4c94bb9bd803fb66d09798e52a3e

crates/pop-cli/src/commands/mod.rs
AlexD10S: Parachain and Contract commands will be removed in the next release and will look similar to `Build`: https://github.com/r0gue-io/pop-cli/blob/main/crates/pop-cli/src/commands/mod.rs#L88:
```rs
#[cfg(any(feature = "parachain", feature = "contract"))]
Self::Up(args) => match args.command {
  None => up::Command::execute(args).await.map(|t| json!(t)),
  Some(mut cmd) => match &mut cmd {
	  #[cfg(feature = "parachain")]
	  up::Command::Network(mut cmd) => {
		  cmd.valid = true;
		  cmd.execute().await.map(|_| Value::Null)
	  },
  },
},
```
So it probably doesn’t make sense to tackle this now. But it's a great point, I’ve added a comment in this commit https://github.com/r0gue-io/pop-cli/pull/403/commits/36b0dbe58df2897c806bdff29c77a247501f9277 to remove these two in the next release:` // TODO: Deprecated, will be removed in v0.8.0`.

crates/pop-cli/src/commands/up/mod.rs
AlexD10S: Looks cleaner! I refactored the test in https://github.com/r0gue-io/pop-cli/pull/403/commits/4b776cafb2c1cb7b911de96e85942964929c1ab2

crates/pop-cli/src/commands/mod.rs
chungquantin: I don't find any document for the best practices but GPT recommends :) : "When specifying multiple conditions, it is best to keep a consistent order for readability. The order in `all(...)` or `any(...)` does not affect behavior, but consistency makes it easier to scan."

crates/pop-cli/src/commands/mod.rs
chungquantin: Gotcha! Thanks for explanation.

crates/pop-cli/src/commands/up/contract.rs
AlexD10S: Nice feedback. I agree, more clear. Changed in https://github.com/r0gue-io/pop-cli/pull/403/commits/8093764f05f37bc34833d8aac62e1e7d134a9c49

crates/pop-cli/src/commands/mod.rs
AlexD10S: Noted for a future refactor. I am keeping it as is for consistency with the rest of the code in this PR.

crates/pop-cli/src/commands/up/network.rs
```diff
@@ -57,6 +60,11 @@ impl ZombienetCommand {
 		intro(format!("{}: Launch a local network", style(" Pop CLI ").black().on_magenta()))?;
 		set_theme(Theme);
 
+		// Show warning if specified as deprecated.
+		if !self.valid {
+			log::warning("NOTE: this command is deprecated. Please use `pop up network` (or simply `pop up n`) in future...")?;
```
chungquantin, line 65:
```suggestion
			log::warning("DEPRECATION: Please use `pop up network` (or simply `pop up n`) in future...")?;
```

chungquantin:
```
base-parachain on  pr/53 [$?] via 🦀 1.81.0
➜ pop up
▲  No contract detected. Ensure you are in a valid project directory.
```
Currently, if running `pop up` inside the parachain it throws the error contract. Based on the PR description, I assume you are already aware of it? If so, everything looks good to me and I am happy to approve.

AlexD10S:
> ```
> base-parachain on  pr/53 [$?] via 🦀 1.81.0
> ➜ pop up
> ▲  No contract detected. Ensure you are in a valid project directory.
> ```
> 
> Currently, if running `pop up` inside the parachain it throws the error contract. Based on the PR description, I assume you are already aware of it? If so, everything looks good to me and I am happy to approve.

Yes, the `pop up` in the parachain folder is addressed in this PR: https://github.com/r0gue-io/pop-cli/pull/404. To launch the local network, you can use either the new `pop up network` or the existing (deprecated) `pop up parachain`.

crates/pop-cli/src/commands/up/network.rs
AlexD10S: Good catch, fixed in https://github.com/r0gue-io/pop-cli/pull/403/commits/4913c92b5604b2d19c30a972297f4886460c6f74

chungquantin (approved):
LGTM!

crates/pop-cli/src/commands/mod.rs
```diff
@@ -30,8 +30,7 @@ pub(crate) enum Command {
 	#[clap(alias = "c")]
 	#[cfg(any(feature = "parachain", feature = "contract"))]
 	Call(call::CallArgs),
-	/// Launch a local network or deploy a smart contract.
```
Daanvdplas, line 33: Why are these docs removed?

crates/pop-cli/src/commands/mod.rs
```diff
@@ -53,6 +52,16 @@ fn about_build() -> &'static str {
 	return "Build a smart contract or Rust package.";
 }
 
+/// Help message for the up command.
```
Daanvdplas, line 55: `up`

crates/pop-cli/src/commands/up/contract.rs
```diff
@@ -78,36 +76,37 @@ pub struct UpContractCommand {
...
 
 impl UpContractCommand {
 	/// Executes the command.
 	pub(crate) async fn execute(mut self) -> anyhow::Result<()> {
 		Cli.intro("Deploy a smart contract")?;
-
-		let project_path = get_project_path(self.path.clone(), self.path_pos.clone());
+		// Show warning if specified as deprecated.
+		if !self.valid {
+			Cli.warning("DEPRECATION: Please use `pop up` (or simply `pop u`) in future...")?;
```
Daanvdplas, line 101: Please use `pop up` (or simply `pop u`) in the future..."

crates/pop-cli/src/commands/up/mod.rs
```diff
@@ -1,29 +1,143 @@
...
+	#[clap(alias = "n")]
+	Network(network::ZombienetCommand),
+	#[cfg(feature = "parachain")]
+	/// [DEPRECATED] Launch a local network (will be removed in v0.8.0).
+	#[clap(alias = "p", hide = true)]
+	Parachain(network::ZombienetCommand),
 	#[cfg(feature = "contract")]
-	/// Deploy a smart contract.
-	#[clap(alias = "c")]
+	/// [DEPRECATED] Deploy a smart contract (will be removed in v0.8.0).
```
Daanvdplas, line 48: Inconsistent with using `TODO` above. I prefer this style of remembering the team to remove it, public docs or not.

crates/pop-cli/src/commands/up/network.rs
```diff
@@ -48,6 +48,9 @@ pub(crate) struct ZombienetCommand {
 	/// Automatically source all needed binaries required without prompting for confirmation.
 	#[clap(short = 'y', long)]
 	skip_confirm: bool,
+	// Deprecation flag, used to specify whether the deprecation warning is shown.
+	#[clap(skip)]
+	pub(crate) valid: bool,
```
Daanvdplas, line 53: Should also be removed in `v0.8`

crates/pop-cli/src/commands/up/network.rs
```diff
@@ -57,6 +60,13 @@ impl ZombienetCommand {
 		intro(format!("{}: Launch a local network", style(" Pop CLI ").black().on_magenta()))?;
 		set_theme(Theme);
 
+		// Show warning if specified as deprecated.
+		if !self.valid {
+			log::warning(
+				"DEPRECATION: Please use `pop up network` (or simply `pop up n`) in future...",
```
Daanvdplas, line 66: ".. the future"

crates/pop-cli/tests/contract.rs
```diff
@@ -180,7 +171,7 @@ async fn contract_lifecycle() -> Result<()> {
 		.assert()
 		.success();
 
-	// pop up contract --upload-only --use-wallet
+	// pop up --upload-only --use-wallet
```
Daanvdplas, line 174: // `pop up --upload-only --use-wallet`

crates/pop-cli/tests/contract.rs
```diff
@@ -89,19 +89,11 @@ async fn contract_lifecycle() -> Result<()> {
 	sleep(Duration::from_secs(5)).await;
 
 	// Only upload the contract
-	// pop up contract --path ./test_contract --upload-only
+	// pop up --path ./test_contract --upload-only
```
Daanvdplas, line 92: // `pop up --path ./test_contract --upload-only`

crates/pop-cli/src/commands/up/mod.rs
```diff
@@ -1,29 +1,143 @@
...
-/// Arguments for launching or deploying.
-#[derive(Args)]
+/// Arguments for launching or deploying a project.
+#[derive(Args, Clone)]
 #[command(args_conflicts_with_subcommands = true)]
 pub(crate) struct UpArgs {
+	/// Path to the project directory.
+	// TODO: Introduce the short option in v0.8.0 once deprecated parachain command is removed.
+	#[arg(long, global = true)]
+	pub path: Option<PathBuf>,
```
Daanvdplas, line 22: Why adding this when it will be deprecated?

Daanvdplas (commented):
Looking good, left a few nitpicks

crates/pop-cli/src/commands/up/mod.rs
AlexD10S: The path command is not deprecated. The comment is just with `//` for us, to introduce the short option once this is removed: https://github.com/r0gue-io/pop-cli/blob/main/crates/pop-cli/src/commands/up/mod.rs#L46

crates/pop-cli/tests/contract.rs
AlexD10S: Is not a public comment, just to help understand what the code is doing. Why is this change needed?

crates/pop-cli/src/commands/up/mod.rs
AlexD10S: I prefer this style too, as the `pop build spec` comment: https://github.com/r0gue-io/pop-cli/blob/main/crates/pop-cli/src/commands/build/spec.rs#L142

crates/pop-cli/src/commands/up/network.rs
AlexD10S: Fixed in the main pop deploy PR, in this commit: https://github.com/r0gue-io/pop-cli/pull/404/commits/89d4c23a01858429552ca9820cc17f9f515d3a76

crates/pop-cli/src/commands/mod.rs
AlexD10S: Fixed in the main pop deploy PR, in this commit: https://github.com/r0gue-io/pop-cli/pull/404/commits/89d4c23a01858429552ca9820cc17f9f515d3a76

crates/pop-cli/src/commands/up/contract.rs
AlexD10S: Fixed in the main pop deploy PR, in this commit: https://github.com/r0gue-io/pop-cli/pull/404/commits/89d4c23a01858429552ca9820cc17f9f515d3a76

PR #404, opened 2025-02-09
feat: deploy parachain

AlexD10S:
Add functionality to deploy a parachain:

- [ ] Parse event to retrieve ID when reserving paraID
- [ ] Logic in pop up to reserve paraId, generate specs and register parachain.
- [ ] Use a proxy to the logic above.
- [ ] Build deterministic release.

### DIFF
--- a/crates/pop-cli/src/commands/mod.rs
+++ b/crates/pop-cli/src/commands/mod.rs
@@ -55,9 +55,9 @@ fn about_build() -> &'static str {
 /// Help message for the up command.
 fn about_up() -> &'static str {
 	#[cfg(all(feature = "parachain", feature = "contract"))]
-	return "Deploy a smart contract or launch a local network.";
+	return "Deploy a parachain, deploy a smart contract or launch a local network.";
 	#[cfg(all(feature = "parachain", not(feature = "contract")))]
-	return "Launch a local network.";
+	return "Deploy a parachain or launch a local network.";
 	#[cfg(all(feature = "contract", not(feature = "parachain")))]
 	return "Deploy a smart contract.";
 }
--- a/crates/pop-cli/src/commands/up/mod.rs
+++ b/crates/pop-cli/src/commands/up/mod.rs
@@ -71,7 +71,13 @@ impl Command {
 			cmd.execute().await?;
 			return Ok("contract");
 		}
-		cli.warning("No contract detected. Ensure you are in a valid project directory.")?;
+		if pop_parachains::is_supported(project_path.as_deref())? {
+			cli.warning("Parachain deployment is currently not implemented.")?;
+			return Ok("parachain");
+		}
+		cli.warning(
+			"No contract or parachain detected. Ensure you are in a valid project directory.",
+		)?;
 		Ok("")
 	}
 }
@@ -83,6 +89,7 @@ mod tests {
 	use cli::MockCli;
 	use duct::cmd;
 	use pop_contracts::{mock_build_process, new_environment};
+	use pop_parachains::{instantiate_template_dir, Config, Parachain};
 	use std::env;
 	use url::Url;
 
@@ -122,14 +129,30 @@ mod tests {
 		assert_eq!(Command::execute_project_deployment(args.clone(), &mut cli).await?, "contract");
 		cli.verify()?;
 
+		// Parachain
+		let name = "parachain";
+		let project_path = temp_dir.path().join(name);
+		let config = Config {
+			symbol: "DOT".to_string(),
+			decimals: 18,
+			initial_endowment: "1000000".to_string(),
+		};
+		instantiate_template_dir(&Parachain::Standard, &project_path, None, config)?;
+
+		args.path = Some(project_path);
+		cli = MockCli::new().expect_warning("Parachain deployment is currently not implemented.");
+		assert_eq!(Command::execute_project_deployment(args.clone(), &mut cli).await?, "parachain");
+		cli.verify()?;
+
 		// Rust project
 		let name = "hello_world";
 		let path = temp_dir.path();
 		let project_path = path.join(name);
 		cmd("cargo", ["new", name, "--bin"]).dir(&path).run()?;
 		args.path = Some(project_path);
-		cli = MockCli::new()
-			.expect_warning("No contract detected. Ensure you are in a valid project directory.");
+		cli = MockCli::new().expect_warning(
+			"No contract or parachain detected. Ensure you are in a valid project directory.",
+		);
 		assert_eq!(Command::execute_project_deployment(args, &mut cli).await?, "");
 		cli.verify()
 	}


crates/pop-cli/src/commands/mod.rs
```diff
@@ -52,12 +52,12 @@ fn about_build() -> &'static str {
...
 
-/// Help message for the up command.
+/// Help message for the `up` command.
 fn about_up() -> &'static str {
 	#[cfg(all(feature = "parachain", feature = "contract"))]
-	return "Deploy a smart contract or launch a local network.";
+	return "Deploy a parachain, deploy a smart contract or launch a local network.";
 	#[cfg(all(feature = "parachain", not(feature = "contract")))]
-	return "Launch a local network.";
+	return "Deploy a parachain or launch a local network.";
```
evilrobot-01, line 60:
```suggestion
	return "Deploy a rollup or launch a local network.";
```

crates/pop-cli/src/commands/up/mod.rs
```diff
@@ -71,7 +71,14 @@ impl Command {
 			cmd.execute().await?;
 			return Ok("contract");
 		}
-		cli.warning("No contract detected. Ensure you are in a valid project directory.")?;
+		#[cfg(feature = "parachain")]
+		if pop_parachains::is_supported(project_path.as_deref())? {
+			cli.warning("Parachain deployment is currently not implemented.")?;
```
evilrobot-01, line 76:
```suggestion
			cli.warning("Rollup deployment is currently not implemented.")?;
```

crates/pop-cli/src/commands/up/mod.rs
```diff
@@ -71,7 +71,14 @@ impl Command {
...
+			return Ok("parachain");
+		}
+		cli.warning(
+			"No contract or parachain detected. Ensure you are in a valid project directory.",
```
evilrobot-01, line 80:
```suggestion
			"No contract or rollup detected. Ensure you are in a valid project directory.",
```

crates/pop-cli/src/commands/up/mod.rs
```diff
@@ -71,7 +71,14 @@ impl Command {
 			cmd.execute().await?;
 			return Ok("contract");
 		}
-		cli.warning("No contract detected. Ensure you are in a valid project directory.")?;
+		#[cfg(feature = "parachain")]
+		if pop_parachains::is_supported(project_path.as_deref())? {
+			cli.warning("Parachain deployment is currently not implemented.")?;
+			return Ok("parachain");
```
evilrobot-01, line 77:
```suggestion
			return Ok(Project::Rollup);
```

Why are we using strings when this is exactly what an enum is for? An enum gives type safety and is more efficient. If you need a string version, get it from the enum variant.

crates/pop-cli/src/commands/up/mod.rs
```diff
@@ -126,6 +134,25 @@ mod tests {
 		cli.verify()
 	}
 
+	#[tokio::test]
+	async fn detects_parachain_correctly() -> anyhow::Result<()> {
```
evilrobot-01, line 138:
```suggestion
	async fn detects_rollup_correctly() -> anyhow::Result<()> {
```

crates/pop-cli/src/commands/up/mod.rs
```diff
@@ -126,6 +134,25 @@ mod tests {
...
+		let temp_dir = tempfile::tempdir()?;
+		let name = "parachain";
+		let project_path = temp_dir.path().join(name);
+		let config = Config {
+			symbol: "DOT".to_string(),
+			decimals: 18,
+			initial_endowment: "1000000".to_string(),
+		};
+		instantiate_template_dir(&Parachain::Standard, &project_path, None, config)?;
```
evilrobot-01, line 147:
```suggestion
		instantiate_template_dir(&Rollup::Standard, &project_path, None, config)?;
```

crates/pop-cli/src/commands/up/mod.rs
```diff
@@ -126,6 +134,25 @@ mod tests {
 		cli.verify()
 	}
 
+	#[tokio::test]
+	async fn detects_parachain_correctly() -> anyhow::Result<()> {
+		let temp_dir = tempfile::tempdir()?;
+		let name = "parachain";
```
evilrobot-01, line 140:
```suggestion
		let name = "rollup";
```

crates/pop-cli/src/commands/up/mod.rs
```diff
@@ -126,6 +134,25 @@ mod tests {
...
+		let config = Config {
+			symbol: "DOT".to_string(),
+			decimals: 18,
+			initial_endowment: "1000000".to_string(),
+		};
+		instantiate_template_dir(&Parachain::Standard, &project_path, None, config)?;
+
+		let args = create_up_args(project_path)?;
+		let mut cli =
+			MockCli::new().expect_warning("Parachain deployment is currently not implemented.");
```
evilrobot-01, line 151:
```suggestion
			MockCli::new().expect_warning("Rollup deployment is currently not implemented.");
```

crates/pop-cli/src/commands/up/mod.rs
```diff
@@ -126,6 +134,25 @@ mod tests {
...
+		assert_eq!(Command::execute_project_deployment(args, &mut cli).await?, "parachain");
```
evilrobot-01, line 152:
```suggestion
		assert_eq!(Command::execute_project_deployment(args, &mut cli).await?, Project::Rollup);
```

crates/pop-cli/src/commands/up/mod.rs
```diff
@@ -135,8 +162,9 @@ mod tests {
 		let args = create_up_args(project_path)?;
 
 		cmd("cargo", ["new", name, "--bin"]).dir(&path).run()?;
-		let mut cli = MockCli::new()
-			.expect_warning("No contract detected. Ensure you are in a valid project directory.");
+		let mut cli = MockCli::new().expect_warning(
+			"No contract or parachain detected. Ensure you are in a valid project directory.",
```
evilrobot-01, line 166:
```suggestion
			"No contract or rollup detected. Ensure you are in a valid project directory.",
```

crates/pop-cli/src/commands/up/network.rs
```diff
@@ -63,7 +64,7 @@ impl ZombienetCommand {
 		// Show warning if specified as deprecated.
 		if !self.valid {
 			log::warning(
-				"DEPRECATION: Please use `pop up network` (or simply `pop up n`) in future...",
+				"DEPRECATION: Please use `pop up network` (or simply `pop up n`) in the future...",
```
evilrobot-01, line 67:
```suggestion
				"DEPRECATION: Please use `pop up network` (or simply `pop u n`) in the future...",
```

evilrobot-01 (commented):
Comments only as still a draft. Suggestions around rollup are simply based on the fact that we have to start somewhere. I think any new code should use rollup. We can then create a separate PR to add aliases etc and then simply rebase this PR as/when. 

Also surprised you are using strings over enums - that should be addressed imo.

A few other general observations:
- Not sure we need `network` if we are specifying `-f` ? Counter argument is if we are using `network` then we dont need to specify `-f`, it could just be `pop u n ./network.toml`. Or if we rename `-f` to `-n`, then `pop up -n ./network.toml`
- `cargo run -- up --help` shows a 'smart contract deployment options' section but nothing for rollups?
- we could still have the subcommands (contract, rollup, network), just optional. These can just be command aliases that map into the existing functions for `pop up`. Help text could be taken from the subcommand when generating the main help for pop up. I cant say what the technical complexities might be tho.
- `pop up --path ../test-chain --relay-url ws://127.0.0.1:49155` in the readme doesnt work
- readme should be updated to rollup

crates/pop-cli/src/commands/mod.rs
```diff
@@ -52,12 +52,12 @@ fn about_build() -> &'static str {
 	return "Build a smart contract or Rust package.";
 }
 
-/// Help message for the up command.
+/// Help message for the `up` command.
 fn about_up() -> &'static str {
 	#[cfg(all(feature = "parachain", feature = "contract"))]
-	return "Deploy a smart contract or launch a local network.";
+	return "Deploy a parachain, deploy a smart contract or launch a local network.";
```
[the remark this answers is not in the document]
AlexD10S: https://github.com/r0gue-io/pop-cli/pull/404/commits/154bc4f036687d2e69d90550fe3648d2f2f73b3d

crates/pop-cli/src/commands/mod.rs
AlexD10S: https://github.com/r0gue-io/pop-cli/pull/404/commits/154bc4f036687d2e69d90550fe3648d2f2f73b3d

crates/pop-cli/src/commands/up/network.rs
AlexD10S: https://github.com/r0gue-io/pop-cli/pull/404/commits/b57fdd1cdfa09ab4fe3bc3edb1e259e6754cd58d

crates/pop-cli/src/commands/up/mod.rs
AlexD10S: https://github.com/r0gue-io/pop-cli/pull/404/commits/b57fdd1cdfa09ab4fe3bc3edb1e259e6754cd58d

crates/pop-cli/src/commands/up/mod.rs
AlexD10S: https://github.com/r0gue-io/pop-cli/pull/404/commits/b57fdd1cdfa09ab4fe3bc3edb1e259e6754cd58d

crates/pop-cli/src/commands/up/mod.rs
AlexD10S: https://github.com/r0gue-io/pop-cli/pull/404/commits/b57fdd1cdfa09ab4fe3bc3edb1e259e6754cd58d

crates/pop-cli/src/commands/up/mod.rs
AlexD10S: https://github.com/r0gue-io/pop-cli/pull/404/commits/b57fdd1cdfa09ab4fe3bc3edb1e259e6754cd58d

crates/pop-cli/src/commands/up/mod.rs
AlexD10S: The struct rename will be addressed once the naming is updated across all commands (e.g., pop new ...). Adding an alias here is possible but unnecessary at this stage.

crates/pop-cli/src/commands/up/mod.rs
AlexD10S: https://github.com/r0gue-io/pop-cli/commit/b57fdd1cdfa09ab4fe3bc3edb1e259e6754cd58d

crates/pop-cli/src/commands/up/mod.rs
AlexD10S: https://github.com/r0gue-io/pop-cli/commit/b57fdd1cdfa09ab4fe3bc3edb1e259e6754cd58d

crates/pop-cli/src/commands/up/mod.rs
AlexD10S: https://github.com/r0gue-io/pop-cli/commit/b57fdd1cdfa09ab4fe3bc3edb1e259e6754cd58d

crates/pop-cli/src/commands/up/mod.rs
AlexD10S: This feels out of scope here since the entire codebase follows this approach. However, I agree with this change, and introducing this struct will help consolidate all telemetry options in one place. I've created a task and will handle it separately.

AlexD10S:
> * Not sure we need `network` if we are specifying `-f` ? Counter argument is if we are using `network` then we dont need to specify `-f`, it could just be `pop u n ./network.toml`. Or if we rename `-f` to `-n`, then `pop up -n ./network.toml`

In my opinion, this approach makes it much clearer for the user: `pop up` deploys the project, while the `network` subcommand is used specifically for launching a local network.

We follow the same pattern with `pop build`, it builds the project, and for more specific logic like building specs, we use the `spec` subcommand.
That said, I agree that the `-f` flag can be avoided.

chungquantin (approved):
Amazing work, Alex! I tested all functionalities of this PR and it works well. Just one feedback that the experience with building deterministic runtime is not clear to the user. Even though there is a spinner, it's hard to see. I would recommend informing the user that the step might take around 10-15 minutes so they are aware of it. 

✅ Reserve a parachain ID with `pop up --path . --relay-chain-url ws://127.0.0.1:50295`
✅ Deploy a parachain and can find the parathread on PJS

Screenshot 2025-03-14 at 22 18 17



Events are printed beautifully!

crates/pop-cli/src/common/chain.rs
```diff
@@ -0,0 +1,60 @@
...
+		},
+	};
+
+	// Parse metadata from chain url.
+	let client = set_up_client(url.as_str()).await?;
+	let mut pallets = parse_chain_metadata(&client)
+		.map_err(|e| anyhow!(format!("Unable to fetch the chain metadata: {}", e.to_string())))?;
+	// Sort by name for display.
+	pallets.sort_by(|a, b| a.name.cmp(&b.name));
+	pallets.iter_mut().for_each(|p| p.functions.sort_by(|a, b| a.name.cmp(&b.name)));
```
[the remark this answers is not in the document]
AlexD10S: https://github.com/r0gue-io/pop-cli/pull/404/commits/3d5a4154975bb97d072c4e634be10ce83e4fc994

crates/pop-cli/src/common/chain.rs
```diff
@@ -0,0 +1,60 @@
...
+			Url::parse(&url)?
+		},
+	};
+
+	// Parse metadata from chain url.
+	let client = set_up_client(url.as_str()).await?;
+	let mut pallets = parse_chain_metadata(&client)
+		.map_err(|e| anyhow!(format!("Unable to fetch the chain metadata: {}", e.to_string())))?;
+	// Sort by name for display.
+	pallets.sort_by(|a, b| a.name.cmp(&b.name));
```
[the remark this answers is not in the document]
AlexD10S: https://github.com/r0gue-io/pop-cli/pull/404/commits/3d5a4154975bb97d072c4e634be10ce83e4fc994

crates/pop-cli/src/commands/up/rollup.rs
```diff
@@ -0,0 +1,441 @@
...
+		match config.register(cli).await {
+			Ok(_) => cli.success(format!(
+				"Deployment successfully {}",
+				style(format!(
+					"https://polkadot.js.org/apps/?rpc={}#/parachains",
+					config.chain.url
+				))
+				.dim()
+			))?,
+			Err(e) => cli.outro_cancel(format!("{}", e))?,
```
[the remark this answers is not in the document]
AlexD10S: https://github.com/r0gue-io/pop-cli/pull/404/commits/3d5a4154975bb97d072c4e634be10ce83e4fc994

crates/pop-cli/src/commands/up/rollup.rs
```diff
@@ -0,0 +1,441 @@
...
+}
+
+impl UpCommand {
+	/// Executes the command.
+	pub(crate) async fn execute(self, cli: &mut impl Cli) -> Result<()> {
+		cli.intro("Deploy a rollup")?;
+		let config = match self.prepare_for_registration(cli).await {
+			Ok(chain) => chain,
+			Err(e) => {
+				cli.outro_cancel(format!("{}", e))?;
```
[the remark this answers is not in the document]
AlexD10S: https://github.com/r0gue-io/pop-cli/pull/404/commits/3d5a4154975bb97d072c4e634be10ce83e4fc994

crates/pop-cli/src/commands/up/mod.rs
```diff
@@ -25,6 +27,10 @@ pub(crate) struct UpArgs {
 	#[arg(value_name = "PATH", index = 1, global = true, conflicts_with = "path")]
 	pub path_pos: Option<PathBuf>,
 
+	#[command(flatten)]
+	#[cfg(feature = "parachain")]
```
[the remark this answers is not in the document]
AlexD10S: Yes

crates/pop-cli/src/commands/mod.rs
```diff
@@ -52,12 +52,12 @@ fn about_build() -> &'static str {
 	return "Build a smart contract or Rust package.";
 }
 
-/// Help message for the up command.
+/// Help message for the `up` command.
 fn about_up() -> &'static str {
 	#[cfg(all(feature = "parachain", feature = "contract"))]
-	return "Deploy a smart contract or launch a local network.";
+	return "Deploy a rollup(parachain), deploy a smart contract or launch a local network.";
```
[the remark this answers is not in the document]
AlexD10S: This was discussed elsewhere. We are going to rename it to 'rollup' in the rest of the codebase and will start introducing 'rollup' in this PR.

chungquantin (approved):
Registered with proxy and build deterministic runtime works perfectly as well. One final question, if my runtime is built deterministically with same configurations before, will it build the runtime again? Because when I run the command again after the built finish, it still take around 10-15 minutes to build the runtime. Could we no reuse the built runtime binary?

AlexD10S:
> Registered with proxy and build deterministic runtime works perfectly as well. One final question, if my runtime is built deterministically with same configurations before, will it build the runtime again? Because when I run the command again after the built finish, it still take around 10-15 minutes to build the runtime. Could we no reuse the built runtime binary?

It needs to be rebuilt as there may be changes, similar to when we build the node. However, I believe the second run takes less time.

chungquantin:
> > Registered with proxy and build deterministic runtime works perfectly as well. One final question, if my runtime is built deterministically with same configurations before, will it build the runtime again? Because when I run the command again after the built finish, it still take around 10-15 minutes to build the runtime. Could we no reuse the built runtime binary?
> 
> It needs to be rebuilt as there may be changes, similar to when we build the node. However, I believe the second run takes less time.

Can we add an option to ask user if they want to rebuild their runtime? Or we can follow similar approach to the benchmarking feature that if binary exists, we don't build it. User can build manually if needed.

AlexD10S:
> > > Registered with proxy and build deterministic runtime works perfectly as well. One final question, if my runtime is built deterministically with same configurations before, will it build the runtime again? Because when I run the command again after the built finish, it still take around 10-15 minutes to build the runtime. Could we no reuse the built runtime binary?
> > 
> > 
> > It needs to be rebuilt as there may be changes, similar to when we build the node. However, I believe the second run takes less time.
> 
> Can we add an option to ask user if they want to rebuild their runtime? Or we can follow similar approach to the benchmarking feature that if binary exists, we don't build it. User can build manually if needed.

Interesting. It doesn’t make sense to ask for it after prompting 'Would you like to build the runtime deterministically?'. However, it might be a good idea to detect if the chain spec file exists after the prompt 'Name or path for the plain chain spec file:' and then ask the user if they want to rebuild it. In https://github.com/r0gue-io/pop-cli/pull/459 we implemented  the logic to prevent unnecessary chain spec rebuilds.

PR #406, opened 2025-02-10
docs: `profile` comment in `build_parachain`

chungquantin:
Minor fix for the parameter type of `build_parachain`


AlexD10S (approved):
Nice catch!

PR #409, opened 2025-02-11
fix: increase `DefaultBodyLimit` to prevent large payload failures

AlexD10S:
When attempting to register a parachain using the wallet to sign the transaction, an error appeared in the frontend:
```
Error: Unable to submit. Is the server closed?
```
This was caused by the request body limit being too low, resulting in a `413 (Payload Too Large)` error.

This PR increases DefaultBodyLimit to handle larger payloads and adds a test to verify large transaction support. The 15MB limit is a bit arbitrary but ensures stability, providing enough room for 5MB transactions while accounting for potential encoding overhead.

Resource: https://stackoverflow.com/questions/75205271/rust-axum-multipart-length-limit-exceeded.

### DIFF
--- a/crates/pop-cli/src/wallet_integration.rs
+++ b/crates/pop-cli/src/wallet_integration.rs
@@ -1,4 +1,5 @@
 use axum::{
+	extract::DefaultBodyLimit,
 	http::HeaderValue,
 	response::Html,
 	routing::{get, post},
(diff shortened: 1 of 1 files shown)


peterwht (approved):
Looks good! Nice fix.

If you have the time, might be easy enough to add a failing test that is 15.1 MB to trigger the error. Not a blocker though

AlexD10S:
> Looks good! Nice fix.
> 
> If you have the time, might be easy enough to add a failing test that is 15.1 MB to trigger the error. Not a blocker though

Make sense to test that. Added in the commit https://github.com/r0gue-io/pop-cli/pull/409/commits/7d7972061447b40cf677d6d9424687fa6f82c2ba

peterwht (approved):

PR #411, opened 2025-02-12
feat: benchmark existing runtime binary and select policy, presets

chungquantin:
Screenshot 2025-02-18 at 21 18 09

# Scenarios
Context: The `pop bench pallet` command is run at the root directory of the parachain project

## Stage 1: Locating and building a `runtime` binary

- Scenario 1: If `runtime_path` (Path to the runtime, hard-coded to "runtime" folder) is not a Rust project, list all directories in the `runtime_path` (similar to `pop-node` which has multiple runtimes inside the `runtime` folder). After user select the runtime, we check if the runtime binary exists or not. If not, we build the runtime with `runtime-benchmarks` feature. 

```
list all directories -> select runtime -> check if binary exists -> build runtime with `runtime-benchmarks` feature.
```

- Scenario 2: If `runtime_path` is a Rust project, located inside the `runtime` folder, which means it is a runtime codebase (Note: There could be a chance that this is a workspace?). We will throw `No runtime found` if the workspace is not valid. We check if the runtime binary exists or not. If not, we build the runtime with `runtime-benchmarks` feature. 

```
check if binary exists -> build runtime with `runtime-benchmarks` feature.
```

- Scenario 3: If no `runtime` found, we throws `No runtime found.` (See diagram). 

## Stage 2: Select genesis builder policy and input the preset. 

After the runtime binary is ready, we prompts user to select the genesis builder (There are two options currently: `none` and `runtime`, [more details can be found here](https://github.com/paritytech/polkadot-sdk/issues/5694)). As we only support the runtime, no chain spec is provided so other genesis builder policies are not valid.

- `none` is selected: We do nothing.
- `runtime` is selected: [The runtime must be configured with `sp-genesis-builder`](https://github.com/paritytech/polkadot-sdk/pull/7476/files#diff-43828bcd5e38b4c5384e0968bca08c018b9f02edda5830bb500a91d817ee5c69)). We prompt the user to select the available genesis builder presets (presets will be registered on the runtime using `BuildGenesisConfig` runtime API). 

By calling [sc-chain-spec::GenesisConfigBuilderRuntimeCaller](https://docs.rs/sc-chain-spec/latest/sc_chain_spec/struct.GenesisConfigBuilderRuntimeCaller.html) `get_preset_names` method, we list all genesis builder presets and expect the user to select one of them. If there is no builder preset found, the CLI throws `No preset found` (can be reproduced with `pop-node`). 

- If `--runtime=` is provided, we don't locate the runtime. 
[...]

PR #420, opened 2025-02-19
feat: pallet benchmarking interactive UI

chungquantin:
# Description

The PR is a part of a bigger PR #424. It implements the full flow of the `pop bench pallet` including new key features: 
- Interactive UI to manage parameters before running the benchmark. 
- Sourcing the `frame-omni-bencher` binary #428. This is required to list pallets and extrinsics. [Explained here](https://github.com/r0gue-io/pop-cli/pull/420#discussion_r1974726796)
- Refactor to use our own version of `PalletCmd` (called `BenchmarkPallet`) instead of `frame-benchmarking-cli::PalletCmd` due to required parameters issue. [Explained here](https://github.com/r0gue-io/pop-cli/pull/420#discussion_r1974742591)

## How to test the PR? 

```
pop bench pallet
```

If arguments are provided like 

```
pop bench pallet --runtime=path-to-runtime --pallet= --extrinsic=
```

We skip the runtime, the pallet and the extrinsic.

To skip the parameter menu, user can specify flag `--skip` then the parameter menu can be skipped and go directly to run the bencmarking. 

```
pop bench pallet --skip
```

Test single line command
```
pop bench pallet --pallet=pallet_balances --extrinsic=burn_allow_death,burn_keep_alive --steps=50 --repeat=20 --external-repeat=1 --db-cache=1024 --map-size=1000000 --additional-trie-layers=2 --output-pov-analysis=median-slopes --runtime=/Users/chungquantin/Developer/pop-network/pop-cli/tests/runtimes/base_parachain_benchmark.wasm --genesis-builder=runtime --genesis-builder-preset=development
```

## Help command

Screenshot 2025-03-05 at 11 56 16

### Full flow breakdown

Note: If `--list` or `--json-output` is provided, we run the command by `frame-benchmarking-cli` directly instead of go through the implementation in pop-cli. 

#### Stage 1: Locate runtime or input runtime path manually
- 1.1. Run the command in a parachain project, which contains `runtime` or `runtimes` (Hyperbridge) folder. This falls to the first stage [**_"Stage 1: Locating and building a runtime binary | Scenario 1"_** of the previous PR](https://github.com/r0gue-io/pop-cli/pull/411). 

Screenshot 2025-02-28 at 17 05 25

- 1.2. Run the command anywhere (no a parachain project), which does contains `runtime` or `runtimes` (Hyperbridge) folder. This falls to the first stage [**_"Stage 1: Locating and building a runtime binary | Scenario 2"_** of the previous PR](https://github.com/r0gue-io/pop-cli/pull/411). 

Screenshot 2025-02-28 at 17 03 33

- 1.3. Inputting the runtime manually. User can input the path to parachain project or path to WASM binary file directly.

NOTE: After the path is located and there is no WASM file found in the `target/<profile>/wbuild/<runtime-name>.wasm`, we will build the runtime with `runtime-benchmarks` feature first.

### Stage 2: Select genesis builder policy & preset

See full breakdown in #411 

### Stage 3: Select pallets (Single Select) & extrinsics (Multiple Select)

At this stage, we will check if the binary of `frame-omni-bencher` exists on the user's machine. [Explained here](https://github.com/r0gue-io/pop-cli/pull/420#discussion_r1974726796)

Note: We only call `frame-omni-bencher` to load pallets and extrinsics once a session. 

- 3.1. **_"🔎 Search for pallets by name"_**: User can specify the pallet name to search for the pallets (fuzzy searching).

For the extrinsic selection, we only allow user to select if there are pallet specified. If all pallet are selected (`--pallet=*`), we skip the next step. 

- 3.2. **_"🔎 Search for extrinsics by name"_**: User can specify the extrinsic name to search for the extrinsics (fuzzy searching). 

<div style="display:flex;">
Screenshot 2025-03-05 at 11 50 10
Screenshot 2025-03-05 at 11 51 05
</div>

### Stage 4: Parameter menu

User can update the parameters by selecting an option in the parameter menu. Some arguments will be disabled depend on the status of other parameters. See `is_disabled` method. For example: 
- we don't allow selecting extrinsics if all pallets are selected.
- we don't allow update the genesis builder preset if policy is equal to `GenesisBuilderPolicy::None`

Screenshot 2025-02-28 at 17 15 29

### Stage 5: Input the output path

This stage is optional, if no value is provided, we don't write to any file. 

Screenshot 2025-02-28 at 17 19 30

### Stage 6: Print the single line command 

Screenshot 2025-02-28 at 17 20 08

[sc-2882]


crates/pop-parachains/src/bench.rs
```diff
@@ -33,7 +53,143 @@ pub fn get_runtime_path(parent: &Path) -> anyhow::Result<PathBuf> {
...
 		.find(|path| path.exists())
-		.ok_or_else(|| anyhow::anyhow!("No runtime found."))
+		.ok_or_else(|| anyhow::anyhow!("No runtime found"))
+}
+
+/// Loads a mapping of pallets and their associated extrinsics from the runtime WASM binary.
+///
+/// # Arguments
+/// * `runtime_path` - Path to the runtime WASM binary.
+pub fn load_pallet_extrinsics(runtime_path: &Path) -> anyhow::Result<PalletExtrinsicsRegistry> {
```
chungquantin, line 63: - `Approach 1 (Subwasm + FRAME metadata)`: Current approach of this PR: to retrieve the metadata from Substrate runtime binary is to use `ac-node-api` of `substrate-api-client` and `subwasm`'s WASM loader crate to load the pallets. However, there are issues that the pallet name is represented in the defined name on the runtime level, not the crate name (e.g. `Timestamp` instead of `pallet_timestamp`), hence, when pass it to the `--pallet=` argument, the `frame-benchmarking-cli` can't identify the pallet to benchmark. On the other hand, `pallet::hooks` are not included in the Substrate runtime metadata. But hooks are also benchmarked, so we need to find a way to get them. 

- `Approach 2 (List pallets using frame-benchmarking-cli)`: See branch `chungquantin/feat-pallet_overwrite_stdio`. This approach works perfectly and it was the original approach. This approach is we need to overwrite the stdout. So that when `println!` is called in `frame-benchmarking-cli`, the output is captured and returned back for processing. We use external crate https://github.com/elichai/stdio-override to redirect the stdout. However, the issue happens on the CI when there's a conflict with the stdout. Code to the issue: https://github.com/elichai/stdio-override/blob/master/src/lib.rs#L151

```
thread 'bench::tests::search_extrinsics_works' panicked at /home/runner/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/stdio-override-0.2.0/src/lib.rs:236:13:
Stdout override reset out of order!
```

crates/pop-parachains/src/bench.rs
```diff
@@ -33,7 +53,143 @@ pub fn get_runtime_path(parent: &Path) -> anyhow::Result<PathBuf> {
...
+		}
+	}
+	Ok(registry)
+}
+
+/// Print the pallet benchmarking command with arguments.
+///
+/// # Arguments
+/// * `cmd` - Command to benchmarking extrinsic weights of FRAME pallets.
+pub fn print_pallet_command(cmd: &PalletCmd) -> String {
```
chungquantin, line 86: There are arguments which are not covered yet:
- `wasm_method` due to private type `WasmExecutionMethod`
- `wasm_instantiation_strategy` due to private type `WasmtimeInstantiationStrategy` 
- `default_pov_mode` due to private type `PovEstimationMode`
- `hostinfo_params` due to private type `HostInfoParams`

crates/pop-cli/src/commands/bench/pallet.rs
```diff
@@ -51,78 +80,513 @@ impl BenchmarkPallet {
...
+		if let Err(e) = result {
 			return display_message(&e.to_string(), false, cli);
 		}
 		display_message("Benchmark completed successfully!", true, cli)?;
 		Ok(())
 	}
 }
 
+#[derive(Clone, Copy, EnumIter, EnumMessageDerive, Eq, PartialEq)]
+enum BenchmarkPalletMenuOption {
```
chungquantin, line 129: Not all arguments are listed in the menu. 
- `wasm_method` due to private type `WasmExecutionMethod`
- `wasm_instantiation_strategy` due to private type `WasmtimeInstantiationStrategy` 
- `default_pov_mode` due to private type `PovEstimationMode`
- `hostinfo_params` due to private type `HostInfoParams`

crates/pop-parachains/src/bench.rs
chungquantin: Update: Follow the second approach but no longer overwrite the `stdout`, the main issue is with the CI. CI environment spawns their own child processes and overwrite the stdout with their own buffer. Hence, it causes the implementation fails. On the other hand, overwrite stdout might not be a stable approach for cross platforms. 

Solution to that is to use the `frame-omni-bencher` and source it from the remote repository, implementation in #428

crates/pop-cli/src/commands/bench/pallet.rs
```diff
@@ -51,78 +74,513 @@ impl BenchmarkPallet {
...
+
+fn fetch_pallet_registry(
+	cmd: &PalletCmd,
+	registry: &mut PalletExtrinsicsRegistry,
+) -> anyhow::Result<()> {
+	if registry.is_empty() {
+		let spinner = spinner();
+		spinner.start("Fetching pallets and extrinsics from your runtime...");
+		let runtime_path = get_runtime_argument(cmd)?;
+		log::set_max_level(LevelFilter::Off);
```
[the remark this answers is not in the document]
chungquantin: No longer needed

crates/pop-parachains/src/bench/mod.rs
```diff
@@ -0,0 +1,352 @@
...
+/// # Arguments
+/// * `runtime_path` - Path to the runtime WASM binary.
+/// * `binary_path` - Path to the binary of FRAME Omni Bencher.
+pub async fn load_pallet_extrinsics(
+	runtime_path: &Path,
+	binary_path: &Path,
+) -> anyhow::Result<PalletExtrinsicsRegistry> {
+	let temp_file = NamedTempFile::new()?;
+
+	let mut process = run_benchmarking_with_binary(
```
chungquantin, line 110: The reason why we need to download the binary.

crates/pop-parachains/src/bench/mod.rs
```diff
@@ -0,0 +1,352 @@
...
+}
+
+fn process_pallet_extrinsics(output_file: PathBuf) -> anyhow::Result<PalletExtrinsicsRegistry> {
+	let mut output_file = File::open(output_file)?;
+	let mut output = String::new();
+	output_file.read_to_string(&mut output)?;
+
+	// Process the captured output and return the pallet extrinsics registry.
+	let mut registry = PalletExtrinsicsRegistry::new();
+	let lines: Vec<String> = output.split("\n").map(String::from).skip(1).collect();
```
chungquantin, line 132: `.skip(1)` because first line is `pallet, extrinsic`

crates/pop-parachains/src/bench.rs
chungquantin, line 1: Move all the code to `bench/mod.rs` to keep it more organized with separated logic.

crates/pop-parachains/Cargo.toml
```diff
@@ -37,6 +37,7 @@ zombienet-sdk.workspace = true
 # Benchmarking
 cumulus-primitives-proof-size-hostfunction.workspace = true
 frame-benchmarking-cli.workspace = true
+fuzzy-matcher = "0.3.7"
```
chungquantin, line 40: Add this crate to fuzzy searching the pallets and extrinsics by name

crates/pop-cli/src/commands/bench/pallet.rs
```diff
@@ -1,71 +1,743 @@
...
+	fs,
+	path::{Path, PathBuf},
 };
-use std::{env::current_dir, fs, path::PathBuf};
+use strum::{EnumMessage, IntoEnumIterator};
+use strum_macros::{EnumIter, EnumMessage as EnumMessageDerive};
+
+const ALL_SELECTED: &str = "*";
+const MAX_EXTRINSIC_LIMIT: usize = 15;
+const MAX_PALLET_LIMIT: usize = 20;
```
chungquantin, line 31: To keep the output on terminal clean, I believe we should limit the number of displayed pallets

crates/pop-cli/src/commands/bench/pallet.rs
```diff
@@ -1,71 +1,743 @@
...
+
+#[derive(Args, Default)]
+pub(crate) struct BenchmarkPallet {
```
chungquantin, line 34: We implement our own version of `frame-benchmarking-cli::Pallet` because two important fields `--pallet` and `--extrinsic` are required in the original struct. Hence, the parsing check fails when user call `pop bench command` without those arguments. The expected behaviour is to allow user to select pallets and extrinsics if the arguments are not provided. One solution is to use `ignore_errors = true` provided by `clap-rs` but it would cause an issue with the `help` command. Discussed here: https://github.com/clap-rs/clap/issues/5919

Missing below arguments defined in `frame-benchmarking-cli::PalletCmd`:

### Parameters too advanced and rarely used so I don't include them

- `pub hostinfo_params: HostInfoParams`

Parameters to configure how the host info will be determined.

- `pub default_pov_mode: command::PovEstimationMode`

The PoV estimation mode of a benchmark if no `pov_mode` attribute is present.

### Not important to Pop CLI

- `pub no_csv_header: bool`

Don't include csv header when listing benchmarks.

- `pub execution: Option<String>` (Deprecated)

DEPRECATED: This argument has no effect.

- `pub unsafe_overwrite_results: bool`

Allow overwriting a single file with multiple results.
This exists only to restore legacy behaviour. It should never actually be needed.

- `quiet: bool`

Do not print a summary at the end of the run.

- `disable_proof_recording: bool`

Do not enable proof recording during time benchmarking.

### Consider to add

- `pub wasm_method: WasmExecutionMethod`

Method for executing Wasm runtime code.

- `pub wasmtime_instantiation_strategy: WasmtimeInstantiationStrategy`

The WASM instantiation method to use. Only has an effect when `wasm-execution` is set to `compiled`.

crates/pop-cli/src/commands/bench/pallet.rs
```diff
@@ -1,71 +1,797 @@
...
+	/// Disable verification logic when running benchmarks.
+	#[arg(long)]
+	pub no_verify: bool,
+
+	/// Display and run extra benchmarks that would otherwise not be needed for weight
+	/// construction.
+	#[arg(long)]
+	pub extra: bool,
 
-const GENESIS_BUILDER_NO_POLICY: &str = "none";
```
chungquantin, line 13: As we implement our own `GenesisBuilderPolicy` enum, this is no longer needed.

crates/pop-cli/src/commands/bench/pallet.rs
```diff
@@ -1,71 +1,797 @@
...
+	/// By default, proof recording is enabled during benchmark execution. This can slightly
+	/// inflate the resulting time weights. For parachains using PoV-reclaim, this is typically the
+	/// correct setting. Chains that ignore the proof size dimension of weight (e.g. relay chain,
+	/// solo-chains) can disable proof recording to get more accurate results.
+	#[arg(long)]
+	disable_proof_recording: bool,
+
+	/// If this is set to true, no parameter menu pops up
+	#[arg(long = "skip")]
+	skip_menu: bool,
```
chungquantin, line 188: Additional argument. When provided, parameter menu won't pop up.

crates/pop-cli/src/common/bench.rs
```diff
@@ -0,0 +1,449 @@
+// SPDX-License-Identifier: GPL-3.0
```
AlexD10S, line 1: Not requesting the change but curious why in `common` folder? The methods are only used in the bench functionality

Cargo.toml
```diff
@@ -76,6 +77,7 @@ git2_credentials = "0.13.0"
 # benchmarking
 cumulus-primitives-proof-size-hostfunction = "0.10.0"
 frame-benchmarking-cli = "46.0.0"
+fuzzy-matcher = "0.3.7"
```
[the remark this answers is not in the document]
chungquantin: Seems like the implementation in cliclack looks better. Created a PR here to add the feature to `Cli`: https://github.com/r0gue-io/pop-cli/pull/435

crates/pop-cli/src/common/bench.rs
chungquantin: These methods will be shared by other bench commands like `overhead` (See #432). So I think it makes more sense to gather them in the common folder.

crates/pop-parachains/src/bench/mod.rs
```diff
@@ -0,0 +1,341 @@
+// SPDX-License-Identifier: GPL-3.0
+
+use clap::Parser;
+use frame_benchmarking_cli::PalletCmd;
```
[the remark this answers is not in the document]
chungquantin: For this PR, agree that we don't need it. But for other commands, we don't want to fork the command struct if possible, so the crate `frame-benchmarking-cli` is still required to implement others. Reference: #432

crates/pop-parachains/src/bench/mod.rs
```diff
@@ -0,0 +1,341 @@
...
+#[derive(clap::ValueEnum, Debug, Eq, PartialEq, Clone, Copy, EnumIter, EnumMessageDerive)]
+#[clap(rename_all = "kebab-case")]
+pub enum GenesisBuilderPolicy {
+	/// Do not provide any genesis state.
+	None,
+	/// Let the runtime build the genesis state through its `BuildGenesisConfig` runtime API.
+	Runtime,
+}
+
+impl ToString for GenesisBuilderPolicy {
```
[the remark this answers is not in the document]
chungquantin: Resolved by refactoring to `impl Display`

crates/pop-cli/src/commands/bench/pallet.rs
```diff
@@ -1,334 +1,1251 @@
...
+
 		cli.warning("NOTE: this may take some time...")?;
-		cli.info("Benchmarking and generating weight file...")?;
-		if let Err(e) = run_pallet_benchmarking(cmd) {
+		cli.info("Benchmarking extrinsic weights of selected pallets...")?;
+		let result = self.run();
+
+		// Display the benchmarking command.
+		let mut message = self.display();
+		if self.skip_menu {
```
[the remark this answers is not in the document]
chungquantin: Resolved in https://github.com/r0gue-io/pop-cli/pull/420/commits/36130e724354c32229ea66379bd7fc3fcdee0080

Cargo.toml
```diff
@@ -37,6 +37,7 @@ tempfile = "3.10"
 thiserror = "1.0.58"
 tokio-test = "0.4.4"
 toml = "0.5.0"
+which = "4.4"
```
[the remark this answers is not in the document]
chungquantin: Resolved by removing the `which` crate and implement which with `cmd` instead. https://github.com/r0gue-io/pop-cli/pull/420/commits/112f98272233e17d38eca656047e3e8acc53fc60.

I don't find the method `check_command_exists` matches my needs and I found it simpler to just call the `cmd` directly. Note it here that the logic can be refactored to standalone function for future us to get path of `command`.

crates/pop-cli/src/commands/bench/pallet.rs
```diff
@@ -1,334 +1,1251 @@
...
-
-	let parsed_policy = parse_genesis_builder_policy(policy)?;
-	cmd.genesis_builder = parsed_policy.genesis_builder;
-	if policy == GENESIS_BUILDER_RUNTIME_POLICY {
-		cmd.genesis_builder_preset =
-			guide_user_to_select_genesis_preset(cli, runtime_path, &cmd.genesis_builder_preset)?;
-	}
-	Ok(())
+#[derive(Clone, Copy, EnumIter, EnumMessageDerive, Eq, PartialEq)]
+enum BenchmarkPalletMenuOption {
```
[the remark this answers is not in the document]
chungquantin: Not sure what is the better way to do this. One way is to build a similar experience is having our own macro to construct the menu from struct. But the drawback that it's harder to manage. There are fields in the `BenchmarkPallet` we don't want to display on the menu.

crates/pop-parachains/src/bench/binary.rs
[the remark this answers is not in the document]
AlexD10S: https://github.com/r0gue-io/polkadot/pull/1

crates/pop-parachains/src/bench/mod.rs
```diff
@@ -0,0 +1,341 @@
...
+	collections::HashMap,
+	fs::{self, File},
+	io::Read,
+	path::{Path, PathBuf},
+	process::{Child, Command, Stdio},
+};
+use strum_macros::{EnumIter, EnumMessage as EnumMessageDerive};
+use tempfile::NamedTempFile;
+
+pub mod binary;
```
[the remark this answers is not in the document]
chungquantin: Resolved in https://github.com/r0gue-io/pop-cli/pull/420/commits/b5747fb90b20ac04de1222f3143a8b058f0e034b

chungquantin:
> Well done on sourcing the binary! I wonder if we can now remove the `frame-benchmarking` library now and fix the CI issues.
> 
> I've left some comments, but I still need to do a more in-depth review of `common/bench.rs`, especially `bench/pallet.rs`, particularly the tests. Since the coverage CI isn’t working, the tests require extra attention.
> 
> I also need to do more manual testing because I found errors in the searcher:
> 
> ```
> ▲  NOTE: the `pop bench pallet` is not yet battle tested - double check the results.
> │  
> ◇  Select the runtime:
> │  pop-runtime-testnet 
> │
> ◇  🔎 Search for pallets by name ("*" to select all)
> │  balances
> │
> Error: No items added to the list
> ```

Issue with the pallet search is resolved in https://github.com/r0gue-io/pop-cli/pull/420/commits/6681706b7e326778144e7cd2c49fe213a3df5233. There was a missing `--genesis-builder=none` when running benchmark with binary.

Cargo.toml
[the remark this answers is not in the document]
chungquantin: Refactored the code to support filter mode in the pallet benchmarking: https://github.com/r0gue-io/pop-cli/pull/442

crates/pop-cli/src/commands/bench/pallet.rs
```diff
@@ -1,334 +1,1251 @@
...
-		let manifest = from_path(Some(path.as_path()))?;
-		let package = manifest.package();
-		let name = package.clone().name;
-		let description = package.description().unwrap_or_default().to_string();
-		prompt = prompt.item(path, &name, &description);
+) -> anyhow::Result<String> {
+	// Prompt for pallet search input.
+	let input = cli
+		.input(r#"🔎 Search for pallets by name ("*" to select all)"#)
+		.placeholder("balances")
```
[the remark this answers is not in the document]
chungquantin: https://github.com/r0gue-io/pop-cli/pull/420/commits/97af1173e49fbb2659f24272e4f8ec2eb314e337 removes the ("*" to select all) logic and confirm with the user first before listing pallets

Screenshot 2025-03-05 at 10 52 32

crates/pop-cli/src/common/bench.rs
```diff
@@ -0,0 +1,449 @@
...
+		project_path = project_path.join(runtime);
+	}
+	Ok(project_path)
+}
+
+/// Guide the user to input a runtime path.
+///
+/// # Arguments
+/// * `cli`: Command line interface.
+pub fn guide_user_to_input_runtime_path(cli: &mut impl Cli) -> anyhow::Result<PathBuf> {
```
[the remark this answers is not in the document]
chungquantin: Agree. This method is redundant. Removed.

crates/pop-parachains/src/bench/mod.rs
```diff
@@ -0,0 +1,238 @@
...
+	binary_path: &Path,
+) -> anyhow::Result<PalletExtrinsicsRegistry> {
+	let temp_file = NamedTempFile::new()?;
+
+	let mut process = run_benchmarking_with_binary(
+		binary_path,
+		Some(temp_file.as_file()),
+		vec![
+			&format!("--runtime={}", runtime_path.display()),
+			"--genesis-builder=none",
```
chungquantin, line 115: We need this flag otherwise error thrown in the child process.

crates/pop-common/src/helpers.rs
```diff
@@ -55,6 +55,20 @@ pub fn prefix_with_current_dir_if_needed(path: PathBuf) -> PathBuf {
 	path
 }
 
+/// Returns the relative path from `base` to `full` if `full` is inside `base`.
+/// If `full` is outside `base`, returns the absolute path instead.
+///
+/// # Arguments
+/// * `base` - The base directory to compare against.
+/// * `full` - The full path to be shortened.
+pub fn get_relative_or_absolute_path(base: &Path, full: &Path) -> PathBuf {
```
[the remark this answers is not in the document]
chungquantin: Resolved in https://github.com/r0gue-io/pop-cli/pull/420/commits/7641a8a3fa76ba6cc40a16dff0652a49ec85ed66

crates/pop-parachains/Cargo.toml
chungquantin: Removed this crate in #442

crates/pop-cli/src/common/bench.rs
```diff
@@ -0,0 +1,449 @@
...
+	let input = cli
+		.input("Please provide the path to the runtime or parachain project.")
+		.required(true)
+		.default_input("./runtime")
+		.placeholder("./runtime")
+		.interact()
+		.map(PathBuf::from)
```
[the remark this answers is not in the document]
chungquantin: Resolved!

crates/pop-cli/src/common/bench.rs
```diff
@@ -0,0 +1,449 @@
...
+/// Guide the user to input a runtime path.
+///
+/// # Arguments
+/// * `cli`: Command line interface.
+pub fn guide_user_to_input_runtime_path(cli: &mut impl Cli) -> anyhow::Result<PathBuf> {
+	let input = cli
+		.input("Please provide the path to the runtime or parachain project.")
+		.required(true)
+		.default_input("./runtime")
+		.placeholder("./runtime")
```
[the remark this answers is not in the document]
chungquantin: Resolved!

crates/pop-cli/src/common/bench.rs
```diff
@@ -0,0 +1,456 @@
...
+		*current_preset = guide_user_to_select_genesis_preset(cli, runtime_path, current_preset)?;
+	}
+	Ok(())
+}
+
+/// Guide the user to select a runtime path.
+///
+/// # Arguments
+/// * `cli`: Command line interface.
+/// * `target_path`: A reference to the target path.
```
Daanvdplas, line 175:
```suggestion
/// * `target_path`: The target path.
```

crates/pop-cli/src/common/bench.rs
```diff
@@ -0,0 +1,456 @@
...
+) -> anyhow::Result<PathBuf> {
+	let mut project_path = match get_runtime_path(target_path) {
+		Ok(path) => path,
+		Err(_) => {
+			cli.warning(format!(
+				"No runtime folder found at {}. Please input the runtime path manually.",
+				target_path.display()
+			))?;
+			let input: PathBuf = cli
+				.input("Please provide the path to the runtime or parachain project.")
```
Daanvdplas, line 188:
```suggestion
				.input("Please provide the path to the runtime.")
```

crates/pop-cli/src/common/bench.rs
```diff
@@ -0,0 +1,456 @@
...
+				.required(true)
+				.default_input(DEFAULT_RUNTIME_DIR)
+				.placeholder(DEFAULT_RUNTIME_DIR)
+				.interact()?
+				.into();
+			input.canonicalize()?
+		},
+	};
+
+	// If there is no TOML file exist, list all directories in the "runtime" folder and prompt the
```
Daanvdplas, line 198:
```suggestion
	// If a TOML file does not exist, list all directories in the "runtime" folder and prompt the
```

crates/pop-cli/src/common/mod.rs
```diff
@@ -1,5 +1,7 @@
 // SPDX-License-Identifier: GPL-3.0
 
+#[cfg(feature = "parachain")]
+pub mod bench;
```
Daanvdplas, line 4: Public docs

crates/pop-parachains/src/bench/mod.rs
```diff
@@ -0,0 +1,238 @@
...
+///
+/// (Recommended for testing with a single node, e.g., for benchmarking)
+pub const GENESIS_BUILDER_DEV_PRESET: &str = "development";
+
+type HostFunctions = (
+	sp_statement_store::runtime_api::HostFunctions,
+	cumulus_primitives_proof_size_hostfunction::storage_proof_size::HostFunctions,
+);
+
+/// Type alias for records where the key is the pallet name and the value is a array of its
```
Daanvdplas, line 32:
```suggestion
/// Type alias for records where the key is the pallet name and the value is an array of its
```

crates/pop-parachains/src/bench/mod.rs
```diff
@@ -0,0 +1,238 @@
...
+			s if s == *"runtime" => GenesisBuilderPolicy::Runtime,
+			_ => unreachable!(),
+		}
+	}
+}
+
+/// Get genesis builder preset names of the runtime.
+///
+/// # Arguments
+/// * `binary_path` - Path to the runtime WASM binary.
```
Daanvdplas, line 69: Suggesting this removal so that when the time comes where runtimes will be ran on PVM we don't have to make this change:
```suggestion
/// * `binary_path` - Path to the runtime binary.
```

crates/pop-parachains/src/bench/mod.rs
```diff
@@ -0,0 +1,238 @@
...
+pub fn get_preset_names(binary_path: &PathBuf) -> anyhow::Result<Vec<String>> {
+	let binary = fs::read(binary_path).expect("No runtime binary found");
+	let genesis_config_builder = GenesisConfigBuilderRuntimeCaller::<HostFunctions>::new(&binary);
+	genesis_config_builder.preset_names().map_err(|e| anyhow::anyhow!(e))
+}
+
+/// Get the runtime folder path and throws error if not exist.
```
Daanvdplas, line 76:
```suggestion
/// Get the runtime folder path and throws error if it does not exist.
```

crates/pop-parachains/src/bench/mod.rs
```diff
@@ -0,0 +1,238 @@
...
+/// * `parent` - Parent path that contains the runtime folder.
+pub fn get_runtime_path(parent: &Path) -> anyhow::Result<PathBuf> {
+	["runtime", "runtimes"]
+		.iter()
+		.map(|f| parent.join(f))
+		.find(|path| path.exists())
+		.ok_or_else(|| anyhow::anyhow!("No runtime found"))
+}
+
+/// Runs FRAME pallet benchmarks using `frame-benchmarking-cli`.
```
Daanvdplas, line 88:
```suggestion
/// Runs pallet benchmarks using `frame-benchmarking-cli`.
```

crates/pop-parachains/src/bench/mod.rs
```diff
@@ -0,0 +1,238 @@
...
+/// # Arguments
+/// * `args` - Arguments to pass to the benchmarking command.
+pub fn generate_benchmarks(args: Vec<String>) -> anyhow::Result<()> {
+	let cmd = PalletCmd::try_parse_from([vec!["".to_string()], args].concat())
+		.map_err(|e| anyhow::anyhow!("Invalid command arguments: {}", e))?;
+	cmd.run_with_spec::<BlakeTwo256, HostFunctions>(None)
+		.map_err(|e| anyhow::anyhow!("Failed to run benchmarking: {}", e))
+}
+
+/// Loads a mapping of pallets and their associated extrinsics from the runtime WASM binary.
```
Daanvdplas, line 99:
```suggestion
/// Loads a mapping of pallets and their associated extrinsics from the runtime binary.
```

crates/pop-parachains/src/bench/mod.rs
```diff
@@ -0,0 +1,238 @@
...
+///
+/// # Arguments
+/// * `runtime_path` - Path to the runtime WASM binary.
```
Daanvdplas, line 102:
```suggestion
/// * `runtime_path` - Path to the runtime binary.
```

crates/pop-parachains/src/bench/mod.rs
```diff
@@ -0,0 +1,238 @@
...
+			["on_finalize", "set"]
+		);
+		assert_eq!(
+			registry.get("pallet_sudo").cloned().unwrap_or_default(),
+			["check_only_sudo_account", "remove_key", "set_key", "sudo", "sudo_as"]
+		);
+		Ok(())
+	}
+
+	fn get_mock_runtime_path(with_runtime_benchmarks: bool) -> PathBuf {
```
Daanvdplas, line 231: Is this double with: https://github.com/r0gue-io/pop-cli/blob/5181c5a64b2e0933b1c406b3499babc8d2db1fc9/crates/pop-cli/src/common/bench.rs#L283

Daanvdplas (commented):
Thanks for the super clear information provided! I was able to immediately start reviewing without any hurdles, fantastic!!!

Also well done on the PR and research that you have put into it!

I have a question and some commands that were failing for me, I have gone through all the code but haven't been able to test the functionality yet so that will come next time. 

Why `pop bench pallet` and not `pop bench`?

Some commands that were failing:
- `pop bench pallet --list` not so clear error:
Screenshot 2025-03-05 at 15 36 21

Resolved by not throwing error anymore. Listing pallets for non `runtime-benchmarks` feature runtime binary works normally. 

- `pop bench pallet`:
Screenshot 2025-03-05 at 16 30 59

@chungquantin Resolved: 
Screenshot 2025-03-06 at 19 02 56

- `pop bench pallet --runtime=runtime --pallet= --extrinsic=` in `pop-node` folder:
Screenshot 2025-03-05 at 16 40 23


@chungquantin Resolved: 
Screenshot 2025-03-06 at 19 07 22

crates/pop-cli/src/commands/bench/pallet.rs
```diff
@@ -1,380 +1,1342 @@
...
-		if cmd.list.is_some() || cmd.json_output {
-			if let Err(e) = run_pallet_benchmarking(cmd) {
-				return display_message(&e.to_string(), false, cli);
-			}
+	pub async fn execute(&mut self, cli: &mut impl cli::traits::Cli) -> anyhow::Result<()> {
+		// If `all` is provided, we override the value of `pallet` and `extrinsic` to select all.
+		if self.all {
+			self.pallet = Some(ALL_SELECTED.to_string());
+			self.extrinsic = Some(ALL_SELECTED.to_string());
+			self.all = false;
```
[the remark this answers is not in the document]
chungquantin: The issue is if you don't set it to false, `--all` will override any value of `--pallet` and `--extrinsic`

crates/pop-cli/src/common/bench.rs
```diff
@@ -0,0 +1,456 @@
...
+/// * `skip_confirm`: A boolean indicating whether to skip confirmation prompts.
+pub async fn check_omni_bencher_and_prompt(
+	cli: &mut impl Cli,
+	cache_path: &Path,
+	skip_confirm: bool,
+) -> anyhow::Result<PathBuf> {
+	Ok(match cmd("which", &["frame-omni-bencher"]).stdout_capture().run() {
+		Ok(output) => {
+			let path = String::from_utf8(output.stdout)?;
+			PathBuf::from(path.replace("\n", ""))
```
AlexD10S, line 37: Not requesting a change, but I belieive you can use `path.trim()` here for the same

crates/pop-cli/src/common/bench.rs
```diff
@@ -0,0 +1,456 @@
...
+/// Guide the user to select a runtime project.
+///
+/// # Arguments
+/// * `cli`: Command line interface.
+/// * `project_path`: Path to the project containing runtimes.
+pub fn guide_user_to_select_runtime(
+	cli: &mut impl Cli,
+	project_path: &PathBuf,
+) -> anyhow::Result<PathBuf> {
+	let runtimes = fs::read_dir(project_path).expect("No project found");
```
AlexD10S, line 216:
```suggestion
	let runtimes = fs::read_dir(project_path)?;
```

crates/pop-parachains/src/bench/mod.rs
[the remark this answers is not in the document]
AlexD10S: It makes sense to keep it then, thanks!

crates/pop-parachains/src/bench/mod.rs
```diff
@@ -0,0 +1,238 @@
...
+		}
+	}
+}
+
+/// Get genesis builder preset names of the runtime.
+///
+/// # Arguments
+/// * `binary_path` - Path to the runtime WASM binary.
+pub fn get_preset_names(binary_path: &PathBuf) -> anyhow::Result<Vec<String>> {
+	let binary = fs::read(binary_path).expect("No runtime binary found");
```
AlexD10S, line 71:
```suggestion
	let binary = fs::read(binary_path)?;
```

crates/pop-parachains/src/bench/mod.rs
```diff
@@ -0,0 +1,238 @@
...
+		write!(f, "{}", s)
+	}
+}
+
+impl From<String> for GenesisBuilderPolicy {
+	fn from(s: String) -> Self {
+		match s {
+			s if s == *"none" => GenesisBuilderPolicy::None,
+			s if s == *"runtime" => GenesisBuilderPolicy::Runtime,
+			_ => unreachable!(),
```
AlexD10S, line 61: `unreachable! `panics right? Better to throw an error

crates/pop-parachains/src/bench/mod.rs
```diff
@@ -0,0 +1,341 @@
...
+		.find(|path| path.exists())
+		.ok_or_else(|| anyhow::anyhow!("No runtime found"))
+}
+
+/// Runs FRAME pallet benchmarks using `frame-benchmarking-cli`.
+///
+/// # Arguments
+/// * `args` - Arguments to pass to the benchmarking command.
+pub fn generate_benchmarks(args: Vec<String>) -> anyhow::Result<()> {
+	let cmd = PalletCmd::try_parse_from([vec!["".to_string()], args].concat())
```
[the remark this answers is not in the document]
AlexD10S: Thanks for the great explanation. Your approach makes more sense

crates/pop-cli/src/commands/bench/pallet.rs
```diff
@@ -1,380 +1,1342 @@
...
-		let path = runtime.unwrap().path();
-		let manifest = from_path(Some(path.as_path()))?;
-		let package = manifest.package();
-		let name = package.clone().name;
-		let description = package.description().unwrap_or_default().to_string();
-		prompt = prompt.item(path, &name, &description);
+) -> anyhow::Result<String> {
+	let pallets = pallets(registry, excluded_pallets);
+	if pallets.is_empty() {
+		return Err(anyhow::anyhow!("No pallets found for the runtime"));
```
AlexD10S, line 798: I noticed that when I run `pop bench pallet` in a parachain built without `--features runtime-benchmarks`, I get the error:
`"Error: No pallets found for the runtime."`
It might be better to improve this error message by suggesting that the parachain should be built with -`-features runtime-benchmarks`. What do you think?

crates/pop-cli/src/common/mod.rs
chungquantin: Added comment. But other public modules in the file is not commented as well.

crates/pop-cli/src/common/bench.rs
```diff
@@ -0,0 +1,456 @@
...
+
+const DEFAULT_RUNTIME_DIR: &str = "./runtime";
+
+/// Checks the status of the `frame-omni-bencher` binary, use the local binary if available.
+/// Otherwise, sources it if necessary, and prompts the user to update it if the existing binary in
+/// cache is not the latest version.
+///
+/// # Arguments
+/// * `cli`: Command line interface.
+/// * `cache_path`: The cache directory path.
```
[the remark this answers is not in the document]
chungquantin: Screenshot 2025-03-06 at 13 44 02


To keep it consistent with other functions.

crates/pop-cli/src/common/bench.rs
chungquantin: Resolved in https://github.com/r0gue-io/pop-cli/pull/420/commits/89791b8657a36ac42a0b9adee27fb0b5a3225e75

crates/pop-parachains/src/bench/mod.rs
chungquantin: Resolved in https://github.com/r0gue-io/pop-cli/pull/420/commits/89791b8657a36ac42a0b9adee27fb0b5a3225e75

crates/pop-parachains/src/bench/mod.rs
chungquantin: Resolved in https://github.com/r0gue-io/pop-cli/pull/420/commits/89791b8657a36ac42a0b9adee27fb0b5a3225e75

crates/pop-cli/src/common/bench.rs
chungquantin: Agree. Resolved in https://github.com/r0gue-io/pop-cli/pull/420/commits/89791b8657a36ac42a0b9adee27fb0b5a3225e75

crates/pop-parachains/src/bench/mod.rs
chungquantin: Resolved in https://github.com/r0gue-io/pop-cli/pull/420/commits/40a80c6c0bcd771ef886c3e5e84b67cc6e9634df

crates/pop-parachains/src/bench/mod.rs
chungquantin: Resolved in https://github.com/r0gue-io/pop-cli/pull/420/commits/40a80c6c0bcd771ef886c3e5e84b67cc6e9634df

crates/pop-parachains/src/bench/mod.rs
chungquantin: Resolved in https://github.com/r0gue-io/pop-cli/pull/420/commits/40a80c6c0bcd771ef886c3e5e84b67cc6e9634df

crates/pop-cli/src/common/bench.rs
```diff
@@ -0,0 +1,456 @@
...
+	}
+	Ok(bencher_path)
+}
+
+/// Ensure that the runtime WASM binary exists and is ready for use. If the binary is not found, it
+/// triggers a build process.
+///
+/// # Arguments
+/// * `cli`: Command line interface.
+/// * `mode`: A reference to the profile mode.
```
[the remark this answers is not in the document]
chungquantin: Resolved in https://github.com/r0gue-io/pop-cli/pull/420/commits/40a80c6c0bcd771ef886c3e5e84b67cc6e9634df

crates/pop-cli/src/common/bench.rs
```diff
@@ -0,0 +1,456 @@
...
+		},
+	}
+}
+
+/// Check the genesis builder policy and prompt the user to select one if there are genesis builder
+/// presets available.
+///
+/// # Arguments
+/// * `cli`: Command line interface.
+/// * `runtime_path`: A reference to the runtime path.
```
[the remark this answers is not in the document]
chungquantin: Resolved in https://github.com/r0gue-io/pop-cli/pull/420/commits/40a80c6c0bcd771ef886c3e5e84b67cc6e9634df

crates/pop-cli/src/common/bench.rs
chungquantin: Resolved in https://github.com/r0gue-io/pop-cli/pull/420/commits/40a80c6c0bcd771ef886c3e5e84b67cc6e9634df

crates/pop-cli/src/common/bench.rs
chungquantin: Resolved in https://github.com/r0gue-io/pop-cli/pull/420/commits/40a80c6c0bcd771ef886c3e5e84b67cc6e9634df

crates/pop-cli/src/common/bench.rs
chungquantin: Resolved in https://github.com/r0gue-io/pop-cli/pull/420/commits/40a80c6c0bcd771ef886c3e5e84b67cc6e9634df

crates/pop-parachains/src/bench/mod.rs
chungquantin: Resolved in https://github.com/r0gue-io/pop-cli/pull/420/commits/40a80c6c0bcd771ef886c3e5e84b67cc6e9634df

crates/pop-parachains/src/bench/mod.rs
chungquantin: Resolved in https://github.com/r0gue-io/pop-cli/pull/420/commits/40a80c6c0bcd771ef886c3e5e84b67cc6e9634df

crates/pop-cli/src/common/bench.rs
```diff
@@ -0,0 +1,456 @@
...
+///
+/// # Arguments
+/// * `cli`: Command line interface.
+/// * `mode`: A reference to the profile mode.
+pub fn ensure_runtime_binary_exists(cli: &mut impl Cli, mode: &Profile) -> anyhow::Result<PathBuf> {
+	let cwd = current_dir().unwrap_or(PathBuf::from("./"));
+	let target_path = mode.target_directory(&cwd).join("wbuild");
+	let runtime_path = guide_user_to_select_runtime_path(cli, &cwd)?;
+
+	// Return immediately if the user has specified a path to the runtime binary.
```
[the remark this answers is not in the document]
chungquantin: Resolved in https://github.com/r0gue-io/pop-cli/pull/420/commits/40a80c6c0bcd771ef886c3e5e84b67cc6e9634df

crates/pop-cli/src/common/bench.rs
```diff
@@ -0,0 +1,456 @@
...
+				binary.path().to_str().unwrap()
+			));
+			bencher_path = binary.path();
+		}
+	}
+	Ok(bencher_path)
+}
+
+/// Ensure that the runtime WASM binary exists and is ready for use. If the binary is not found, it
+/// triggers a build process.
```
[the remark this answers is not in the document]
chungquantin: Resolved in https://github.com/r0gue-io/pop-cli/pull/420/commits/40a80c6c0bcd771ef886c3e5e84b67cc6e9634df

crates/pop-cli/src/common/bench.rs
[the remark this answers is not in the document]
chungquantin: Resolved in https://github.com/r0gue-io/pop-cli/pull/420/commits/40a80c6c0bcd771ef886c3e5e84b67cc6e9634df by removing the argument

crates/pop-cli/src/common/bench.rs
```diff
@@ -0,0 +1,456 @@
...
+	fs,
+	path::{Path, PathBuf},
+};
+use strum::{EnumMessage, IntoEnumIterator};
+
+const DEFAULT_RUNTIME_DIR: &str = "./runtime";
+
+/// Checks the status of the `frame-omni-bencher` binary, use the local binary if available.
+/// Otherwise, sources it if necessary, and prompts the user to update it if the existing binary in
+/// cache is not the latest version.
```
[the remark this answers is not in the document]
chungquantin: Resolved in https://github.com/r0gue-io/pop-cli/pull/420/commits/40a80c6c0bcd771ef886c3e5e84b67cc6e9634df

crates/pop-cli/src/commands/bench/pallet.rs
```diff
@@ -1,380 +1,1342 @@
...
 };
-use std::{env::current_dir, fs, path::PathBuf};
+use strum::{EnumMessage, IntoEnumIterator};
+use strum_macros::{EnumIter, EnumMessage as EnumMessageDerive};
+
+const ALL_SELECTED: &str = "*";
+
+#[derive(Args)]
+pub(crate) struct BenchmarkPallet {
+	/// Select a FRAME Pallet to benchmark, or `*` for all (in which case `extrinsic` must be `*`).
```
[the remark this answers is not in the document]
chungquantin: Resolved in https://github.com/r0gue-io/pop-cli/pull/420/commits/40a80c6c0bcd771ef886c3e5e84b67cc6e9634df

crates/pop-parachains/src/bench/mod.rs
chungquantin: Yes they are the same but I don't find trying to make this shared across packages seems reasonable. It is only a method to get mock runtime and the path can be different between folder levels.

crates/pop-cli/src/commands/bench/pallet.rs
chungquantin: I agree. This could be improved. Resolved in https://github.com/r0gue-io/pop-cli/pull/420/commits/ff2f39d1f59d7bb7766d604804d2c18fc19bfcc3

Now if the runtime-benchmarks feature is not enabled, it will throw error: 

Screenshot 2025-03-06 at 18 57 24

crates/pop-cli/src/common/bench.rs
```diff
@@ -0,0 +1,456 @@
...
+	Ok(bencher_path)
+}
+
+/// Ensure that the runtime WASM binary exists and is ready for use. If the binary is not found, it
+/// triggers a build process.
+///
+/// # Arguments
+/// * `cli`: Command line interface.
+/// * `mode`: A reference to the profile mode.
+pub fn ensure_runtime_binary_exists(cli: &mut impl Cli, mode: &Profile) -> anyhow::Result<PathBuf> {
```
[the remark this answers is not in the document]
chungquantin: Resolved in https://github.com/r0gue-io/pop-cli/pull/420/commits/8e3fee344e1d7beb1c80fdf51913c5b781a8ac31

crates/pop-cli/src/commands/bench/pallet.rs
```diff
@@ -1,380 +1,1342 @@
...
+
+		// Only prompt user to update parameters when `skip_menu` is not provided.
+		if !self.skip_menu {
+			self.ensure_pallet_registry(cli, &mut registry).await?;
+			loop {
+				let option = guide_user_to_select_menu_option(self, cli, &mut registry).await?;
+				match option.update_arguments(self, &mut registry, cli).await {
+					Ok(true) => break,
+					Ok(false) => continue,
+					Err(e) => cli.info(e)?,
```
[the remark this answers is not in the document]
AlexD10S: I don't mind to throw an error, it was more about the message` cliclack::log::error` or maybe a `cliclack::log::warning` instead of an `info`.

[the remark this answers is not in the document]
chungquantin: Ah I see, apologies I did not aware of the feature in `cliclack`

crates/pop-cli/src/commands/bench/pallet.rs
AlexD10S: Really cool!

crates/pop-cli/src/commands/bench/pallet.rs
```diff
@@ -1,380 +1,1342 @@
...
+	#[arg(short, long, value_parser = parse_pallet_name, default_value_if("all", "true", Some("*".into())))]
+	pub pallet: Option<String>,
+
+	/// Select an extrinsic inside the pallet to benchmark, or `*` for all.
+	#[arg(short, long, default_value_if("all", "true", Some("*".into())))]
+	pub extrinsic: Option<String>,
+
+	/// Comma separated list of pallets that should be excluded from the benchmark.
+	#[arg(long, value_parser, num_args = 1.., value_delimiter = ',')]
+	pub exclude_pallets: Vec<String>,
```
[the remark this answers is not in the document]
Daanvdplas: And with the flag `all`? If that is not allowed we could add the `conflicts_with`

Daanvdplas:
- I did `pop bench pallet` and gave input to points 0 to 4, now in the parameter menu they are asked again:
Screenshot 2025-03-07 at 16 19 28
If it is an easy fix, can we eliminate the things that are provided by the user from the menu?

- Could we ask if the user would like to go through the parameter menu? The skip flag is already a great flag!
- Loading the menu is also slow, do we know why?
- Can we remove the word FRAME here?
Screenshot 2025-03-07 at 16 25 09

Could we add a newline before outputting the generated command (great feature!):
Screenshot 2025-03-07 at 16 39 25

When we can choose the extrinsics to benchmark, can we add `(select with space)`?
Screenshot 2025-03-07 at 16 42 29

Daanvdplas (approved):
I left some UX improvements / suggestions, but wow what an amazing PR!

chungquantin:
@Daanvdplas 

> If it is an easy fix, can we eliminate the things that are provided by the user from the menu?

This goes against the benefit of the parameter menu that we defined in the spec. The benefit of the menu is to allow user to update the parameter again if they mistype. Reason for that because benchmarking has more parameters compared to other commands. Hence, this gives the user more control to the parameters of the command. 

> Could we ask if the user would like to go through the parameter menu? The skip flag is already a great flag!

Agree with this suggestion. Added in the next commit. 

> When we can choose the extrinsics to benchmark, can we add (select with space)?

Agree with this suggestion. Added in the next commit.

Daanvdplas (commented):
Thanks for the last UX improvements!

> This goes against the benefit of the parameter menu that we defined in the spec. The benefit of the menu is to allow user to update the parameter again if they mistype. Reason for that because benchmarking has more parameters compared to other commands. Hence, this gives the user more control to the parameters of the command.

I disagree, one provides inputs to not have to provide them again. This also works for the other interactive UI's for the CLI.

crates/pop-cli/src/commands/bench/pallet.rs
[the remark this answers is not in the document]
chungquantin: Resolved by changing it to `cliclack::log::error`

PR #427, opened 2025-02-26
feat: use a proxy for the registration

AlexD10S:
This is the last step in a larger PR: https://github.com/r0gue-io/pop-cli/pull/404. It is a recommended practice to use a proxy when registering a parachain. Using a proxy allows for ownership changes, as the registering account itself cannot be modified later.

This PR introduces the following changes:

- Prompts the user to confirm whether they want to register using a proxy. If they choose to proceed, they can input the proxy account.
- Wraps the registration call within a proxy extrinsic, following the existing approach used for sudo calls (https://github.com/r0gue-io/pop-cli/blob/main/crates/pop-parachains/src/call/mod.rs#L44)

**How to test**
I start the Paseo parachain locally by running:
```
pop up parachain -f ./tests/networks/paseo.toml
```

I add a proxy into my account manually using PolkadotJS:
Captura de pantalla 2025-02-28 a las 13 33 33

 And then  run `pop up` in your parachain folder (or specify the folder with the `path` flag)
```
pop up --path ../test-chain --relay-chain-url ws://127.0.0.1:49155
```
This will prompt you to add the proxied address (your account) and open the signing portal, allowing you to sign with the proxy account.

[sc-2542]


crates/pop-parachains/src/call/metadata/mod.rs
```diff
@@ -82,7 +82,6 @@ pub fn parse_chain_metadata(client: &OnlineClient<SubstrateConfig>) -> Result<Ve
 											// dispatchable function as unsupported rather than
 											// error.
 											is_supported = false;
-											parsed_params.clear();
```
AlexD10S, line 85: When parsing the chain metadata, we can't process arguments that are nested calls (e.g., `sudo` or `proxy`), they are not supported via `pop call`. I removed this line to ensure we can still parse the remaining arguments, allowing us to manually construct the `proxy` call (see `construct_proxy_extrinsic`).

crates/pop-cli/src/commands/up/parachain.rs
```diff
@@ -37,6 +39,10 @@ pub struct UpChainCommand {
 	/// Websocket endpoint of the relay chain.
 	#[arg(long)]
 	pub(crate) relay_url: Option<Url>,
+	///  Proxied account address. The proxy can act on behalf of this account.
+	/// Specify the address type, e.g. `Id(5FHneW46xGXgs5mUiveU4sbTyGBzmstUspZC92UhjJM694ty)`.
```
evilrobot-01, line 43: Assume it has to be `Id(..)` as its a `MultiAddress`? Can we also support a raw 'id' and wrap it in `Id()`, whilst still preserving the ability to use the other variants of MultiAddress?

Can a better type be used than string perhaps, as with Url?

crates/pop-cli/src/commands/up/parachain.rs
```diff
@@ -66,18 +72,40 @@ impl UpChainCommand {
...
+		Ok(UpChain { id: para_id, genesis_state, genesis_code, chain, proxy })
+	}
+
+	// Retrieves the proxied address, prompting the user if none is specified.
+	fn resolve_proxied_address(&self, chain: &Chain, cli: &mut impl Cli) -> Result<ProxyConfig> {
+		if let Some(addr) = &self.proxied_address {
+			return Ok(ProxyConfig::Address(addr.clone()));
+		}
+		if cli.confirm("Would you like to use a proxy for registration? This is considered a best practice.").interact()? {
+			cli.info("Enter the account the proxy will represent.")?;
```
evilrobot-01, line 88: Is this a little ambigous to someone that doesnt know what is required here? The description should also align with that of the command.

The proxy dispatchable uses ` - real: The account that the proxy will make a call on behalf of.`

crates/pop-cli/src/commands/up/parachain.rs
```diff
@@ -37,6 +39,10 @@ pub struct UpChainCommand {
 	/// Websocket endpoint of the relay chain.
 	#[arg(long)]
 	pub(crate) relay_url: Option<Url>,
+	///  Proxied account address. The proxy can act on behalf of this account.
```
evilrobot-01, line 42:
```suggestion
	///  Proxied address. Your account must be registered as a proxy which can act on behalf of this account.
```

crates/pop-cli/src/commands/up/parachain.rs
```diff
@@ -97,12 +125,19 @@ impl UpChainCommand {
 	}
 }
 
+#[derive(Clone, Debug, PartialEq)]
+enum ProxyConfig {
```
evilrobot-01, line 129:
```suggestion
enum Proxy {
```

I dont see that the `Config` suffix adds any value based on how simple the enum definition is.

crates/pop-cli/src/commands/up/parachain.rs
```diff
@@ -124,15 +159,18 @@ impl UpChain {
 		let state = std::fs::read_to_string(genesis_state)
 			.map_err(|err| anyhow!("Failed to read genesis state file: {}", err.to_string()))?;
 		let code = std::fs::read_to_string(genesis_code)
-			.map_err(|err| anyhow!("Failed to read genesis code file: {}", err.to_string()))?;
-		let xt = construct_extrinsic(ex, vec![id.to_string(), state, code])?;
+			.map_err(|err| anyhow!("Failed to read genesis state file: {}", err.to_string()))?;
```
evilrobot-01, line 162:
```suggestion
			.map_err(|err| anyhow!("Failed to read genesis code file: {}", err.to_string()))?;
```

crates/pop-cli/src/commands/up/parachain.rs
```diff
@@ -115,7 +150,7 @@ impl UpChain {
 
 	// Prepares and returns the encoded call data for registering a parachain.
 	fn prepare_register_parachain_call_data(&self) -> Result<Vec<u8>> {
-		let UpChain { id, genesis_code, genesis_state, chain } = self;
+		let UpChain { id, genesis_code, genesis_state, chain, proxy } = self;
 		let ex = find_dispatchable_by_name(
```
evilrobot-01, line 154: `let register` might be more informative

crates/pop-cli/src/commands/up/parachain.rs
evilrobot-01: Can you not achieve the same with `type Proxy = Option<String>;`?

crates/pop-cli/src/commands/up/parachain.rs
```diff
@@ -124,15 +159,18 @@ impl UpChain {
 		let state = std::fs::read_to_string(genesis_state)
 			.map_err(|err| anyhow!("Failed to read genesis state file: {}", err.to_string()))?;
 		let code = std::fs::read_to_string(genesis_code)
-			.map_err(|err| anyhow!("Failed to read genesis code file: {}", err.to_string()))?;
-		let xt = construct_extrinsic(ex, vec![id.to_string(), state, code])?;
+			.map_err(|err| anyhow!("Failed to read genesis state file: {}", err.to_string()))?;
+		let mut xt = construct_extrinsic(ex, vec![id.to_string(), state, code])?;
```
evilrobot-01, line 163: Why is the para id encoded as a string? subxt thing?

crates/pop-parachains/src/call/mod.rs
```diff
@@ -46,6 +46,36 @@ pub fn construct_sudo_extrinsic(xt: DynamicPayload) -> DynamicPayload {
...
+	pallets: &[Pallet],
+	proxied_account: String,
+	xt: DynamicPayload,
+) -> Result<DynamicPayload, Error> {
+	let proxy_function = find_dispatchable_by_name(pallets, "Proxy", "proxy")?;
+	// `find_dispatchable_by_name` doesn't support parsing parameters that are calls.
+	// Therefore, we only parse the first two parameters for the proxy call
+	// using `parse_dispatchable_arguments`, while the last parameter (which is the call)
+	// must be manually added.
+	let required_params: Vec<Param> = proxy_function.params.iter().take(2).cloned().collect();
```
evilrobot-01, line 66: Can you not consume the params collection with into_iter() to avoid cloning?

crates/pop-parachains/src/call/mod.rs
```diff
@@ -46,6 +46,36 @@ pub fn construct_sudo_extrinsic(xt: DynamicPayload) -> DynamicPayload {
...
+	let parsed_args: Vec<Value> = metadata::parse_dispatchable_arguments(
+		&required_params,
+		vec![proxied_account, "None()".to_string()],
+	)?;
+
+	Ok(subxt::dynamic::tx(
+		"Proxy",
+		"proxy",
+		[parsed_args[0].clone(), parsed_args[1].clone(), xt.into_value()].to_vec(),
```
evilrobot-01, line 75: Can avoid clone with: `let real = parsed_args.remove(0)`

evilrobot-01 (commented):
Really enjoyed this PR, seems powerful!

Quick high level review only, didnt look at tests.

crates/pop-parachains/src/call/mod.rs
AlexD10S: I can't do it because `find_dispatachable_by_name` returns a `&Function`

crates/pop-parachains/src/call/mod.rs
AlexD10S: https://github.com/r0gue-io/pop-cli/pull/427/commits/bfbb8e44ba4218d20ce9f9b834793f8cf47068e2

crates/pop-cli/src/commands/up/parachain.rs
AlexD10S: In the main feat branch name was changed to `dispatachable`

crates/pop-cli/src/commands/up/parachain.rs
AlexD10S: https://github.com/r0gue-io/pop-cli/pull/427/commits/6fdf6338d83bfd89ab0c91543e0ab0ac30ea8db7

crates/pop-cli/src/commands/up/parachain.rs
AlexD10S: Yes, the type `Proxy = Option<String>;` seems enough. Refactored in this commit https://github.com/r0gue-io/pop-cli/pull/427/commits/f18de68e93111b9f14bde797ec587825157f3696

crates/pop-cli/src/commands/up/parachain.rs
AlexD10S: https://github.com/r0gue-io/pop-cli/pull/427/commits/34b95399b9308df0c38f1fcaea77b71019b3cd91

crates/pop-cli/src/commands/up/parachain.rs
AlexD10S: https://github.com/r0gue-io/pop-cli/pull/427/commits/34b95399b9308df0c38f1fcaea77b71019b3cd91

crates/pop-cli/src/commands/call/chain.rs
```diff
@@ -473,7 +473,7 @@ fn prompt_predefined_actions(pallets: &[Pallet], cli: &mut impl Cli) -> Result<O
 }
 
 // Prompts the user for the value of a parameter.
```
Daanvdplas, line 475: Public docs

crates/pop-cli/src/commands/up/rollup.rs
```diff
@@ -37,6 +41,12 @@ pub struct UpCommand {
 	/// Websocket endpoint of the relay chain.
 	#[arg(long)]
 	pub(crate) relay_chain_url: Option<Url>,
+	/// Proxied address. Your account must be registered as a proxy which can act on behalf of this
+	/// account.  You can specify the MultiAddress type explicitly, e.g.,
+	/// `Id(5FHneW46xGXgs5mUiveU4sbTyGBzmstUspZC92UhjJM694ty)`. If no type is provided, it
+	/// defaults to `Id()`.
```
Daanvdplas, line 47:
```suggestion
	/// Proxied address. The account used for signing must be registered as a proxy for this address. You can specify the MultiAddress type explicitly, e.g.,
	/// `Id(5FHneW46xGXgs5mUiveU4sbTyGBzmstUspZC92UhjJM694ty)`. If no type is provided, it
	/// defaults to `Id()`.
```

Daanvdplas: Further improvements but out of scope:
````
	/// Proxied address. The account used for signing must be registered as a proxy for this address. You can specify the address type explicitly, e.g.,
	/// `Polkadot(5FHneW46xGXgs5mUiveU4sbTyGBzmstUspZC92UhjJM694ty)`. If no type is provided, it
	/// defaults to `Polkadot()`.
```

crates/pop-cli/src/commands/up/rollup.rs
```diff
@@ -69,18 +79,41 @@ impl UpCommand {
...
-		let id = self.resolve_id(&chain, cli).await?;
+		let proxy = self.resolve_proxied_address(&chain, cli)?;
+		let id = self.resolve_id(&chain, &proxy, cli).await?;
 		let (genesis_code, genesis_state) = self.resolve_genesis_files(id, cli).await?;
-		Ok(Registration { id, genesis_state, genesis_code, chain })
+		Ok(Registration { id, genesis_state, genesis_code, chain, proxy })
+	}
+
+	// Retrieves the proxied address, prompting the user if none is specified.
+	fn resolve_proxied_address(&self, chain: &Chain, cli: &mut impl Cli) -> Result<Proxy> {
```
Daanvdplas, line 89:
```suggestion
	fn resolve_proxied_address(&self, chain: &Chain, cli: &mut impl Cli) -> Result<Proxy> {
		let proxy = find_dispatchable_by_name(&chain.pallets, "Proxy", "proxy")?;
		let proxy_address = if let Some(a) = &self.proxied_address {
			Some(if proxy.params[0].sub_params.iter().any(|p| a.starts_with(&format!("{}(", p.name))) {
				a.to_string()
			} else {
				format!("Id({})", a)
			})
		} else if cli.confirm("Would you like to use a proxy for registration? This is considered a best practice.").interact()? {
			cli.info("Enter the account that the proxy will make a call on behalf of.")?;
			Some(prompt_for_param(cli, &proxy.params[0])?)
		} else { None };
		Ok(proxy_address)
	}
```

crates/pop-cli/src/commands/up/rollup.rs
```diff
@@ -236,11 +343,23 @@ mod tests {
...
 		assert_eq!(call_data, decode_call_data(encoded_reserve_extrinsic)?);
+
+		// Ensure `prepare_reserve_call_data` works with a proxy.
+		let call_data = prepare_reserve_call_data(
+			&chain,
+			&Some("Id(13czcAAt6xgLwZ8k6ZpkrRL5V2pjKEui3v9gHAN9PoxYZDbf)".to_string()),
+			&mut cli,
+		)?;
+		// Encoded call data for a proxy extrinsic with reserve as the call.
+		// Reference: https://polkadot.js.org/apps/?rpc=wss%3A%2F%2Fpolkadot.public.curie.radiumblock.co%2Fws#/extrinsics/decode/0x1d000073ebf9c947490b9170ea4fd3031ae039452e428531317f76bf0a02124f8166de004605
```
Daanvdplas, line 359: I have seen this a few cases in this codebase, but why are we not encoding it ourselves?

crates/pop-parachains/src/call/mod.rs
```diff
@@ -46,6 +46,33 @@ pub fn construct_sudo_extrinsic(xt: DynamicPayload) -> DynamicPayload {
...
+/// * `pallets`: List of pallets available within the chain's runtime.
+/// * `proxied_account` - The account on whose behalf the proxy will act.
+/// * `xt`: The extrinsic representing the dispatchable function call to be dispatched using the
+///   proxy.
+pub fn construct_proxy_extrinsic(
+	pallets: &[Pallet],
+	proxied_account: String,
+	xt: DynamicPayload,
+) -> Result<DynamicPayload, Error> {
+	let proxy_function = find_dispatchable_by_name(pallets, "Proxy", "proxy")?;
```
Daanvdplas, line 61: Mostly curious so don't take it as a request, by why you are searching for the function, is it not known (e.g. constant)?

Daanvdplas (commented):
Looking good! Left a few comments.

Maybe a little bit out of scope but is there a way to improve these prompts?
Screenshot 2025-03-05 at 09 58 56
Screenshot 2025-03-05 at 10 05 20

`What type of address do you want to use?`
`Enter your address`

`Id` is also too vague for me, if we can, can we rename this to Polkadot address or something.

AlexD10S:
> `What type of address do you want to use?` `Enter your address`
> 
> `Id` is also too vague for me, if we can, can we rename this to Polkadot address or something.

The reason it looks is because  we retrieve the proxy metadata from the specific chain the user selects. This approach is more generic compared to hardcoding exactly what we need.

The reason for using `Id` is that it corresponds to the [MultiAddress enum](https://github.com/paritytech/polkadot-sdk/blob/master/substrate/primitives/runtime/src/multiaddress.rs#L35).

PolkadotJS follows the same approach.
Captura de pantalla 2025-03-05 a las 11 12 46

Frank suggested [adding a more informative message before this](https://github.com/r0gue-io/pop-cli/pull/427#discussion_r1975579879 ), so I updated it to: 'Enter the account that the proxy will act on behalf of when making a call.'  Maybe we can add more info in the message?

crates/pop-parachains/src/call/mod.rs
AlexD10S: Because we retrieve the proxy metadata from the specific chain the user selects, so we have the exact type that chain uses for the proxy call.

crates/pop-cli/src/commands/up/rollup.rs
```diff
@@ -37,6 +41,12 @@ pub struct UpCommand {
 	/// Websocket endpoint of the relay chain.
 	#[arg(long)]
 	pub(crate) relay_chain_url: Option<Url>,
+	/// Proxied address. Your account must be registered as a proxy which can act on behalf of this
+	/// account.  You can specify the MultiAddress type explicitly, e.g.,
+	/// `Id(5FHneW46xGXgs5mUiveU4sbTyGBzmstUspZC92UhjJM694ty)`. If no type is provided, it
+	/// defaults to `Id()`.
+	#[arg(long = "proxy")]
+	pub(crate) proxied_address: Option<String>,
```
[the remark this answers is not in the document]
chungquantin: Just look at the code again. Seems like this field is for the account that proxied. If so, no need to change.

chungquantin:
Screenshot 2025-03-05 at 17 33 16

PR description needs to be updated

crates/pop-cli/src/commands/call/chain.rs
AlexD10S: It should be only `pub(crate)`, change visibility https://github.com/r0gue-io/pop-cli/pull/427/commits/ceec043d267218288db1109a5d140cab8f0b6b69

crates/pop-cli/src/commands/up/rollup.rs
AlexD10S: Refactored to https://github.com/r0gue-io/pop-cli/pull/427/commits/be1b4d6e94ada51a2ff8d1a613e748cd860a8f7d It’s not exactly the same, but I see the improvement in how you return `proxy_address`, reducing the number of `Some(` calls.

crates/pop-cli/src/commands/up/rollup.rs
```diff
@@ -69,18 +79,41 @@ impl UpCommand {
...
 		let (genesis_code, genesis_state) = self.resolve_genesis_files(id, cli).await?;
-		Ok(Registration { id, genesis_state, genesis_code, chain })
+		Ok(Registration { id, genesis_state, genesis_code, chain, proxy })
+	}
+
+	// Retrieves the proxied address, prompting the user if none is specified.
+	fn resolve_proxied_address(&self, chain: &Chain, cli: &mut impl Cli) -> Result<Proxy> {
+		let proxy = find_dispatchable_by_name(&chain.pallets, "Proxy", "proxy")?;
+		if let Some(addr) = &self.proxied_address {
+			let valid_multi_address: Vec<String> =
```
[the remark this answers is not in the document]
AlexD10S: `MultiAddress` https://github.com/paritytech/polkadot-sdk/blob/master/substrate/primitives/runtime/src/multiaddress.rs#L35

crates/pop-cli/src/commands/up/rollup.rs
```diff
@@ -223,6 +266,70 @@ mod tests {
...
+		cli.verify()
+	}
+
+	#[tokio::test]
+	async fn resolve_proxied_address_works() -> Result<()> {
+		let mut cli = MockCli::new()
+			.expect_confirm("Would you like to use a proxy for registration? This is considered a best practice.", true)
+			.expect_info("Enter the account that the proxy will make a call on behalf of.")
+			.expect_select(
+				"Select the value for the parameter: real",
```
chungquantin, line 282: Even though I understand we retrieve this programatically, are there any way that we can improve this parameter name? Or describes what the parameter is. If I am not familiar with the structure of pallet proxy, I find this quite hard to understand,

crates/pop-cli/src/commands/up/rollup.rs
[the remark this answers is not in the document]
AlexD10S: Unfortunately if is Proxy even being an alias it forces to specify `proxied_address`, so we need the `Option` here

crates/pop-cli/src/commands/up/rollup.rs
```diff
@@ -69,18 +79,41 @@ impl UpCommand {
...
+	}
+
+	// Retrieves the proxied address, prompting the user if none is specified.
+	fn resolve_proxied_address(&self, chain: &Chain, cli: &mut impl Cli) -> Result<Proxy> {
+		let proxy = find_dispatchable_by_name(&chain.pallets, "Proxy", "proxy")?;
+		if let Some(addr) = &self.proxied_address {
+			let valid_multi_address: Vec<String> =
+				proxy.params[0].sub_params.iter().map(|p| p.name.to_string()).collect();
+			return Ok(
+				if valid_multi_address.iter().any(|t| addr.starts_with(&format!("{}(", t))) {
```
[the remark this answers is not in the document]
AlexD10S: Good point, done in https://github.com/r0gue-io/pop-cli/pull/427/commits/8897f2ab54b9a60ea6b182e3d118e727aee377a5

crates/pop-cli/src/commands/up/rollup.rs
```diff
@@ -69,18 +79,41 @@ impl UpCommand {
...
+					Some(addr.to_string())
+				} else {
+					Some(format!("Id({})", addr))
+				},
+			);
+		}
+		if cli.confirm("Would you like to use a proxy for registration? This is considered a best practice.").interact()? {
+			cli.info("Enter the account that the proxy will make a call on behalf of.")?;
+			let address = prompt_for_param(cli, &proxy.params[0])?;
```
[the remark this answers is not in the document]
AlexD10S: https://github.com/r0gue-io/pop-cli/pull/427/commits/8897f2ab54b9a60ea6b182e3d118e727aee377a5

crates/pop-cli/src/commands/up/rollup.rs
```diff
@@ -154,14 +191,17 @@ async fn reserve(chain: &Chain, cli: &mut impl Cli) -> Result<u32> {
...
+fn prepare_reserve_call_data(chain: &Chain, proxy: &Proxy, cli: &mut impl Cli) -> Result<Vec<u8>> {
 	let dispatchable = find_dispatchable_by_name(
 		&chain.pallets,
 		Action::Reserve.pallet_name(),
 		Action::Reserve.function_name(),
 	)?;
-	let xt = Call { function: dispatchable.clone(), use_wallet: true, ..Default::default() }
+	let mut xt = Call { function: dispatchable.clone(), use_wallet: true, ..Default::default() }
 		.prepare_extrinsic(&chain.client, cli)?;
+	if let Some(addr) = proxy {
```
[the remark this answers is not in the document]
AlexD10S: Agreed! The logic for using a proxy can be added in the `pop call`, just like we do with `sudo`, but it would require more changes in `pop call`. I feel it's out of scope for this PR, I can create an Issue.

crates/pop-cli/src/commands/up/rollup.rs
```diff
@@ -223,6 +266,70 @@ mod tests {
...
+						("Address32".to_string(), "".to_string()),
+						("Address20".to_string(), "".to_string()),
+					]
+					.to_vec(),
+				),
+				0, // "Id" action
+			)
+			.expect_input(
+				"Enter the value for the parameter: Id",
+				"5FHneW46xGXgs5mUiveU4sbTyGBzmstUspZC92UhjJM694ty".into(),
```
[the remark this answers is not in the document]
AlexD10S: https://github.com/r0gue-io/pop-cli/pull/427/commits/8897f2ab54b9a60ea6b182e3d118e727aee377a5

crates/pop-cli/src/commands/up/rollup.rs
```diff
@@ -214,6 +254,9 @@ mod tests {
 			id: Some(2000),
 			genesis_state: Some(genesis_state.clone()),
 			genesis_code: Some(genesis_code.clone()),
+			proxied_address: Some(
+				"Id(13czcAAt6xgLwZ8k6ZpkrRL5V2pjKEui3v9gHAN9PoxYZDbf)".to_string(),
```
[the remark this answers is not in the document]
AlexD10S: https://github.com/r0gue-io/pop-cli/pull/427/commits/8897f2ab54b9a60ea6b182e3d118e727aee377a5

crates/pop-common/src/account_id.rs
```diff
@@ -0,0 +1,39 @@
+// SPDX-License-Identifier: GPL-3.0
```
AlexD10S, line 1: This function + tests where in the pop-contracts crate now is used in chain functionality has to be moved in pop-common

crates/pop-contracts/src/lib.rs
```diff
@@ -30,7 +30,7 @@ pub use up::{
 };
```
AlexD10S, line 30: The changes in the `pop-contracts` files are only for the moving of `parse_account` into the common crate.

crates/pop-cli/src/commands/up/rollup.rs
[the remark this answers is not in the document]
chungquantin: Thank you!

chungquantin (approved):
Approved! PR looks very clean to me! It provides a smooth dev ex and I really love the flow and how smooth it is!

crates/pop-cli/src/commands/up/rollup.rs
Daanvdplas: Really not the same?

crates/pop-cli/src/commands/up/rollup.rs
Daanvdplas: Yes but why don't we encode the value ourselves in the code in stead of copy pasting a value from a website :) ?

crates/pop-cli/src/commands/up/rollup.rs
```diff
@@ -236,6 +236,32 @@ async fn generate_spec_files(
 	))
 }
 
+// Prompt the user to select the value of the variant parameter and the address.
+// Output example: `Id(5GrwvaEF5zXb26Fz9rcQpDWS57CtERHpNehXCPcNoHGKutQY)` for the `Id` variant.
+fn prompt_for_proxy_address(cli: &mut impl Cli, param: &Param) -> Result<String> {
```
Daanvdplas: Unit test?

crates/pop-cli/src/commands/up/rollup.rs
```diff
@@ -225,6 +270,37 @@ mod tests {
 		assert_eq!(chain_config.genesis_code, genesis_code);
 		assert_eq!(chain_config.genesis_state, genesis_state);
 		assert_eq!(chain_config.chain.url, Url::parse(POLKADOT_NETWORK_URL)?);
+		assert_eq!(chain_config.proxy, Some(format!("Id({})", MOCK_PROXIED_ADDRESS.to_string())));
+		cli.verify()
+	}
+
+	#[tokio::test]
```
Daanvdplas, line 277: Does this need to be a tokio / async test?

AlexD10S:
> DQ: what is the `crates/pop-cli/src/assets/index.html` for?
This was introduced in the wallet signing feature to open a UI for signing via the wallet. In this PR, the only modification is in the UI to display a proxy call clearly.
This was approved in the `wallet_signing` portal repo: https://github.com/r0gue-io/pop-wallet-signing-portal/pull/7. The explanation was originally in a comment inside the code but seems to have been lost after all the changes. Next time, I'll make sure to include it in the PR description instead.

crates/pop-cli/src/commands/up/rollup.rs
AlexD10S: True. https://github.com/r0gue-io/pop-cli/pull/427/commits/911ee99dd1b9727eefc727206d31c126abc4c808

crates/pop-common/src/lib.rs
```diff
@@ -14,6 +15,8 @@ pub use subxt::{Config, PolkadotConfig as DefaultConfig};
 pub use subxt_signer::sr25519::Keypair;
 pub use templates::extractor::extract_template_files;
 
+/// Module for parsing and handling account IDs.
```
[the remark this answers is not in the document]
AlexD10S: 100% agree

PR #429, opened 2025-02-27
refactor: check binary and prompt

chungquantin:
Refactoring `check_contracts_node_and_prompt` to use the `generic check_binary_and_prompt` for improved reusability. This ensures a unified method for prompting users and sourcing binaries, simplifying future extensions with similar logic.
- Add a new trait `BinaryGenerator` that is implemented to the wrapper struct of the generate method like `contracts_node_generator`
- Add a new macro `impl_binary_generator` that must be implemented by the binary generator. For example: 

```rs
impl_binary_generator!(ContractsNodeGenerator, contracts_node_generator);
```


crates/pop-cli/src/common/mod.rs
```diff
@@ -1,5 +1,6 @@
 // SPDX-License-Identifier: GPL-3.0
 
+pub mod binary;
```
AlexD10S, line 3: Add a public comment

crates/pop-cli/src/common/binary.rs
```diff
@@ -0,0 +1,164 @@
...
+				.initial_value(true)
+				.interact()?
+		} else {
+			true
+		};
+		if latest {
+			let spinner = spinner();
+			spinner.start(format!("📦 Sourcing {binary_name}..."));
+
+			binary.source(false, &(), true).await?;
```
AlexD10S, line 51:
```suggestion
			binary.source(false, &(), true).await?;
			set_executable_permission(binary.path())?;
```

I was getting an error:
```
◇  ✅ substrate-contracts-node successfully sourced. Cached at: /Users/alexbean/Library/Caches/pop/substrate-contracts-node-v0.42.0
│
Error: IO error: Permission denied (os error 13)
```

AlexD10S (changes requested):
Amazing! Thanks for making this refactor, very useful.

Only one comment, a fix that I found testing it, include `set_executable_permission(binary.path())?;` after sourcing the `substrate-contracts-node` binary.

crates/pop-cli/src/common/binary.rs
chungquantin: Hmm quite strange we did not have this [before](https://github.com/r0gue-io/pop-cli/pull/429/files#:~:text=node...%22)%3B-,binary.source(false%2C%20%26()%2C%20true).await%3F%3B,-spinner.stop). I added your suggestion

crates/pop-cli/src/common/mod.rs
chungquantin: Resolved in the latest commit

AlexD10S (approved):

PR #434, opened 2025-03-03
fix: benchmarking logger

chungquantin:
- Remove deprecated code to set `RUST_LOG`
- Update telemetry to filter out unwanted targets following the [code in `frame-omni-bencher`](https://github.com/paritytech/polkadot-sdk/blob/master/substrate/utils/frame/omni-bencher/src/main.rs#L32).


AlexD10S (approved):
Approved!

Issue #438, opened 2025-03-04
Vulnerability dependency


chungquantin:
Issue open to track an issue with one of our dependencies when running cargo deny check advisories:

- Advisory: https://rustsec.org/advisories/RUSTSEC-2020-0163.html
- https://github.com/clap-rs/term_size-rs/pull/31

### Summary
```toml
[advisory]
id = "RUSTSEC-2020-0163"
package = "term_size"
date = "2020-11-03"
informational = "unmaintained"
url = "https://github.com/clap-rs/term_size-rs/pull/31"
[versions]
patched = []
unaffected = []
```

# `term_size` is unmaintained; use `terminal_size` instead

The [`term_size`](https://crates.io/crates/term_size) crate is no longer maintained. Consider using
[`terminal_size`](https://crates.io/crates/terminal_size) instead.

AlexD10S:
Closed with https://github.com/r0gue-io/pop-cli/pull/1006

Issue #437, opened 2025-03-04
Vulnerability dependency


chungquantin:
Issue open to track an issue with one of our dependencies when running cargo deny check advisories:

- Advisory: https://rustsec.org/advisories/RUSTSEC-2024-0384.html
- https://crates.io/crates/instant/0.1.13
- https://github.com/sebcrozet/instant/issues/52

### Summary
```toml
[advisory]
id = "RUSTSEC-2024-0384"
package = "instant"
date = "2024-09-01"
references = [
    "https://crates.io/crates/instant/0.1.13",
    "https://github.com/sebcrozet/instant/issues/52",
]
informational = "unmaintained"

[versions]
patched = []
```

# `instant` is unmaintained

This crate is no longer maintained, and the author recommends using the maintained [`web-time`] crate instead.

[`web-time`]: https://crates.io/crates/web-time

AlexD10S:
Closed with https://github.com/r0gue-io/pop-cli/pull/1006

Issue #436, opened 2025-03-04
Vulnerability dependency


chungquantin:
Issue open to track an issue with one of our dependencies when running cargo deny check advisories:

- Advisory: https://rustsec.org/advisories/RUSTSEC-2024-0388
- Original issue: https://github.com/mcarton/rust-derivative/issues/117

# Description

```toml
[advisory]
id = "RUSTSEC-2024-0388"
package = "derivative"
date = "2024-06-26"
url = "https://github.com/mcarton/rust-derivative/issues/117"
informational = "unmaintained"

[versions]
patched = []

```
# `derivative` is unmaintained; consider using an alternative

The [`derivative`](https://crates.io/crates/derivative) crate is no longer maintained.
Consider using any alternative, for instance:
- [derive_more](https://crates.io/crates/derive_more)
- [derive-where](https://crates.io/crates/derive-where)
- [educe](https://crates.io/crates/educe)

AlexD10S:
Closed with https://github.com/r0gue-io/pop-cli/pull/1006

PR #439, opened 2025-03-04
chore: update cargo-deny-action@v2

chungquantin:
There was an issue with the CI cargo-deny tackled in the PR: https://github.com/EmbarkStudios/cargo-deny-action/issues/91

Hence, we need to update our `cargo-deny-action` to v2 to sync with the update referenced from https://github.com/restatedev/restate/commit/b16d7da8812f87cd0763e96f37519426b6c9b1f2

By updating to V2, `unmaintained = "warn"` is removed completely and all unmaintained crates now emit errors. As mentioned [here](https://embarkstudios.github.io/cargo-deny/checks/advisories/cfg.html#the-version-field-optional)

Hence, reported three dependency vulnerabilities due to unmaintained crates.

- #436 
- #437 
- #438  
- #440
- #441

### DIFF
--- a/crates/pop-parachains/src/relay.rs
+++ b/crates/pop-parachains/src/relay.rs
@@ -5,7 +5,6 @@ use sp_core::twox_128;
 use subxt::{
 	config::BlockHash,
 	dynamic::{self, Value},
-	ext::sp_core,
 	OnlineClient, PolkadotConfig,
 };
 
@@ -91,7 +90,7 @@ impl RelayChain {
 #[cfg(test)]
 mod tests {
 	use super::*;
-	use subxt::ext::sp_core::twox_64;
+	use sp_core::twox_64;
 	use RelayChain::*;
 
 	#[test]
--- a/crates/pop-parachains/src/relay.rs
+++ b/crates/pop-parachains/src/relay.rs
@@ -5,6 +5,7 @@ use sp_core::twox_128;
 use subxt::{
 	config::BlockHash,
 	dynamic::{self, Value},
+	ext::sp_core,
 	OnlineClient, PolkadotConfig,
 };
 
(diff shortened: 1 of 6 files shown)


crates/pop-parachains/src/relay.rs
```diff
@@ -91,7 +90,7 @@ impl RelayChain {
 #[cfg(test)]
 mod tests {
 	use super::*;
-	use subxt::ext::sp_core::twox_64;
```
chungquantin, line 94: There was errors thrown for these crates. There was no `sp_core` in `subxt::ext`

AlexD10S:
The docker issue seems related to the `deny`, looking at the fix there https://github.com/EmbarkStudios/cargo-deny-action/pull/92/files and the rustup  latest release: https://blog.rust-lang.org/2025/03/02/Rustup-1.28.0.html#whats-new-in-rustup-1280

> rustup will no longer automatically install the active toolchain if it is not installed.
> To ensure its installation, run rustup toolchain install with no arguments.
> The following command installs the active toolchain both before and after this change:
> rustup show active-toolchain || rustup toolchain install
> 
It seems we have to include this line in the Dockerfile (Before `RUN cargo build --release`)
```
RUN rustup show active-toolchain || rustup toolchain install
```

You can the test docker builds locally: 
```
docker build -t pop .  --platform linux/amd64
```

AlexD10S (approved):
The removal of unmaintained dependencies is a hassle, but according to the docs: 

> If you want to ignore a specific advisory, add it to the ignore field

The approach you took makes sense. Approved.
But it would be good to check which libraries depend on the ones with vulnerabilities and see if we can eliminate them by upgrading.

crates/pop-parachains/src/relay.rs
chungquantin: This need to be updated when `zombienet` is upgraded to `0.2.25`, there was error thrown

chungquantin: Screenshot 2025-03-05 at 01 44 25

AlexD10S (approved):

PR #448, opened 2025-03-09
fix: hardcoded test failing in the CI

AlexD10S:
After the updates in the `pop-node` release, one test was failing in the CI (https://github.com/r0gue-io/pop-cli/actions/runs/13666853327/job/38359250240?pr=435) due to a hardcoded value for the node name, `pop-devnet` which is no longer applicable in the new release. The hardcoded value has been removed. 

Additionally, this update fixes cargo deny in the CI by bumping dependency versions. It was automatic done by the [dependabot](https://github.com/apps/dependabot) https://github.com/r0gue-io/pop-cli/pull/447. But deny was still failing so had to add a vulnerability in the `deny.toml`.

The PR introduces two new vulnerability dependency: 
- #451 
- #450

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1010,9 +1010,9 @@ dependencies = [
 
 [[package]]
 name = "cc"
-version = "1.2.4"
+version = "1.2.16"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


chungquantin (approved):
Thank you for the fix!

PR #449, opened 2025-03-09
chore: add password to the Cli

AlexD10S:
Add the new component password() to the `Cli` struct that enables prompt for values but masks the input. Needed for a upcoming PR.


chungquantin (approved):
Looks clean! I have no concern for this PR. 

Just a minor thing that I suggest having tests for `cli.rs` in the future so when we add new CLI features, it's easier to review.

PR #453, opened 2025-03-10
feat: `pop build --benchmark`

chungquantin:
Build a parachain with `runtime-benchmarks` feature. To test the PR, run the below command on any parachain project: 

```
pop build --benchmark
```
Or 
```
pop build parachain --benchmark
```


crates/pop-parachains/src/build.rs
```diff
@@ -508,19 +510,41 @@ mod tests {
 		Ok(())
 	}
 
+	fn add_feature(project: &Path, feature: &str) -> Result<()> {
```
AlexD10S, line 513: Why this? should not we used the `add_feature` you created in `pop-common/manifest.rs`?

crates/pop-common/src/manifest.rs
```diff
@@ -136,6 +136,27 @@ pub fn add_production_profile(project: &Path) -> anyhow::Result<()> {
 	Ok(())
 }
 
+/// Add a new feature to the Cargo.toml manifest if it doesn't already exist.
+///
+/// # Arguments
+/// * `project` - The path to the project directory.
+/// * `(key, items)` - The feature key and its associated items.
+pub fn add_feature(project: &Path, (key, items): (String, Vec<String>)) -> anyhow::Result<()> {
```
AlexD10S, line 144: This is only used for tests no? might be good to specify it with `#[cfg(test)]`

AlexD10S (commented):
Great job!! Fantastic to all the test coverage for the new functions!
I left a couple of comments regarding the `add_feature` function.

crates/pop-common/src/manifest.rs
chungquantin: Same as `add_production_profile` above, can't share between crates with test feature gated

crates/pop-parachains/src/build.rs
chungquantin: Removed in https://github.com/r0gue-io/pop-cli/pull/453/commits/d3bffbb08bb7cc12a820f509feac378d7d846bf6. Agree that tests can resuse the method in `pop-common`

AlexD10S (approved):
Thanks for fixing my comment so quickly!

PR #456, opened 2025-03-11
fix: resolve cargo deny and build script to fix the CI

chungquantin:
With the introduction of some new crates for #424, especially the `frame-benchmarking-cli` crate, there are couples of vulnerabilities and changes to the build script. 
- Install `protobuf-compiler` in the dockerfile and Github action script. 
- Add dependency vulnerabilities to `deny.toml`: https://github.com/r0gue-io/pop-cli/issues/458


AlexD10S (approved):
Approved!

PR #464, opened 2025-03-14
fix: excluded arguments

chungquantin:
PR to exclude custom arguments like `--profile` or `--skip` from `machine`, `block` and `storage`. This was missing from the implementation PR.


AlexD10S (changes requested):
Unit tests are failing.

Not a blocker, but the same code is repeated three times in storage, machine, and block. Wouldn't it be better to consolidate it in `generate_binary_benchmarks` instead?

PR #465, opened 2025-03-14
feat: add select profile in pallet benchmarking

chungquantin:
Guide user to select profile before building the runtime.


AlexD10S (approved):
LGTM!

PR #468, opened 2025-03-17
feat: overwrite executed command in weight files

chungquantin:
Overwrites the `// Executed Command:` comment block of the generated weight files. As we run the command with `omni-bencher` or `frame-benchmarking-cli`, the generated command will be something like: 

```
// Executed Command: 
//  /user/cache/pop/frame-omni-bencher
// v1
// benchmark
// pallet
// --pallet=
// --extrinsic
```

The PR will overwrites the `Executed Command` with `pop bench command`: 

```
// Executed Command: 
//  pop
//  bench
//  pallet
//  --pallet=
//  --extrinsic=
```

Another issue is with the interactive UI, we prompt user to update the command arguments. With the current weight file, it does not display the updated arguments. 

[sc-2888]

### DIFF
--- a/crates/pop-cli/src/commands/bench/pallet.rs
+++ b/crates/pop-cli/src/commands/bench/pallet.rs
@@ -9,6 +9,7 @@ use crate::{
 		bench::{
 			check_omni_bencher_and_prompt, ensure_runtime_binary_exists,
 			guide_user_to_select_genesis_policy, guide_user_to_select_genesis_preset,
+			overwrite_weight_file_command,
 		},
 		builds::guide_user_to_select_profile,
 		prompt::display_message,
@@ -28,6 +29,7 @@ use std::{
 };
 use strum::{EnumMessage, IntoEnumIterator};
 use strum_macros::{EnumIter, EnumMessage as EnumMessageDerive};
+use tempfile::tempdir;
 
 const ALL_SELECTED: &str = "*";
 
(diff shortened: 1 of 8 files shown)


crates/pop-parachains/src/bench/mod.rs
```diff
@@ -121,12 +121,20 @@ pub fn generate_pallet_benchmarks(args: Vec<String>) -> anyhow::Result<()> {
 /// # Arguments
 /// * `binary_path` - Path to the binary of FRAME Omni Bencher.
 /// * `command` - Command to run for benchmarking.
-pub fn generate_binary_benchmarks(binary_path: &PathBuf, command: &str) -> anyhow::Result<()> {
+/// * `update_args` - Function to update the arguments before running the benchmark.
+pub fn generate_binary_benchmarks<F>(
+	binary_path: &PathBuf,
+	command: &str,
+	update_args: F,
```
AlexD10S, line 128: Nice

crates/pop-cli/src/common/bench.rs
```diff
@@ -258,6 +260,76 @@ pub(crate) fn get_mock_runtime(with_benchmark_features: bool) -> PathBuf {
...
+		let destination = dest_path.join(path.file_name().unwrap());
+		overwrite_weight_file_command(&path, destination.as_path(), arguments)?;
+	}
+	Ok(())
+}
+
+/// Overwrites the weight file's executed command with the given arguments.
+///
+/// # Arguments
+/// * `temp_file` - The path to the temporary.
```
AlexD10S, line 290:
```suggestion
/// * `temp_file` - The path to the temporary file.
```

crates/pop-cli/src/commands/bench/pallet.rs
```diff
@@ -342,18 +344,43 @@ impl BenchmarkPallet {
 		Ok(())
 	}
 
-	fn run(&self) -> anyhow::Result<()> {
-		generate_pallet_benchmarks(self.collect_arguments())
+	fn run(&mut self) -> anyhow::Result<()> {
+		match self.output.clone() {
```
AlexD10S, line 348: Up to you, but might be cleaner this?
```
if let Some(original_weight_path) = self.output.clone() {
..
else {
	generate_pallet_benchmarks(self.collect_arguments())?;
};
```

crates/pop-parachains/src/bench/mod.rs
```diff
@@ -121,12 +121,20 @@ pub fn generate_pallet_benchmarks(args: Vec<String>) -> anyhow::Result<()> {
 /// # Arguments
 /// * `binary_path` - Path to the binary of FRAME Omni Bencher.
 /// * `command` - Command to run for benchmarking.
-pub fn generate_binary_benchmarks(binary_path: &PathBuf, command: &str) -> anyhow::Result<()> {
+/// * `update_args` - Function to update the arguments before running the benchmark.
+pub fn generate_binary_benchmarks<F>(
+	binary_path: &PathBuf,
+	command: &str,
```
AlexD10S, line 127: Small nitpick, this can be a `OmniBencherCommand`?

crates/pop-cli/src/common/bench.rs
```diff
@@ -258,6 +260,76 @@ pub(crate) fn get_mock_runtime(with_benchmark_features: bool) -> PathBuf {
...
+		if line.starts_with(EXECUTED_COMMAND_COMMENT) {
+			inside_command_block = true;
+			continue;
+		} else if inside_command_block {
+			if line.starts_with("//") {
+				continue;
+			} else if line.trim().is_empty() {
+				// Write new command block to the generated weight file.
+				new_lines.push(EXECUTED_COMMAND_COMMENT.to_string());
+				for argument in arguments {
```
AlexD10S, line 314: Quick question: Is it better to display the command on separate lines? I assume it's for readability, especially with long arguments, right?

AlexD10S (approved):
Approved! Great job and solid tests!
Left a few minor nitpicks and questions.

crates/pop-cli/src/common/bench.rs
chungquantin: Yes, original implementation also display commands on separate lines so I follow that. In my opinion, it also enhances readability

crates/pop-cli/src/common/bench.rs
chungquantin: Resolved in https://github.com/r0gue-io/pop-cli/pull/468/commits/945f0bd57bcde8e2d78dcb0c5b381333e7734c65

PR #470, opened 2025-03-18
feat: save parameters of `pop bench pallet` to `.bench` TOML file

chungquantin:
## Description

Introduces the benchmarking parameters TOML file [default value: `pop-bench.toml`]. The benchmark pallet struct is versioned by introducing new struct `VersionedBenchmarkPallet`. If there are further versions of `.bench` file introduced in the future, we parse the current `bench.toml` by provided `version` in the file with the matched struct `BenchmarkPalletV_`. Then convert the `BenchmarkPalletV_` to the current `BenchmarkPallet` struct. 

Before running benchmark, we prompt user to update the bench file path: 

```
│
◇  Do you want to save the updated parameters?
│  Yes
│
◇  Provide the output path for benchmark parameters
│  pop-bench.toml
│
⚙  Parameters saved successfully to "pop-bench.toml"
│
▲  NOTE: this may take some time...
```
If the user provide a bench file with `-f` or `--bench-file`, we read the bench file and load parameters from the file. 

```
➜ pop bench pallet -f pop-bench.toml
┌   Pop CLI : Benchmarking your pallets
│
▲  NOTE: the `pop bench pallet` is not yet battle tested - double check the results.
│
⚙  Benchmarking parameter file found at "pop-bench.toml". Loading parameters...
│
◆  Do you want to open the parameter menu?
│  ○ Yes  / ● No
└
```

## How to test the command? 

Reuse the existing bench file with `-f` flag. 
```
pop bench pallet -f pop-bench.toml
```

## Outcome
Example output of the benchmarking parameters file: 
```toml
version = "1"
pallet = "pallet_timestamp"
extrinsic = "*"
exclude_pallets = []
all = false
steps = 50
lowest_range_values = []
highest_range_values = []
repeat = 20
external_repeat = 1
json_output = false
no_median_slopes = false
no_min_squares = false
output_pov_analysis = "median-slopes"
no_verify = false
extra = false
runtime = "/Users/chungquantin/Developer/pop-network/pop-cli/tests/runtimes/base_parachain_benchmark.wasm"
allow_missing_host_functions = false
genesis_builder = "Runtime"
genesis_builder_preset = "development"
database_cache_size = 1024
list = false
no_storage_info = false
worst_case_map_values = 1000000
additional_trie_layers = 2
disable_proof_recording = false
skip_menu = false
skip_confirm = false
```


crates/pop-cli/src/commands/bench/pallet.rs
```diff
@@ -825,6 +855,53 @@ impl BenchmarkPalletMenuOption {
...
+			VersionedBenchmarkPallet::V1(parameters) => parameters.clone(),
+		}
+	}
+}
+
+impl TryFrom<&Path> for VersionedBenchmarkPallet {
+	type Error = anyhow::Error;
+
+	fn try_from(bench_file: &Path) -> anyhow::Result<Self> {
+		if !bench_file.exists() || bench_file.is_dir() {
```
AlexD10S, line 879: Minor improvement: I believe `!bench_file.is_file()` is enough because it checks if exists too

crates/pop-cli/src/commands/bench/pallet.rs
```diff
@@ -330,6 +351,15 @@ impl BenchmarkPallet {
 			self.output = if !input.is_empty() { Some(input.into()) } else { None };
 		}
 
+		// Prompt user to save benchmarking parameters to output file if there are changes made.
+		if let Some(bench_file) =
+			guide_user_to_update_bench_file_path(self, cli, original_cmd != self.clone())?
```
AlexD10S, line 356:
```suggestion
			guide_user_to_update_bench_file_path(self, cli, self != &original_cmd)?
```

AlexD10S (approved):
Great job on this, Tin! 
In my review, I noticed quite a few `.clone()`, but most seemed necessary, so I only left two small comments.
Is cool having all the parameters in a file, can be great feature for script! Also, the DevEx of detecting changes and prompting to overwrite the file is a great touch.

crates/pop-cli/src/commands/bench/pallet.rs
chungquantin: Resolved in https://github.com/r0gue-io/pop-cli/pull/470/commits/57c8256a57407bd4a87e91ee6a3853e22329c07c

crates/pop-cli/src/commands/bench/pallet.rs
chungquantin: Resolved in https://github.com/r0gue-io/pop-cli/pull/470/commits/57c8256a57407bd4a87e91ee6a3853e22329c07c

PR #472, opened 2025-03-19
fix: excluded custom arguments

chungquantin:
Exclude custom arguments from the actual command arguments run with `frame-omni-bencher` or node binary.


AlexD10S (approved):
LGTM! The only issue before was CI, and it's green now. 🚀

PR #473, opened 2025-03-19
feat: update template & optimize printed command

chungquantin:
- Add method to update the weight file template based on the feedback on #424 
- Optimize the printed command to only display arguments which are different from the default values. 

### Outcome
```
// Executed Command:
//  pop
//  bench
//  pallet
//  --pallet=pallet_timestamp
//  --extrinsic=
//  --steps=50
//  --output=./weights.rs
//  --runtime=/Users/chungquantin/Developer/pop-network/pop-cli/tests/runtimes/base_parachain_benchmark.wasm
//  --genesis-builder=runtime
```

### DIFF
--- a/crates/pop-cli/src/commands/bench/pallet.rs
+++ b/crates/pop-cli/src/commands/bench/pallet.rs
@@ -36,6 +36,7 @@ use tempfile::tempdir;
 
 const ALL_SELECTED: &str = "*";
 const DEFAULT_BENCH_FILE: &str = "pop-bench.toml";
+const ARGUMENT_NO_VALUE: &str = "None";
 
 #[derive(Args, Serialize, Deserialize, Clone, PartialEq, Eq, Debug)]
 pub(crate) struct BenchmarkPallet {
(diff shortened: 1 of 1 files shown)


crates/pop-cli/src/commands/bench/pallet.rs
```diff
@@ -590,6 +609,19 @@ impl BenchmarkPallet {
 		Ok(())
 	}
 
+	fn update_template_path(&mut self, cli: &mut impl cli::traits::Cli) -> anyhow::Result<()> {
```
AlexD10S, line 613: I see this doesn't have unit test coverage: https://app.codecov.io/gh/r0gue-io/pop-cli/blob/chungquantin%2Ffeat-template_path/crates%2Fpop-cli%2Fsrc%2Fcommands%2Fbench%2Fpallet.rs#L612
This might be a simple test to add.

AlexD10S (commented):
Nice improvement!
The `update_template_path` function is not covered (https://app.codecov.io/gh/r0gue-io/pop-cli/blob/chungquantin%2Ffeat-template_path/crates%2Fpop-cli%2Fsrc%2Fcommands%2Fbench%2Fpallet.rs#L612) If you could add a test for it, that would be great.

crates/pop-cli/src/commands/bench/pallet.rs
chungquantin: Resolved in https://github.com/r0gue-io/pop-cli/pull/473/commits/0d892b8df4424f74c5285405903c98b3a7e88ee3

PR #474, opened 2025-03-19
fix: `fungible` typo in `pop new parachain` command

franciscoaguirre:
When picking templates, there was a typo in `fungible`.


chungquantin (approved):
Thanks for fixing this!

PR #476, opened 2025-03-20
feat: build with try-runtime feature enabled

chungquantin:
Allows user to specify features to run `pop build`.

## How to test? 

```
pop build --features try-runtime
```
or 
```
pop build --try-runtime
```
or
```
pop build parachain --try-runtime
```


crates/pop-cli/src/commands/build/mod.rs
```diff
@@ -115,13 +133,22 @@ impl Command {
...
+		let features = args.features.unwrap_or_default();
+		let feature_arg = format!("--features={}", features);
+		if !features.is_empty() {
+			_args.push(&feature_arg);
 		}
 
 		cmd("cargo", _args).dir(args.path.unwrap_or_else(|| "./".into())).run()?;
 
-		cli.info(format!("The {project} was built in {} mode.", profile))?;
+		if features.is_empty() {
```
AlexD10S, line 144: Nice message! No change required, but if you prefer, this can be a one line
```
cli.info(format!("The {project} was built in {profile}{}.", if features.is_empty() { String::new() } else { format!(" with the following features: {}", features) }))?;
```

crates/pop-cli/src/commands/build/mod.rs
```diff
@@ -41,9 +48,17 @@ pub(crate) struct BuildArgs {
 	/// Build profile [default: debug].
 	#[clap(long, value_enum)]
 	pub(crate) profile: Option<Profile>,
-	/// For benchmarking, always build with `runtime-benchmarks` feature.
+	/// List of features that project is built with, separated by commas.
+	///
+	/// Common features:
```
AlexD10S, line 53: Would it make sense to keep it as you had it before to make it easier to specify the most common features while still allowing additional ones via features?
For example
```
/// For benchmarking, always build with `runtime-benchmarks` feature.
#[clap(short, long)]
pub(crate) benchmark: bool,
/// For `try-runtime`, alway build with `try-runtime` feature.
#[clap(short, long)]
pub(crate) try_runtime: bool,
/// Other list of features that project is built with, separated by commas.
#[clap(short, long)]
pub(crate) features: Option<String>,
```
The `--help` output
```
Options:
      --path <PATH>          Directory path with flag for your project [default: current directory]
  -p, --package <PACKAGE>    The package to be built
  -r, --release              For production, always build in release mode to exclude debug features
      --profile <PROFILE>    Build profile [default: debug]
  -b, --benchmark            For benchmarking, always build with `runtime-benchmarks` feature
  -t, --try-runtime          Whether to build the parachain with `try-runtime` feature
  -f, --features <FEATURES>  Other list of features that project is built with, separated by commas
  -h, --help                 Print help
```

Alternatively, if you want to group parachain-specific flags you can use `help_heading` `#[clap(short, long, help_heading = "Chain options")]`

AlexD10S (commented):
I left a question. It might make the `--help` output clearer for users, but I'd like to hear your perspective on it.

I also think it might be worth pointing this change directly to `main`

crates/pop-cli/src/commands/build/mod.rs
chungquantin: Resolved in https://github.com/r0gue-io/pop-cli/pull/476/commits/71d3c61a5bff45673f9ceb4a5bad014c5605b4ba

crates/pop-cli/src/commands/build/mod.rs
chungquantin: Resolved in https://github.com/r0gue-io/pop-cli/pull/476/commits/71d3c61a5bff45673f9ceb4a5bad014c5605b4ba

chungquantin:
> I left a question. It might make the `--help` output clearer for users, but I'd like to hear your perspective on it.
> 
> I also think it might be worth pointing this change directly to `main`

This PR reuse couple of features from `pop bench` like `pop build --benchmark` so it is currently pointing to the fork of #424

AlexD10S (approved):
Nice and clean!

Issue #475, opened 2025-03-19
Improve error messaging when building specs fails


AlexD10S:
The current error message when building specs fails is unclear and does not provide enough context for debugging, we should improve this messaging to improve developer experience and troubleshooting.

**How to replicate it**

If I run `pop build spec --chain ./wrong-chain-spec` I get this error:
```
Error: IO error: command ["/templates_test/evm/target/release/parachain-template-node", "build-spec", "--chain", "./wrong-chain-spec", "--disable-default-bootnode"] exited with code 1

Caused by:
    command ["templates_test/evm/target/release/parachain-template-node", "build-spec", "--chain", "./wrong-chain-spec", "--disable-default-bootnode"] exited with code 1
```
While If I run 
```
./target/release/parachain-template-node build-spec --chain ./wrong-chain-spec
```

I get a more clear one:
```
Error: Input("Error opening spec file `./c`: No such file or directory (os error 2)")
```

PR #477, opened 2025-03-20
fix: improve build spec error messaging

AlexD10S:
This PR improves the error message when building specs fails, making it clearer and more aligned with the output of the underlying command.

**Before:**

Running `pop build spec --chain ./wrong-chain-spec`:
```
Error: IO error: command ["/templates_test/evm/target/release/parachain-template-node", "build-spec", "--chain", "./wrong-chain-spec", "--disable-default-bootnode"] exited with code 1

Caused by:
    command ["templates_test/evm/target/release/parachain-template-node", "build-spec", "--chain", "./wrong-chain-spec", "--disable-default-bootnode"] exited with code 1
```
This error message was not very informative and did not clearly indicate the root cause of the failure.
**Now:**
```
 Failed to build the chain spec. Error: Input("Error opening spec file `./wrong-chain-spec`: No such file or directory (os error 2)")
```
This message explicitly states the failure reason and matches the output of the underlying command:
```
./target/release/parachain-template-node build-spec --chain ./wrong-chain-spec
```
```
Error: Input("Error opening spec file `./c`: No such file or directory (os error 2)")
```

Closes https://github.com/r0gue-io/pop-cli/issues/475

[sc-3371]

### DIFF
--- a/crates/pop-parachains/src/build/mod.rs
+++ b/crates/pop-parachains/src/build/mod.rs
@@ -97,7 +97,16 @@ pub fn generate_plain_chain_spec(
 	// Create a temporary file.
 	let temp_file = tempfile::NamedTempFile::new_in(std::env::temp_dir())?;
 	// Run the command and redirect output to the temporary file.
-	cmd(binary_path, args).stdout_path(temp_file.path()).stderr_null().run()?;
+	let output = cmd(binary_path, args)
+		.stdout_path(temp_file.path())
+		.stderr_capture()
+		.unchecked()
+		.run()?;
+	// Check if the command failed.
+	if !output.status.success() {
+		let stderr_msg = String::from_utf8_lossy(&output.stderr);
+		return Err(Error::BuildSpecError(stderr_msg.to_string()));
+	};
 	// Atomically replace the chain spec file with the temporary file.
 	temp_file.persist(plain_chain_spec).map_err(|e| {
 		AnyhowError(anyhow!(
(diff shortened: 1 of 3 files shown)


crates/pop-parachains/src/build/mod.rs
```diff
@@ -215,6 +227,15 @@ fn check_command_exists(binary_path: &Path, command: &str) -> Result<(), Error>
 	Ok(())
 }
 
+// Handles command execution errors by extracting and returning the stderr message.
+fn handle_command_error(output: &std::process::Output) -> Result<(), Error> {
```
chungquantin, line 231: I think this method is not necessary, I already handled the error returned from `cmd` in the benchmarking feature: https://github.com/r0gue-io/pop-cli/pull/424/files#diff-da93d4ed174e40b1698b8143d236c9670cbe245017ec777f112ba7ebfb9d939cR168

I got the same issue that whenever errors thrown, it throws the status code instead of the actual error. And the code in my PR already handles that. Maybe we can reuse it? Worth creating an issue to provide our own method `cmd` that returns better error so we can reuse in the future

crates/pop-parachains/src/build/mod.rs
```diff
@@ -199,8 +209,10 @@ pub fn generate_genesis_state_file(
 		],
 	)
 	.stdout_null()
-	.stderr_null()
+	.stderr_capture()
```
chungquantin, line 212: With my comment below, I don't think we need to capture the error this way.

chungquantin (changes requested):
Seems this PR's issue is already handled in the benchmarking. As benchmarking is not merged to main yet, my suggestion is to have a common method that can be used for both features and then I'll refactor the benchmarking feature after it is merged to main.

crates/pop-parachains/src/build/mod.rs
AlexD10S: Nice, I'd like to keep it consistent, let me review the code in benchmark and make the necessary adjustments.

AlexD10S: Moved the logic into a shared `errors` module in the `pop_parachain` crate to allow reuse across the crate. This makes it possible to return custom errors depending on the context, `BuildSpecError` in this case, but any custom error can be specified as needed via a provided error constructor https://github.com/r0gue-io/pop-cli/pull/477/commits/d42cb8107b9414bfa63925a8b3ce2942fac72cf8

crates/pop-parachains/src/build/mod.rs
```diff
@@ -97,10 +97,16 @@ pub fn generate_plain_chain_spec(
 	// Create a temporary file.
 	let temp_file = tempfile::NamedTempFile::new_in(std::env::temp_dir())?;
 	// Run the command and redirect output to the temporary file.
-	cmd(binary_path, args).stdout_path(temp_file.path()).stderr_null().run()?;
+	let output = cmd(binary_path, args)
```
chungquantin, line 100: @AlexD10S Curious why don't we have a shared method like `generate_binary_benchmarks` in #424 so we don't have to call `handle_command_error` and `stderr_capture()`. Like our custom `cmd()` that can be used whenever we want a beautiful error output

crates/pop-parachains/src/build/mod.rs
AlexD10S: Same as above

crates/pop-parachains/src/build/mod.rs
AlexD10S: The reason I didn’t do it that way is because the `stdout_path` varies depending on the case. I suppose you could still handle it with an if inside the helper method, but I’m not sure it would actually make things much cleaner.

crates/pop-parachains/src/build/mod.rs
chungquantin: Thank you

chungquantin (approved):
Looks great!

Issue #444, opened 2025-03-05
fix: cli prompt for bootnode even if a chain spec is provided and no changes desired.


Daanvdplas:
<img width="1016" alt="Image" src="https://github.com/user-attachments/assets/5b2e672a-ca27-4e3b-a346-71fb3e4253fa" />

I provide a chainspec and I don't want to make changes, why is it still asked of me if I want to add a bootnode?

PR #482, opened 2025-03-21
fix(build spec): default bootnode prompt

Daanvdplas:
When the user doesn't want to make changes to the spec that it provides, no prompt for the default bootnode.

Closes https://github.com/r0gue-io/pop-cli/issues/444


AlexD10S (approved):
Thanks for fixing this!

PR #483, opened 2025-03-23
chore: improve pop up flow

AlexD10S:
This PR introduces several UX enhancements and minor improvements gathered from the recent demo.

**Changes:**
-  Pause warning messages briefly for clarity.
- Added direct links to the Portal provider at various points in the flow.
- Improved messaging, when fetching collator keys using a spinner.
-  Prefixed `r0gue` templates for easier identification.
-  Added a `--profile` flag (default: Release) allowing users to specify the profile used when building the chain spec, reducing interactive prompts.
- User guidance: Included numbered steps throughout the flow to better assist first-time users.

flow

### DIFF
--- a/crates/pop-cli/src/cli.rs
+++ b/crates/pop-cli/src/cli.rs
@@ -150,7 +150,7 @@ impl traits::Cli for Cli {
 	fn warning(&mut self, message: impl Display) -> Result<()> {
 		cliclack::log::warning(message)?;
 		#[cfg(not(test))]
-		sleep(Duration::from_secs(2));
+		sleep(Duration::from_secs(1));
 		Ok(())
 	}
 }
--- a/crates/pop-cli/src/cli.rs
+++ b/crates/pop-cli/src/cli.rs
@@ -1,6 +1,8 @@
 // SPDX-License-Identifier: GPL-3.0
 
 use std::{fmt::Display, io::Result};
+#[cfg(not(test))]
+use std::{thread::sleep, time::Duration};
 #[cfg(test)]
 pub(crate) use tests::MockCli;
 
@@ -146,7 +148,10 @@ impl traits::Cli for Cli {
 
 	/// Prints a warning message.
 	fn warning(&mut self, message: impl Display) -> Result<()> {
-		cliclack::log::warning(message)
+		cliclack::log::warning(message)?;
+		#[cfg(not(test))]
+		sleep(Duration::from_secs(2));
+		Ok(())
 	}
 }
 
--- a/crates/pop-cli/src/commands/build/parachain.rs
+++ b/crates/pop-cli/src/commands/build/parachain.rs
@@ -4,8 +4,6 @@ use crate::{cli, style::style};
 use pop_common::Profile;
 use pop_parachains::build_parachain;
 use std::path::PathBuf;
-#[cfg(not(test))]
-use std::{thread::sleep, time::Duration};
 
 // Configuration for building a parachain.
 pub struct BuildParachain {
@@ -33,8 +31,6 @@ impl BuildParachain {
 
 		if self.profile == Profile::Debug {
 			cli.warning("NOTE: this command now defaults to DEBUG builds. Please use `--release` (or simply `-r`) for a release build...")?;
-			#[cfg(not(test))]
-			sleep(Duration::from_secs(3))
 		}
 
 		// Build parachain.
--- a/crates/pop-cli/src/commands/build/spec.rs
+++ b/crates/pop-cli/src/commands/build/spec.rs
@@ -21,8 +21,6 @@ use std::{
 	fs::{self, create_dir_all},
 	path::{Path, PathBuf},
 };
-#[cfg(not(test))]
-use std::{thread::sleep, time::Duration};
 use strum::{EnumMessage, VariantArray};
 use strum_macros::{AsRefStr, Display, EnumString};
 
@@ -488,8 +486,6 @@ impl BuildSpecCommand {
 
 		if release {
 			cli.warning("NOTE: release flag is deprecated. Use `--profile` instead.")?;
-			#[cfg(not(test))]
-			sleep(Duration::from_secs(3));
 			profile = Profile::Release;
 		}
 
--- a/crates/pop-cli/src/commands/up/mod.rs
+++ b/crates/pop-cli/src/commands/up/mod.rs
@@ -168,7 +168,15 @@ mod tests {
 			Some(
 				DeploymentProvider::VARIANTS
 					.into_iter()
-					.map(|action| (action.name().to_string(), format!("{}", style(format!("{}", action.base_url())).bold().underlined())))
+					.map(|action| {
+						(
+							action.name().to_string(),
+							format!(
+								"{}",
+								style(format!("{}", action.base_url())).bold().underlined()
+							),
+						)
+					})
 					.chain(std::iter::once((
 						"Register".to_string(),
 						"Register the rollup on the relay chain without deploying with a provider"
--- a/crates/pop-cli/src/commands/up/mod.rs
+++ b/crates/pop-cli/src/commands/up/mod.rs
@@ -94,7 +94,7 @@ impl Command {
 #[cfg(test)]
 mod tests {
 	use super::{contract::UpContractCommand, *};
-	use crate::style::style;
+	use crate::style::{format_url, style};
 	use cli::MockCli;
 	use duct::cmd;
 	use pop_contracts::{mock_build_process, new_environment};
@@ -168,12 +168,7 @@ mod tests {
 			Some(
 				DeploymentProvider::VARIANTS
 					.into_iter()
-					.map(|action| {
-						(
-							action.name().to_string(),
-							format!("{}", style(action.base_url().to_string()).bold().underlined()),
-						)
-					})
+					.map(|action| (action.name().to_string(), format_url(action.base_url())))
 					.chain(std::iter::once((
 						"Register".to_string(),
 						"Register the rollup on the relay chain without deploying with a provider"
--- a/crates/pop-cli/src/commands/up/mod.rs
+++ b/crates/pop-cli/src/commands/up/mod.rs
@@ -94,7 +94,7 @@ impl Command {
 #[cfg(test)]
 mod tests {
 	use super::{contract::UpContractCommand, *};
-
+	use crate::style::style;
 	use cli::MockCli;
 	use duct::cmd;
 	use pop_contracts::{mock_build_process, new_environment};
@@ -168,7 +168,7 @@ mod tests {
 			Some(
 				DeploymentProvider::VARIANTS
 					.into_iter()
-					.map(|action| (action.name().to_string(), action.description().to_string()))
+					.map(|action| (action.name().to_string(), format!("{}", style(format!("{}", action.base_url())).bold().underlined())))
 					.chain(std::iter::once((
 						"Register".to_string(),
 						"Register the rollup on the relay chain without deploying with a provider"
--- a/crates/pop-cli/src/commands/up/mod.rs
+++ b/crates/pop-cli/src/commands/up/mod.rs
@@ -171,10 +171,7 @@ mod tests {
 					.map(|action| {
 						(
 							action.name().to_string(),
-							format!(
-								"{}",
-								style(format!("{}", action.base_url())).bold().underlined()
-							),
+							format!("{}", style(action.base_url().to_string()).bold().underlined()),
 						)
 					})
 					.chain(std::iter::once((
--- a/crates/pop-cli/src/commands/up/rollup.rs
+++ b/crates/pop-cli/src/commands/up/rollup.rs
@@ -11,7 +11,7 @@ use crate::{
 		wallet::submit_extrinsic,
 	},
 	deployment_api::{DeployRequest, DeployResponse, DeploymentApi},
-	style::style,
+	style::{format_url, style},
 };
 use anyhow::Result;
 use clap::Args;
@@ -451,7 +451,7 @@ fn prompt_provider(
 		predefined_action = predefined_action.item(
 			Some(action.clone()),
 			action.name(),
-			format!("{}", style(action.base_url().to_string()).bold().underlined()),
+			format_url(action.base_url()),
 		);
 	}
 	if !skip_registration {
@@ -619,12 +619,7 @@ mod tests {
 			Some(
 				DeploymentProvider::VARIANTS
 					.into_iter()
-					.map(|action| {
-						(
-							action.name().to_string(),
-							format!("{}", style(action.base_url().to_string()).bold().underlined()),
-						)
-					})
+					.map(|action| (action.name().to_string(), format_url(action.base_url())))
 					.chain(std::iter::once((
 						"Register".to_string(),
 						"Register the rollup on the relay chain without deploying with a provider"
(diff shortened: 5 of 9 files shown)


crates/pop-cli/src/commands/build/parachain.rs
```diff
@@ -33,8 +31,6 @@ impl BuildParachain {
 
 		if self.profile == Profile::Debug {
 			cli.warning("NOTE: this command now defaults to DEBUG builds. Please use `--release` (or simply `-r`) for a release build...")?;
-			#[cfg(not(test))]
```
AlexD10S, line 36: Removed because `cli.warning` will manage the sleep for consistency

crates/pop-cli/src/commands/build/spec.rs
```diff
@@ -486,8 +484,6 @@ impl BuildSpecCommand {
 
 		if release {
 			cli.warning("NOTE: release flag is deprecated. Use `--profile` instead.")?;
-			#[cfg(not(test))]
```
AlexD10S, line 489: Same as above

crates/pop-cli/src/commands/up/mod.rs
```diff
@@ -168,7 +168,15 @@ mod tests {
 			Some(
 				DeploymentProvider::VARIANTS
 					.into_iter()
-					.map(|action| (action.name().to_string(), action.description().to_string()))
+					.map(|action| {
+						(
+							action.name().to_string(),
+							format!(
+								"{}",
+								style(format!("{}", action.base_url())).bold().underlined()
```
AlexD10S, line 176: The combination of description and URL didn’t look great (too long). Showing only the URL might be more helpful to the user.

crates/pop-cli/src/cli.rs
```diff
@@ -146,7 +148,10 @@ impl traits::Cli for Cli {
 
 	/// Prints a warning message.
 	fn warning(&mut self, message: impl Display) -> Result<()> {
-		cliclack::log::warning(message)
+		cliclack::log::warning(message)?;
+		#[cfg(not(test))]
+		sleep(Duration::from_secs(2));
```
chungquantin, line 153: Is it better to have a parameter `ms: u32` in the `warning` method? I feel like 2 second is a bit long, with that, it will be more flexible...

crates/pop-cli/src/commands/up/mod.rs
```diff
@@ -168,7 +168,12 @@ mod tests {
 			Some(
 				DeploymentProvider::VARIANTS
 					.into_iter()
-					.map(|action| (action.name().to_string(), action.description().to_string()))
+					.map(|action| {
+						(
+							action.name().to_string(),
+							format!("{}", style(action.base_url().to_string()).bold().underlined()),
```
chungquantin, line 174: Not related to this PR but if this logic can be in a separate function `format_url`, we can also use it for other URLs in the future. e.g. in benchmarking as well

crates/pop-cli/src/cli.rs
chungquantin, line 1: Screenshot 2025-03-25 at 13 15 51

From your GIF, I feel like now we're having too many texts on the terminals. Not sure what is the best way to improve it but we maybe for less important texts like "NOTE:...", we can use `dim()` to darken the text color. 

Screenshot 2025-03-25 at 13 18 13

This is what I did in the benchmarking feature.

crates/pop-cli/src/commands/up/rollup.rs
```diff
@@ -209,16 +232,27 @@ impl UpCommand {
 				..Default::default()
 			});
 		}
-		cli.info("Generating the chain spec for your project")?;
+		cli.info(format!(
+			"{}Generating the chain spec for your project",
+			if show_deployment_steps { "Step 3/5: " } else { "" }
```
chungquantin, line 237:
```suggestion
			if show_deployment_steps { "[3/5]: " } else { "" }
```

chungquantin: Will this break the step indices if I provide the chain-spec? This step will be skipped if I am correct then this should be `[2/5]` instead of `[3/5]`. This approach is not easy to maintain from my opinion. My suggestion is to have a counter at the start of the session and then whenever steps reached, we increment the counter. 

- Maybe having a method called `move_next_step()` and if `step > 0` and `step < MAX_STEPS`, we will display `[step / MAX_STEP]` instead of hard-coding the value. 
- Instead of checking manually `if show_deployment_steps { "Step 3/5: " } else { "" }`, would be better if we refactor to a separate method `print_with_step(content: &str, index: usize)`. 

As this logic will be probably used by future implementation, even current pending PR like #424 can also reuse this logic so if we can make it easy to integrate, that would be perfect!

crates/pop-cli/src/commands/up/rollup.rs
```diff
@@ -246,7 +285,11 @@ impl Deployment {
 			let template_name = prompt_template_used(cli)?;
 			request.runtime_template = Some(template_name.to_string());
 		}
-		cli.info(format!("Starting deployment with {}", api.provider.name()))?;
+		cli.info(format!(
+			"{}Starting deployment with {}",
+			if show_deployment_steps { "Step 5/5: " } else { "" },
```
chungquantin, line 290: `print_with_step`

crates/pop-cli/src/commands/up/rollup.rs
```diff
@@ -260,8 +303,8 @@ struct Registration {
 }
 impl Registration {
 	// Registers by submitting an extrinsic.
-	async fn register(&self, cli: &mut impl Cli) -> Result<()> {
-		cli.info(format!("You will need to sign a transaction to register on {}, using the `Registrar::register` function.", self.chain.url))?;
+	async fn register(&self, show_deployment_steps: bool, cli: &mut impl Cli) -> Result<()> {
+		cli.info(format!("{}You will need to sign a transaction to register on {}, using the `Registrar::register` function.", if show_deployment_steps { "Step 4/5: " } else {""}, self.chain.url))?;
```
chungquantin, line 307: `print_with_step`

crates/pop-cli/src/commands/up/rollup.rs
```diff
@@ -401,8 +448,11 @@ fn prompt_provider(
 ) -> Result<Option<DeploymentProvider>> {
 	let mut predefined_action = cli.select("Select your deployment method:");
 	for action in DeploymentProvider::VARIANTS {
-		predefined_action =
-			predefined_action.item(Some(action.clone()), action.name(), action.description());
+		predefined_action = predefined_action.item(
+			Some(action.clone()),
+			action.name(),
+			format!("{}", style(action.base_url().to_string()).bold().underlined()),
```
chungquantin, line 454: `format_url`

crates/pop-cli/src/commands/up/rollup.rs
```diff
@@ -509,7 +563,15 @@ mod tests {
 				Some(
 					DeploymentProvider::VARIANTS
 						.into_iter()
-						.map(|action| (action.name().to_string(), action.description().to_string()))
+						.map(|action| {
+							(
+								action.name().to_string(),
+								format!(
```
chungquantin, line 569:
```suggestion
								style(action.base_url().to_string()).bold().underlined().to_string()
```

crates/pop-cli/src/commands/up/rollup.rs
```diff
@@ -668,7 +744,7 @@ mod tests {
                 Some(
                     DeploymentProvider::VARIANTS
                         .into_iter()
-                        .map(|action| (action.name().to_string(), action.description().to_string()))
+                        .map(|action| (action.name().to_string(), format!("{}", style(action.base_url().to_string()).bold().underlined())))
```
chungquantin, line 747:
```suggestion
                        .map(|action| (action.name().to_string(), format!("{}", format_url(action.base_url()))))
```

crates/pop-cli/src/commands/up/rollup.rs
```diff
@@ -707,7 +783,7 @@ mod tests {
...
```
chungquantin, line 786:
```suggestion
                        .map(|action| (action.name().to_string(), format!("{}", format_url(action.base_url()))))
```

crates/pop-cli/src/commands/up/rollup.rs
```diff
@@ -852,6 +929,27 @@ mod tests {
...
+		let deployment = Deployment {
+			api: Some(DeploymentApi::new(
+				"api_test_key".to_string(),
+				DeploymentProvider::PDP,
+				"PASEO".to_string(),
+			)?),
+			..Default::default()
+		};
+		// Nothing provided, should show steps.
+		assert!(UpCommand { ..Default::default() }.should_show_deployment_steps(&deployment));
```
chungquantin, line 943:
```suggestion
		assert!(UpCommand::default().should_show_deployment_steps(&deployment));
```

crates/pop-cli/src/commands/up/rollup.rs
```diff
@@ -209,16 +232,27 @@ impl UpCommand {
...
 			deployment_config,
 			id,
 			self.path.as_deref(),
+			self.profile.clone(),
 			cli,
 		)
 		.await
 	}
+
+	fn should_show_deployment_steps(&self, deployment: &Deployment) -> bool {
```
chungquantin, line 250: Combining with the request above about the `step`, a trait called `MultistepCommand` that is implemented for any command whenever they want to implement a multi step flow with indices will keep the code cleaner and scalable. All the logic like printing steps, `step_shown` are trait methods.

crates/pop-cli/src/deployment_api.rs
```diff
@@ -285,7 +285,7 @@ mod tests {
 			json!({
 				"name": "Development",
 				"properties": {
-					"basedOn": "standard",
+					"basedOn": "r0gue-io/standard",
```
chungquantin, line 288:
```suggestion
					"basedOn": "r0gue-io/base-parachain",
```
As I remember from the feedback, we also want that user can easily copy this string and paste on google to find the repo: 

Screenshot 2025-03-25 at 13 33 49

chungquantin (changes requested):

crates/pop-cli/src/cli.rs
AlexD10S: The idea behind using the same number for all warnings was to maintain consistency in timing. I can reduce it to 1 if the issue is that 2 is too long.

crates/pop-cli/src/cli.rs
AlexD10S: It seems the GIF doesn't clearly show the color differences, but the NOTE messages already include your suggestion.
Captura de pantalla 2025-03-25 a las 8 44 32

crates/pop-cli/src/commands/up/rollup.rs
AlexD10S: We only show the deployment steps if the user hasn't skipped them via the command line and isn't using the 'register only' option, see `should_show_deployment_steps`. The idea behind showing the steps is to help guide newcomers.

AlexD10S: The `print_with_steps` is a good idea to keep consistency across all the steps we integrate: https://github.com/r0gue-io/pop-cli/pull/483/commits/cc3036d3a76b78300f303f745f5e01a4cc8f2bf5

crates/pop-cli/src/commands/up/rollup.rs
AlexD10S: https://github.com/r0gue-io/pop-cli/pull/483/commits/cc3036d3a76b78300f303f745f5e01a4cc8f2bf5

crates/pop-cli/src/commands/up/rollup.rs
AlexD10S: https://github.com/r0gue-io/pop-cli/pull/483/commits/cc3036d3a76b78300f303f745f5e01a4cc8f2bf5

crates/pop-cli/src/commands/up/rollup.rs
AlexD10S: https://github.com/r0gue-io/pop-cli/pull/483/commits/eca9285464cf2e0ec0ee8e4c3d4effe1c59e9788

crates/pop-cli/src/commands/up/rollup.rs
AlexD10S: https://github.com/r0gue-io/pop-cli/pull/483/commits/eca9285464cf2e0ec0ee8e4c3d4effe1c59e9788

crates/pop-cli/src/commands/up/rollup.rs
AlexD10S: https://github.com/r0gue-io/pop-cli/pull/483/commits/eca9285464cf2e0ec0ee8e4c3d4effe1c59e9788

crates/pop-cli/src/commands/up/rollup.rs
AlexD10S: https://github.com/r0gue-io/pop-cli/pull/483/commits/eca9285464cf2e0ec0ee8e4c3d4effe1c59e9788

crates/pop-cli/src/commands/up/rollup.rs
AlexD10S: `should_show_deployment_steps` is very specific method of this logic. Other features implementing the steps will have its own logic to decide if show steps or not. The `bool` has to be passed to the `print_with_step` you suggested https://github.com/r0gue-io/pop-cli/pull/483/commits/cc3036d3a76b78300f303f745f5e01a4cc8f2bf5

crates/pop-cli/src/commands/up/mod.rs
AlexD10S: Added in style. https://github.com/r0gue-io/pop-cli/pull/483/commits/eca9285464cf2e0ec0ee8e4c3d4effe1c59e9788

crates/pop-cli/src/cli.rs
AlexD10S: After some testing, 1 second stop seems enough: https://github.com/r0gue-io/pop-cli/pull/483/commits/f13958dc8615cc93d0e0aa245fc896d0bd9b5602

crates/pop-cli/src/deployment_api.rs
AlexD10S: You are right: https://github.com/r0gue-io/pop-cli/pull/483/commits/f2b0338f5ef9ac54176b195c137486d2d843a764

crates/pop-cli/src/commands/up/rollup.rs
chungquantin: > We only show the deployment steps if the user hasn't skipped them via the command line and isn't using the 'register only' option, see `should_show_deployment_steps`. The idea behind showing the steps is to help guide newcomers.

Hmm personally I think the step is a great add for the whole flow not just beneficial to newcomers.

crates/pop-cli/src/cli.rs
chungquantin: I see thank you!

chungquantin (approved):
PR looks clean! Happy to approve. I left one comment about display steps the incremental steps design but up to you if you think it makes sense.

crates/pop-cli/src/commands/up/rollup.rs
AlexD10S: Thanks! I'll leave it as-is for now, might revisit it later, especially if more steps get added.

PR #484, opened 2025-03-24
chore: update prompts, refactor profile variants, dim weight path description

chungquantin:
## Description
- [x] Update prompt of the “save updated parameter”
- [x] Update prompt of the “parameter menu”
- [x] Fix the weight output path when running pop bench pallet to print the actual path instead of the temp path.


AlexD10S (approved):
Nice improvements!!

PR #489, opened 2025-03-26
fix: sort releases by published_at

chungquantin:
Odds that latest release is not always at the top of the release page, this issue happens with https://github.com/r0gue-io/polkadot/releases where the order is messed up for all the releases. This PR sorts the fetched release by the `published_at` field. See https://api.github.com/repos/r0gue-io/polkadot/releases


AlexD10S (approved):
LGTM! Thanks for making this change

PR #491, opened 2025-03-26
refactor: support wasm and polkavm contracts

evilrobot-01:
Enables support for either wasm (default) or polkavm contracts. Tested for successful compilation only, not that tests pass. 

The `pop-contracts` crate is split via `v5` (default) and `v6` features. The `pop-cli` crate is split into `wasm-contracts` (default) and `polkavm-contracts` features. I found being explicit much easier.  The existing `contract` feature is retained as an alias to `wasm-contracts`. A new `contracts` feature alias wraps `polkavm-contracts` and would be the preferred option once wasm contracts are removed.

Default install (wasm contracts): 
```shell
cargo install --path ./crates/pop-cli --locked
```

PolkaVM contracts only: 
```shell
cargo install --path ./crates/pop-cli --locked --no-default-features -F polkavm-contracts
```

PolkaVM and parachain support: 
```shell
cargo install --path ./crates/pop-cli --locked --no-default-features -F polkavm-contracts,parachain
```


AlexD10S:
The content of this PR was merged into https://github.com/r0gue-io/pop-cli/pull/500

PR #492, opened 2025-03-26
feat: new pop add-to command + pop new pallet add pallet to runtime

tsenovilla:
Hey there!  

This PR adds a new command to the POP CLI: `pop add-to`, which enables modifying existing Polkadot-SDK workspaces by adding predefined structures to their code.  

## Overview  
The command follows a structured format:  
- **`pop add-to pallet config-type`** → Adds a new config type to a pallet.  
- **`pop add-to runtime pallet`** → Adds a common pallet to a runtime.  

Each command ensures proper integration and includes a rollback mechanism to prevent partial modifications if an error occurs.  

### `pop add-to pallet config-type`  
Used to modify a **pallet** by adding a new config type. The changes are applied to:  
✅ The pallet itself.  
✅ The pallet’s mock runtime.  
✅ The pallet’s config preludes (if they exist).  
✅ The runtime’s implementation block (if needed).  

#### 📌 Usage:  
- Can be run inside a pallet directory.  
- Can be run from another location by specifying the pallet’s directory path.  

#### 📌 Supported Config Types:  
- `RuntimeEvent`  
- `RuntimeOrigin`  
- `RuntimeHoldReason`  
- `RuntimeFreezeReason`  
- `Fungibles`  

New types can be added easily by modifying:  
`crates/pop-cli/src/commands/add/pallet_config_type/common_types.rs`  

### `pop add-to runtime pallet`  
Used to add a **pallet** to an existing runtime, supporting both `construct_runtime` and `#[runtime]` macros. This command:  
✅ Adds the pallet to the runtime manifest.  
✅ Creates a new file `configs/{pallet_name}.rs` to manage its configuration.  
✅ Allows specifying an alternative path for the implementation block if needed.  

#### 📌 Usage:  
- Can be run inside a runtime directory.  
- Can be run from another location by specifying the runtime’s directory path.  

#### 📌 Supported Pallets:  
- `Balances`  
- `Contracts`  

New pallets can be added easily by modifying:  
`crates/pop-cli/src/commands/runtime_pallet/common_pallets.rs`  

-----

Note that `pop add-to runtime` and `pop add-to pallet` are both valid commands, and `pop add-to runtime pallet` / `pop add-to pallet config-type` are just subcommands. Maybe other subcommands can be added in the future or maybe not, but anyway I think it's worth it to use this format as it clearly specifies what is being added (config-type/pallet) and where (pallet/runtime).

Additionally, the PR modifies the pop new pallet command, so now if it's run inside a workspace containing a runtime, the new pallet will be automatically added to that runtime.

I hope you may find this contribution useful, and happy to discuss any improvement in the UX that you can think of :)


crates/pop-cli/src/commands/mod.rs
```diff
@@ -45,6 +46,9 @@ pub(crate) enum Command {
 	/// Remove generated/cached artifacts.
 	#[clap(alias = "C")]
 	Clean(clean::CleanArgs),
+	/// Add a new feature to your existing polkadot-sdk project
+	#[clap(name = "add-to", alias = "a")]
```
chungquantin, line 50: I feel like `pop add-to` does not look very good as a CLI subcommand so proposes a few suggestions below

```suggestion
	#[clap(name = "insert", alias = "a")]
```

or 

```suggestion
	#[clap(name = "expand", alias = "a")]
```

or maybe just 

```suggestion
	#[clap(name = "add", alias = "a")]
```

crates/pop-cli/src/commands/mod.rs
```diff
@@ -45,6 +46,9 @@ pub(crate) enum Command {
 	/// Remove generated/cached artifacts.
 	#[clap(alias = "C")]
 	Clean(clean::CleanArgs),
+	/// Add a new feature to your existing polkadot-sdk project
```
chungquantin, line 49:
```suggestion
	/// Add a new feature to your existing parachain project
```

We call it `parachain project` in other places if my memory is correct.

crates/pop-cli/src/commands/new/contract.rs
```diff
@@ -75,7 +75,7 @@ impl NewContractCommand {
 		generate_contract_from_template(name, path, &template)?;
 
 		// If the contract is part of a workspace, add it to that workspace
-		if let Some(workspace_toml) = find_workspace_toml(path) {
+		if let Some(workspace_toml) = rustilities::manifest::find_workspace_manifest(path) {
```
chungquantin, line 78:
```rs
pub fn find_workspace_manifest<P: AsRef<Path>>(path: P) -> Option<PathBuf> {
	fn do_find_workspace_manifest(path: &Path) -> Option<PathBuf> {
		let mut path = path;
		// If the target itself contains a manifest, return it
		let cargo_toml_path = path.join("Cargo.toml");
		match Manifest::from_path(&cargo_toml_path) {
			Ok(manifest) if manifest.workspace.is_some() => return Some(cargo_toml_path),
			_ => (),
		}

		// Otherwise, search in the parent dirs
		while let Some(parent) = path.parent() {
			let cargo_toml_path = parent.join("Cargo.toml");
			match Manifest::from_path(&cargo_toml_path) {
				Ok(manifest) if manifest.workspace.is_some() => return Some(cargo_toml_path),
				_ => path = parent,
			}
		}
		None
	}
	do_find_workspace_manifest(path.as_ref())
}
```

Can we simply reimplement this logic in pop-cli and add tests for it? That way we can ensure the test coverage

crates/pop-cli/src/commands/new/pallet.rs
```diff
@@ -47,6 +62,11 @@ pub struct NewPalletCommand {
 	pub(crate) authors: Option<String>,
 	#[arg(short, long, help = "Pallet description", default_value = "Frame Pallet")]
 	pub(crate) description: Option<String>,
+	#[arg(
+		long,
+		help = "If your pallet is created in a workspace containing a runtime, Pop-Cli will place the impl blocks for your pallets' Config traits inside a dedicated file under configs directory. Use this argument to point to other path."
```
chungquantin, line 67:
```suggestion
		help = "If your pallet is created within a workspace that includes a runtime, the `impl` blocks will be generated for your pallet's `Config` traits in a dedicated file under the configs directory. Use this argument to specify a different path."
```

crates/pop-cli/src/commands/new/pallet.rs
```diff
@@ -146,10 +166,10 @@ impl NewPalletCommand {
 		// If the user has introduced something like pallets/my_pallet, probably it refers to
 		// ./pallets/my_pallet. We need to transform this path, as otherwise the Cargo.toml won't be
 		// detected and the pallet won't be added to the workspace.
-		let pallet_path = prefix_with_current_dir_if_needed(pallet_path);
+		let pallet_path = rustilities::paths::prefix_with_current_dir(pallet_path);
```
chungquantin, line 169: Can we modify the existing method instead of introducing the crate method?

crates/pop-cli/src/commands/new/pallet.rs
```diff
@@ -176,24 +196,232 @@ impl NewPalletCommand {
...
+				),
+				Span::call_site(),
+			);
+			let pallet_type = parse_str::<Type>(&pallet_crate_name.replace("-", "_"))?;
+
+			let pallet_name =
+				pallet_crate_name.splitn(2, '-').nth(1).unwrap_or("pallet").to_string();
+
+			let (runtime_lib_path, configs_rs_path, configs_folder_path, configs_mod_path) =
+				rust_writer_helpers::compute_pallet_related_paths(&runtime_path);
```
chungquantin, line 237: I think this logic need to be in pop-cli as well.

chungquantin (commented):
Thanks @tsenovilla for your great work. I just gave a high level review, my suggestion is we should reduce the number of external crates added to `pop-cli` because it increase the cost of maintaining, especially crates are not battle-tested. 

On the other hand, could you add to the README the overall flow of the functionalities? It would be very helpful for reviewers to understand how the code logic works exactly 💪

crates/pop-cli/src/commands/new/pallet.rs
```diff
@@ -7,13 +7,28 @@ use crate::{
...
+use proc_macro2::Span;
+use rust_writer::{
+	ast::{
+		finder::{Finder, ToFind},
+		implementors::{ItemToFile, ItemToMod, TokenStreamToMacro},
+		mutator::{Mutator, ToMutate},
+	},
+	preserver::Preserver,
+};
+use rustilities::manifest::{ManifestDependencyConfig, ManifestDependencyOrigin};
```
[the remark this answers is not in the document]
AlexD10S: Good point, Tin! I totally understand your concerns about maintenance on our side. However, I suggested to him that it should be in a separate crate because it was introducing a lot of code into `pop-cli` that wasn’t directly related to `pallet` development. Tomas did a great job with that library, and I believe it could be useful across different ecosystem projects and other Rust projects as well. So it makes more sense for it to live in its own crate and be made more widely available.

Cargo.toml
```diff
@@ -24,14 +24,26 @@ version = "0.6.0"
 anyhow = "1.0"
 assert_cmd = "2.0.14"
 cargo_toml = "0.20.3"
+# deranged 0.4.1 implements PartialOrd for some types leading to conflicts with core primitive types.
+# Some related discussion: 'https://github.com/jhpratt/deranged/issues/18'
+# This creates conflicts on the dependency tree of Pop-cli as deranged is somewhere there and some downstream crates 
+# are affected in the way described in the discussion. Til that's fixed, a pin to the version 0.4.0 is necessary as 
+# a workaround. Pinned in all pop-cli crates for this reason.
+deranged = { version = "=0.4.0", default-features = false }
```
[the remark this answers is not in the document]
AlexD10S: The same fix was applied in `ink!` https://github.com/use-ink/ink/pull/2446
I am not sure which crate uses this dependency either, will probably need to investigate it better

crates/pop-cli/src/commands/new/pallet.rs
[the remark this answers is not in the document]
tsenovilla: Hey @chungquantin! 

Those new crates (`rust_writer`, `fs_rollback` and `rustilities`) were born as a natural consequence of this specific PR. All that functionality (in a primitive way sometimes, as I redefined many parts of it in those crates) were originally part of this PR, being part of the manifest file you mentioned and many others. As @AlexD10S said, we agreed off-chain to move all that functionality out of this repo, cause it's "general Rust functionality", meaning that it can be used here and in other projects as well. I partially agree with you when you said that it might increase maintenance cost, but having all that code inside `pop-cli` is kinda unnecessary, it'd just flood the repo with code that doesn't need to be defined here, so that also increase maintenance cost.

I invite you to check [this commit](https://github.com/r0gue-io/pop-cli/pull/492/commits/fb68d9929f8d9c56e1375aa7358bda6a27cdbc09), which is the last one that contains all that functionality inside `pop cli`, so you'd get a clearer picture of how it'd look like.

Finally, IMHO, it makes more sense to keep the actual approach of having that functionality outside `pop-cli`, but at the end of the day this is your product, so if your team reaches consensus in the other direction we can just revert the commit I mentioned above, just let me know what you think :)

Cargo.toml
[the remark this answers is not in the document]
tsenovilla: > The same fix was applied in `ink!` [use-ink/ink#2446](https://github.com/use-ink/ink/pull/2446) I am not sure which crate uses this dependency either, will probably need to investigate it better

yep, if you check the `deranged` repo you'll find more than one issue  talking about this thing. It seems that it have broken many crates out there. Dunno if don't updating the lock file is a possibility here as the PR needs to add some new crates (either the ones I created or some of their deps). Additionally I believe that @AlexD10S found this issue in a different branch (please correct me here), so maybe your downstream's already affected

Cargo.toml
```diff
@@ -24,14 +24,26 @@ version = "0.6.0"
...
 dirs = "5.0"
 duct = "0.13"
 env_logger = "0.11.7"
 flate2 = "1.0.30"
+fs_rollback = "3.0.0"
 git2 = { version = "0.18", features = ["vendored-openssl"] }
 glob = "0.3.1"
 log = "0.4.20"
 mockito = "1.4.0"
+pathdiff = "0.2.1"
```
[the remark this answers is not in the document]
tsenovilla: Let me know if the team agrees on this and I'd implement the code in your codebase then :)

crates/pop-cli/src/commands/new/pallet.rs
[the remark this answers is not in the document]
chungquantin: Hey @tsenovilla, thanks for your detailed explanation! I discussed with Alex and I agree this approach makes sense to me know. Will review it more to see if there are any parts to optimize.

crates/pop-cli/src/commands/mod.rs
tsenovilla: Hey Tin, the command was originally called `pop add` as you mentioned. As a standalone command, it looks much better than `pop add-to`, the issue here comes when you actually use it. `pop add runtime pallet` may feel a bit ambiguous IMO, what are you adding? a pallet **to** a runtime? a runtime and a pallet to the workspace? 

That's why I came up with the name `pop add-to` as I believe it emphasizes better what it does: `pop add-to runtime pallet` -> Add a pallet to a runtime. 

But maybe I'm overthinking 😅, just let me know what do you think about this, we can easily change this name if you still consider other names fit better

crates/pop-cli/src/commands/mod.rs
tsenovilla: Mmm, the command is useful in any project using the polkadot-sdk, doesn't need to be created with `pop`, that's why I used that word there. You can also add a config type to a standalone pallet, so dunno if the word parachain would be more correct here

crates/pop-cli/src/commands/new/contract.rs
chungquantin: Edit: Understand the motivation now, I believe there are code in the codebase has similar logic to this, I plan to add a similar one as well so worths a refactor.

tsenovilla: I resolve this conversation due to the discussion above.

crates/pop-cli/src/commands/new/pallet.rs
tsenovilla: I resolve this conversation due to the discussion above.

crates/pop-cli/src/commands/mod.rs
chungquantin: Hmm I see. What do you think about `expand`? Expanding a runtime with a pallet, expanding a pallet with a config type. 

```
pop expand runtime pallet
```

```
pop expand pallet config-type
```

crates/pop-cli/src/commands/new/pallet.rs
tsenovilla: I don't fully understand what you mean here, can you be a bit more explicit or suggest the change you're thinking of??

chungquantin (commented):
Another high level review.

crates/pop-cli/tests/pop_add_to_runtime_pallet.rs
```diff
@@ -0,0 +1,228 @@
+// SPDX-License-Identifier: GPL-3.0
+
+use assert_cmd::Command;
+use similar::{ChangeTag, TextDiff};
+
+#[test]
+fn pop_add_to_runtime_pallet_runtime_macro_v2_works() {
```
AlexD10S, line 7: Like the previous case, it correctly shows that the required lines are being injected. However, it doesn't yet confirm that everything works end-to-end, for example, that it compiles. Since the goal of an e2e test is to validate the full flow, including `pop build` would be great.

crates/pop-common/src/manifest.rs
```diff
@@ -29,33 +29,6 @@ pub fn from_path(path: Option<&Path>) -> Result<Manifest, Error> {
 	Ok(Manifest::from_path(path.canonicalize()?)?)
 }
 
-/// This function is used to determine if a Path is contained inside a workspace, and returns a
-/// PathBuf to the workspace Cargo.toml if found.
-///
-/// # Arguments
-/// * `target_dir` - A directory that may be contained inside a workspace
-pub fn find_workspace_toml(target_dir: &Path) -> Option<PathBuf> {
```
AlexD10S, line 37: As above, since we already have the `find_workspace_toml` code, if it needs a tweak, it’s better to update it or leave it as is and create a separate function specifically for the pallet if needed, rather than remove it and rely on an external library.

crates/pop-common/src/rust_writer_helpers/sample_files/basic_pallet.rs
```diff
@@ -0,0 +1,35 @@
+// SPDX-License-Identifier: GPL-3.0
```
AlexD10S, line 1: To keep things simpler and avoid introducing extra new code (even if it's just sample files for testing), would it be possible to just instantiate a new pallet in a temporary file instead?

For reference, we follow a similar approach for parachain testing here: https://github.com/r0gue-io/pop-cli/blob/main/crates/pop-parachains/src/new_parachain.rs#L102

crates/pop-common/src/rust_writer_helpers/sample_files/runtime_using_construct_runtime_macro.rs
```diff
@@ -0,0 +1,258 @@
+// SPDX-License-Identifier: GPL-3.0
```
AlexD10S, line 1: Same as above, to avoid introducing extra new code (even if it's just sample files for testing), you can use something like this: https://github.com/r0gue-io/pop-cli/blob/main/crates/pop-parachains/src/new_parachain.rs#L102

crates/pop-common/src/rust_writer_helpers/tests.rs
```diff
@@ -0,0 +1,118 @@
+// SPDX-License-Identifier: GPL-3.0
+
+use super::*;
```
AlexD10S, line 3: For consistency with the rest of the codebase, we keep unit tests in the same file as the implementation. You can move this code to `rust_writer_helpers`.

.github/workflows/ci.yml
```diff
@@ -199,3 +199,43 @@ jobs:
 
       - name: Run integration tests
         run: cargo test --no-default-features --features parachain --test parachain
+
+  file-writer-integration-tests:
```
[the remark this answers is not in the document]
AlexD10S: It's currently not covered in `parachain-integration-tests` because we have `cargo test --no-default-features --features parachain --test parachain`. But definitely easier to just add it  there (https://github.com/r0gue-io/pop-cli/blob/main/.github/workflows/ci.yml#L201) than having a whole new job:
``` 
run: |
    cargo test --no-default-features --features parachain --test parachain
    cargo test --no-default-features --features parachain --test new_pallet_added_to_existing_runtime
    cargo test --no-default-features --features parachain --test pop_add_to_pallet_config_type
    cargo test --no-default-features --features parachain --test pop_add_to_runtime_pallet
```

crates/pop-cli/src/commands/add/runtime_pallet/common_pallets.rs
```diff
@@ -0,0 +1,96 @@
...
+			  ///TEMP_DOC
+			  #[derive_impl(pallet_balances::config_preludes::TestDefaultConfig)]
+			  impl pallet_balances::Config for Runtime{
+				type AccountStore = System;
+			  }
+			},
+			CommonPallets::Contracts => parse_quote! {
+			  ///TEMP_DOC
+			  #[derive_impl(pallet_contracts::config_preludes::TestDefaultConfig)]
+			  impl pallet_contracts::Config for Runtime{
```
AlexD10S, line 88: Is this not wrong config? e.g our template: https://github.com/r0gue-io/contracts-parachain/blob/main/runtime/src/configs/contracts.rs#L77

AlexD10S (changes requested):
Nice job! There's a lot of great work here—¡, really cool stuff to handle such a complex feature.
I’ve left a few comments in the code, but we’ll definitely need a more thorough review before merging to main. My main concern right now is around polishing the core functionality and improving the DevEx.

As mentioned in one comment, I’d recommend simplifying the CLI commands to something like:
- `pop add pallet`
- `pop add config` (although, personally, I don’t find this second one super useful)

A few functionality issues I ran into:
- Running `pop add-to runtime pallet` in a freshly generated template fails if I select `Balances` (because it already exists), and selecting `Contracts` also fails, I believe due to version conflicts.
Can we detect maybe if Balances already exist and not display it? or too complex?

- Running `pop new pallet` followed by `pop add-to pallet config-type` (selecting `Fungibles`) throws this error:
```
Error: Cannot mutate using Mutator: ItemToImpl { trait_name: Some("Config"), implementor_name: "Test", impl_item: ImplItem::Type { attrs: [], vis: Visibility::Inherited, defaultness: None, type_token: Type, ident: Ident(Fungibles), generics: Generics { lt_token: None, params: [], gt_token: None, where_clause: None }, eq_token: Eq, ty: Type::Path { qself: None, path: Path { leading_colon: None, segments: [PathSegment { ident: Ident(Balances), arguments: PathArguments::None }] } }, semi_token: Semi } }
```

crates/pop-cli/tests/new_pallet_added_to_existing_runtime.rs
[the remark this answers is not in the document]
tsenovilla: This is a nice point, but not straightforward probably. I also thought about that possibility, but there's the painpoint that `assert_cmd` has some limitations out of integration tests: https://docs.rs/assert_cmd/latest/assert_cmd/cargo/index.html 

So that'd need some rework for all the tests, not just the ones introduced by this PR

crates/pop-common/src/rust_writer_helpers.rs
```diff
@@ -0,0 +1,249 @@
...
+
+#[derive(Debug, Clone, PartialEq)]
+pub enum DefaultConfigType {
+	Default { type_default_impl: ImplItem },
+	NoDefault,
+	NoDefaultBounds { type_default_impl: ImplItem },
+}
+
+#[derive(Debug, Clone, PartialEq)]
+pub enum RuntimeUsedMacro {
```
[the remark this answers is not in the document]
tsenovilla: Sure sorry, I though it was clear enough, but this is definitely something to do

crates/pop-cli/src/commands/add/pallet_config_type.rs
```diff
@@ -0,0 +1,322 @@
...
+	pub(crate) types: Vec<common_types::CommonTypes>,
+	#[arg(
+		short,
+		long,
+		help = "Pop-CLI will add the config type to the current directory lib if there's one. Use this argument to point to other path."
+	)]
+	pub(crate) pallet_path: Option<PathBuf>,
+	#[arg(
+		long,
+		help = "If your pallet is part of a workspace containing a runtime, Pop-Cli will look for the impl block in configs/your_pallet_name.rs or in the runtime lib file to add the new type. Use this argument to point to other path."
```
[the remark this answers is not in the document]
tsenovilla: Hmmm no but this is a great idea, currently the command works either but being run in a runtime folder or specifying the runtime path. Maybe an improvement for a future PR?

crates/pop-cli/src/commands/add/pallet_config_type.rs
```diff
@@ -0,0 +1,322 @@
...
+		// This may be Err cause the config preludes file may not exist. And it's perfect
+		let _ = rollback.note_file(&config_preludes_path);
+		// Note pallet_impl_path if needed
+		if let Some(ref pallet_impl_path) = pallet_impl_path {
+			rollback.note_file(&pallet_impl_path)?;
+		}
+
+		let roll_pallet_lib_path = rollback
+			.get_noted_file(&pallet_lib_path)
+			.expect("The file has been noted above; qed;");
```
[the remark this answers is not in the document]
tsenovilla: This isn't an error. `get_noted_file` returns `Option<&Path>` (https://docs.rs/fs_rollback/latest/fs_rollback/struct.Rollback.html#method.get_noted_file). That expect means that at this point of the flow it's always `Some` cause you've used the [note_file](https://docs.rs/fs_rollback/latest/fs_rollback/struct.Rollback.html#method.note_file) method above

crates/pop-cli/src/commands/add/runtime_pallet/common_pallets.rs
```diff
@@ -0,0 +1,96 @@
...
+			  ///TEMP_DOC
+				#[runtime::pallet_index(#highest_index)]
+				pub type Contracts = pallet_contracts;
+			},
+		}
+	}
+
+	pub fn get_version(&self) -> String {
+		match self {
+			CommonPallets::Balances => "39.0.0".to_string(),
```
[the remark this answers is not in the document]
tsenovilla: agree

crates/pop-cli/src/commands/add/runtime_pallet.rs
```diff
@@ -0,0 +1,277 @@
...
+		let pallets = if self.pallets.is_empty() {
+			multiselect_pick!(
+				common_pallets::CommonPallets,
+				"Select the pallets you want to include in your runtime"
+			)
+		} else {
+			self.pallets
+		};
+
+		let mut rollback = Rollback::default();
```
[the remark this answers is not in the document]
tsenovilla: It's needed due to the command modifies several files: at least the runtime lib and the runtime manifest, but also the file where the new pallet is implemented if it's not at runtime lib. So this rollback takes care of everything and doesn't modify any of the files if it cannot modify all of them. IMO it's a really nice to have, so the command modifies the file system atomically

tsenovilla:
As discussed offline, we'll split this PR in smaller ones and share the remaining work (mostly refactors). I'll think of the best way to do it and create a dedicated issue asap.

We should keep this branch as reference for the moment imo

tsenovilla:
As discussed offline, this PR should be splitted in smaller ones. 

I've just created the first one:

- #524 

So I'm closing this as is not longer needed. I keep the branch for the moment as we might be interested in bringing the rest of the functionality in further PRs

PR #493, opened 2025-03-26
feat: revive compatibility

AlexD10S:


AlexD10S:
The logic of this PR has been replaced with https://github.com/r0gue-io/pop-cli/pull/500

PR #494, opened 2025-03-27
fix: argument exists in bench commands & skip parameters flag

chungquantin:
- Add patch fix the `display()` of `storage`, `machine` and `block`. 
- Rename `skip_configuration` to `skip_parameters`


AlexD10S (approved):
LGTM!

PR #496, opened 2025-03-28
feat: try-runtime feature

chungquantin:
poptest

# test

The Pop CLI feature is built on top of the [`try-runtime-cli`](https://github.com/paritytech/try-runtime-cli?tab=readme-ov-file), a programmatic testing framework designed for Substrate. It allows you to test runtime upgrades and state transitions.

- `on-runtime-upgrade`: Tests migrations.
- `execute-block`: Executes a specified block against some state.
- `fast-forward`: Executes a runtime upgrade (optional), then mines a number of blocks while performing try-state checks.
- `create-snapshot`: Creates a chain state snapshot.

## test on-runtime-upgrade

To test [runtime upgrades](https://docs.polkadot.com/develop/parachains/maintenance/runtime-upgrades/), you can use the `on-runtime-upgrade` subcommand. This command allows you to test migrations by executing the [`OnRuntimeUpgrade`](https://paritytech.github.io/polkadot-sdk/master/frame_support/traits/trait.OnRuntimeUpgrade.html) hooks of a runtime against the state of a live blockchain or a snapshot.

```bash
pop test on-runtime-upgrade live
```

**Note:**

If you choose to run migrations on a local runtime binary, you will be prompted to provide a valid runtime binary path. [Ensure the binary is built with the `try-runtime` feature](./build.md).

> Pop CLI will automatically locate the runtime binary based on the provided `--profile`. If the binary is not found, Pop CLI will build it automatically.

If a runtime is not specified, the migration will be executed on the code currently running on the remote node of a live blockchain or the one stored in the provided snapshot file.

### test on-runtime-upgrade live

To run migrations against a live blockchain, use the `on-runtime-upgrade live` subcommand:

```bash
pop test on-runtime-upgrade live
```

You'll be prompted to provide an RPC URI of a live blockchain. A valid RPC URI must start with `ws://` or `wss://`. You can also specify an optional block hash to define the state execution point.

To manually provide the URI and block hash, use:

```bash
pop test on-runtime-upgrade \
    --runtime=<path_to_runtime_binary> \
    live \
    --uri=wss://rpc1.paseo.popnetwork.xyz \
    --at=0x1234567890abcdef1234567890abcdef1234567890abcdef1234567890abcdef
```

### test on-runtime-upgrade snap

To run migrations against a snapshot, use the `on-runtime-upgrade snap` subcommand:

```bash
pop test on-runtime-upgrade \
    --runtime=<path_to_runtime_binary> \
    snap \
    --path=./example.snap
```

[...]

PR #497, opened 2025-03-28
feat: canonical command names

Daanvdplas:
Refactors the implementations for collecting events and data from executing commands.

Essentially we convert the command before it is executed in the primary event. After execution the cli returns additional data. When an error occurs the error is added as data.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -41,7 +41,7 @@ tracing-subscriber = "0.3.19"
 
 # networking
 reqwest = { version = "0.12", features = ["json", "multipart", "stream"] }
-tokio = { version = "1.0", features = ["macros", "rt-multi-thread"] }
+tokio = { version = "1.0", features = ["macros", "rt-multi-thread", "process"] }
 url = "2.5.4"
 
 # contracts
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -41,7 +41,7 @@ tracing-subscriber = "0.3.19"
 
 # networking
 reqwest = { version = "0.12", features = ["json", "multipart", "stream"] }
-tokio = { version = "1.0", features = ["macros", "rt-multi-thread", "process"] }
+tokio = { version = "1.0", features = ["macros", "rt-multi-thread"] }
 url = "2.5.4"
 
 # contracts
--- a/crates/pop-cli/src/common/mod.rs
+++ b/crates/pop-cli/src/common/mod.rs
@@ -23,69 +23,94 @@ pub mod wallet;
 use std::fmt::{Display, Formatter, Result};
 use strum::VariantArray;
 
+/// Data after command execution.
 #[derive(Debug, PartialEq)]
-pub enum Telemetry {
-	Null,
+pub enum Data {
+	/// Project that was built.
 	Build(Project),
-	Test { project: Project, feature: Feature },
-	Install(Os),
+	/// Project and feature test details.
+	Test {
+		/// Project tested.
+		project: Project,
+		/// Test feature.
+		feature: TestFeature,
+	},
+	/// Project that was started.
 	Up(Project),
+	/// OS where installation occurred.
+	Install(Os),
+	/// Template that was created.
 	New(Template),
+	/// No additional telemetry data.
+	Null,
 }
 
-/// Represents the type of project being operated on.
+/// Project type.
 #[derive(Debug, PartialEq, Clone, VariantArray)]
 pub enum Project {
-	Network,
-	/// A blockchain project (parachain, etc.).
-	Chain,
-	/// A smart contract project.
+	/// Smart contract.
 	Contract,
-	/// Unknown project type.
+	/// Chain.
+	Chain,
+	/// Network.
+	Network,
+	/// Unidentified project.
 	Unknown,
 }
 
-#[derive(Debug, PartialEq)]
-pub enum Template {
-	Chain(pop_parachains::Parachain),
-	Contract(pop_contracts::Contract),
-	Pallet,
-}
-
+/// Test feature.
 #[derive(Debug, PartialEq, Clone, VariantArray)]
-pub enum Feature {
+pub enum TestFeature {
+	/// Unit tests.
 	Unit,
+	/// End-to-end tests.
 	E2e,
 }
 
+/// Project templates.
+#[derive(Debug, PartialEq, Clone)]
+pub enum Template {
+	/// Smart contract template.
+	#[cfg(feature = "contract")]
+	Contract(pop_contracts::Contract),
+	/// Chain template.
+	#[cfg(feature = "parachain")]
+	Chain(pop_parachains::Parachain),
+	/// Pallet template.
+	Pallet,
+}
+
+/// Supported operating systems.
 #[derive(Debug, PartialEq, Clone, VariantArray)]
 pub enum Os {
-	Mac,
+	/// Linux.
 	Linux,
+	/// macOS.
+	Mac,
+	/// Unsupported.
 	Unsupported,
 }
 
 // Display the telemetry in a human-readable format while excluding the command name to prevent
 // double display.
-impl Display for Telemetry {
+impl Display for Data {
 	fn fmt(&self, f: &mut Formatter<'_>) -> Result {
 		use strum::EnumMessage;
-		use Telemetry::*;
+		use Data::*;
 		use Template::*;
 
 		match self {
-			Null => write!(f, "null"),
+			Null => write!(f, ""),
 			Build(project) => write!(f, "{}", project),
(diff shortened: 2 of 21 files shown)


crates/pop-cli/src/common/mod.rs
```diff
@@ -16,3 +16,219 @@ pub mod prompt;
 /// Contains runtime utilities.
 pub mod runtime;
 pub mod wallet;
+
+use std::fmt::{Display, Formatter, Result};
+use strum::VariantArray;
+
+#[derive(Debug, PartialEq)]
+pub enum Telemetry {
```
peterwht, line 24: any reason not to put the contents of this file in the `pop-telemetry` crate?

crates/pop-cli/src/common/mod.rs
```diff
@@ -16,3 +16,219 @@ pub mod prompt;
...
+	Test { project: Project, feature: Feature },
+	Install(Os),
+	Up(Project),
+	New(Template),
+}
+
+/// Represents the type of project being operated on.
+#[derive(Debug, PartialEq, Clone, VariantArray)]
+pub enum Project {
+	Network,
```
peterwht, line 36: nitpick: missing doc comment. Will just leave this one comment as there are other pubs missing doc comments and this is a high-level review (as requested)

crates/pop-cli/src/common/mod.rs
```diff
@@ -16,3 +16,219 @@ pub mod prompt;
...
+
+#[derive(Debug, PartialEq)]
+pub enum Template {
+	Chain(pop_parachains::Parachain),
+	Contract(pop_contracts::Contract),
+	Pallet,
+}
+
+#[derive(Debug, PartialEq, Clone, VariantArray)]
+pub enum Feature {
```
peterwht, line 53: maybe a different name for this -- variants are all around testing? Something like: "TestFeatures", or something like that.

crates/pop-cli/src/common/mod.rs
```diff
@@ -16,3 +16,219 @@ pub mod prompt;
...
+
+		match self {
+			Null => write!(f, "null"),
+			Build(project) => write!(f, "{}", project),
+			Test { project, feature } => write!(f, "{} {}", project, feature),
+			Install(os) => write!(f, "{}", os),
+			Up(project) => write!(f, "{}", project),
+			New(template) => {
+				match template {
+					// Chain(chain) => write!(f, "{}", chain.get_str("Message").unwrap_or("")),
```
peterwht, line 81: nit: leftover comment.

crates/pop-cli/src/main.rs
```diff
@@ -31,29 +31,19 @@ async fn main() -> Result<()> {
 	let maybe_tel = init().unwrap_or(None);
 
 	let cli = Cli::parse();
-	let res = cli.command.execute().await;
+	#[cfg(feature = "telemetry")]
+	let command = cli.command.to_string();
```
peterwht, line 35: hmm, the telemetry feature gating here seems a little off. Shouldn't the `command` variable exist regardless of telemetry or not?

crates/pop-cli/src/main.rs
```diff
@@ -31,29 +31,19 @@ async fn main() -> Result<()> {
...
-				tel.clone(),
-				&command,
-				json!({&subcommand: sub_data.to_string()}),
-			)
-			.await;
-		} else {
-			let _ = record_cli_command(tel, "error", json!({&command: &subcommand})).await;
-		}
+	if let Some(tel) = maybe_tel {
+		let _ = record_cli_command(tel.clone(), &command, &data).await;
```
peterwht, line 44: make sure to keep this comment: "// Best effort to send on first try, no action if failure."

crates/pop-cli/src/main.rs
```diff
@@ -134,19 +120,210 @@ mod tests {
...
+				"bench pallet"
+			);
+			// Bench error.
+			let error = "bench failed: runtime error";
+			let (command, data) = simulate_command_flow(
+				Command::Bench(BenchmarkArgs { command: Pallet(Default::default()) }),
+				Err(anyhow!(error)) as Result<common::Telemetry>,
+			);
+			assert_eq!(command, "bench pallet");
+			assert_eq!(data, error);
```
peterwht, line 326: These tests are missing a few commands, such as `call` and `clean`.

peterwht (changes requested):
This is looking good! 

One thing I noticed outside the changes in this PR (but are relevant), is that there are several commands that have the `Null` variant. This would probably be a good PR to fix those. These include commands such as `call` and `clean`.

It would also be great to show what the telemetry output is in Umami as well so you can see what's expected. Especially with a more complex command with a few arguments / features, etc.

crates/pop-cli/src/main.rs
Daanvdplas: Thank you

crates/pop-cli/src/main.rs
Daanvdplas: Thank you

crates/pop-cli/src/common/mod.rs
Daanvdplas: Because it is under the telemetry feature. I renamed it to `Data` and remains in common as the cli crates modules do not need to know that this data is going to telemetry. It can simply be seen as data that is returned in main, and opt in send to telemetry

crates/pop-cli/src/main.rs
Daanvdplas: Renamed it to event.

crates/pop-cli/src/common/mod.rs
```diff
@@ -16,3 +16,248 @@ pub mod prompt;
...
+			Build(project) => write!(f, "{}", project),
+			Test { project, feature } => write!(f, "{} {}", project, feature),
+			Install(os) => write!(f, "{}", os),
+			Up(project) => write!(f, "{}", project),
+			New(template) => match template {
+				#[cfg(feature = "parachain")]
+				Chain(chain) => write!(f, "{}", chain.get_message().unwrap_or("")),
+				#[cfg(feature = "contract")]
+				Contract(contract) => write!(f, "{}", contract.get_message().unwrap_or("")),
+				Pallet => write!(f, "pallet"),
```
peterwht, line 110: does this need to be feature gated behind `parachain`?

crates/pop-cli/src/common/mod.rs
```diff
@@ -16,3 +16,248 @@ pub mod prompt;
...
+
+impl Display for Template {
+	fn fmt(&self, f: &mut Formatter<'_>) -> Result {
+		use Template::*;
+		match self {
+			#[cfg(feature = "parachain")]
+			Chain(chain) => write!(f, "{}", chain),
+			#[cfg(feature = "contract")]
+			Contract(contract) => write!(f, "{}", contract),
+			Pallet => write!(f, "pallet"),
```
peterwht, line 136: similarly (does it need to be feature gated)

peterwht (commented):
awesome PR! Great refactor. 

I just left two small comments, and then after your answer I am ready to approve.

crates/pop-cli/src/common/mod.rs
Daanvdplas: Yes and yes below, the nested types are feature gated.

Cargo.toml
```diff
@@ -41,7 +41,7 @@ tracing-subscriber = "0.3.19"
 
 # networking
 reqwest = { version = "0.12", features = ["json", "multipart", "stream"] }
-tokio = { version = "1.0", features = ["macros", "rt-multi-thread"] }
+tokio = { version = "1.0", features = ["macros", "rt-multi-thread", "process"] }
```
AlexD10S, line 44: What is the reason of this?

AlexD10S (commented):
Great! Thanks for tackling this.
Just need to fix a unit test and a clippy warning, and we’re good to go.

Also tagging @chungquantin — if this gets merged before https://github.com/r0gue-io/pop-cli/pull/496, we have to be sure it follows the same format in Try Runtime.

Daanvdplas:
> Also tagging @chungquantin — if this gets merged before https://github.com/r0gue-io/pop-cli/pull/496, we have to be sure it follows the same format in Try Runtime.

I suggest I handle that after that is merged

Cargo.toml
Daanvdplas: There was an error of something not found due to being under the `process` feature. I have removed it, lets see what the CI says.

AlexD10S (approved):
LGTM! Thanks for the quick fixes

peterwht (approved):
LGTM!

AlexD10S (approved):
LGTM!

PR #498, opened 2025-03-28
feat: `pop test create-snapshot` logic implementation

chungquantin:
Screenshot 2025-03-30 at 16 07 26

## Description

PR implements the feature `pop test create-snapshot` that allows user to generate a snapshot file from chain state. 

```
➜ pop test create-snapshot --help
Create a chain state snapshot

Usage: pop test create-snapshot [OPTIONS] [SNAPSHOT_PATH]

Arguments:
  [SNAPSHOT_PATH]
          The snapshot path to write to.

          If not provided `<spec-name>-<spec-version>@<block-hash>.snap` will be used.

Options:
  -u, --uri <URI>
          The url to connect to

  -a, --at <AT>
          The block hash at which to fetch the state.

          If not provided the latest finalised head is used.

  -p, --pallet <PALLET>...
          A pallet to scrape. Can be provided multiple times. If empty, entire chain state will be scraped.

          This is equivalent to passing `xx_hash_64(pallet)` to `--hashed_prefixes`.

      --prefix <HASHED_PREFIXES>...
          Storage entry key prefixes to scrape and inject into the test externalities. Pass as 0x prefixed hex strings. By default, all keys are scraped and
          included

      --child-tree
          Fetch the child-keys.

          Default is `false`, if `--pallets` are specified, `true` otherwise. In other words, if you scrape the whole state the child tree data is included
          out of the box. Otherwise, it must be enabled explicitly using this flag.

  -y, --skip-confirm
          Automatically source the needed binary required without prompting for confirmation

  -h, --help
          Print help (see a summary with '-h')
```

## How to test the PR? 

Run the below command: 

```
pop test create-snapshot --uri=ws://127.0.0.1:9944 example.snap
```
You can also test with a live URL

```
pop test create-snapshot --uri=wss://rpc.ibp.network/paseo example.snap
```

## Error handling

Screenshot 2025-03-30 at 16 11 52


crates/pop-cli/src/commands/test/create_snapshot.rs
```diff
@@ -0,0 +1,271 @@
+use crate::{
```
AlexD10S, line 1: Just small nitpick: 
```suggestion
// SPDX-License-Identifier: GPL-3.0

use crate::{
```

crates/pop-cli/src/commands/test/create_snapshot.rs
```diff
@@ -0,0 +1,271 @@
...
+		cli.warning(
+			"NOTE: `create-snapshot` only works with the remote node. No runtime required.",
+		)?;
+
+		if self.from.uri.is_none() {
+			self.from.uri = Some(
+				cli.input(format!(
+    				"Enter the URI of the remote node:\n{}",
+    				style(
+    					"Ensures your remote node is built with the `try-runtime` feature enabled. \
```
AlexD10S, line 47: Does this message make sense if you're creating a snapshot of a live network?

crates/pop-cli/src/commands/test/create_snapshot.rs
```diff
@@ -0,0 +1,271 @@
...
+		try_runtime::{check_try_runtime_and_prompt, ArgumentConstructor},
+	},
+};
+use clap::Args;
+use cliclack::spinner;
+use console::style;
+use pop_parachains::{run_try_runtime, state::LiveState, TryRuntimeCliCommand};
+
+const CUSTOM_ARGS: [&str; 2] = ["--skip-confirm", "-y"];
+const DEFAULT_REMOTE_NODE_URL: &str = "ws://127.0.0.1:9944";
```
AlexD10S, line 14: Just a question, but totally up to you. Does make sense to add here Pop Network? or a live network as example? not sure if is common to create snapshots of localhost

crates/pop-cli/src/commands/test/create_snapshot.rs
```diff
@@ -0,0 +1,271 @@
...
+			return display_message(&e.to_string(), false, cli);
+		}
+		display_message(
+			&format!(
+				"Snapshot is created successfully!{}",
+				if let Some(p) = self.snapshot_path {
+					style(format!(
+						"\n{} Generated snapshot file: {}",
+						console::Emoji("●", ">"),
+						p.to_string()
```
AlexD10S, line 85: Just clippy warning

AlexD10S (approved):
Nice and very clean PR. I left a few minor comments, but approved.

The error message when the chain isn't running doesn't seem to make sense, but the error is coming from `try-runtime`. I'll open an issue there.
Captura de pantalla 2025-03-30 a las 20 19 42

chungquantin:
> Nice and very clean PR. I left a few minor comments, but approved.
> 
> 
> 
> The error message when the chain isn't running doesn't seem to make sense, but the error is coming from `try-runtime`. I'll open an issue there.
> 
> Captura de pantalla 2025-03-30 a las 20 19 42
> 
> 

I saw your input had an extra space which causes the parser to be failed. I'll trim the input to avoid this issue.

crates/pop-cli/src/commands/test/create_snapshot.rs
chungquantin: Resolved in https://github.com/r0gue-io/pop-cli/pull/498/commits/4a49b67e2f58b55dc9af3a2422d68776849d443c

crates/pop-cli/src/commands/test/create_snapshot.rs
chungquantin: Agree, I think using Pop network rpc url here makes more sense. I resolved it in https://github.com/r0gue-io/pop-cli/pull/498/commits/4a49b67e2f58b55dc9af3a2422d68776849d443c

crates/pop-cli/src/commands/test/create_snapshot.rs
chungquantin: Revisit this and I don't think it is needed, agree with your suggestion. Resolved in https://github.com/r0gue-io/pop-cli/pull/498/commits/4a49b67e2f58b55dc9af3a2422d68776849d443c

crates/pop-cli/src/commands/test/create_snapshot.rs
chungquantin: Resolved inhttps://github.com/r0gue-io/pop-cli/pull/498/commits/4a49b67e2f58b55dc9af3a2422d68776849d443c

AlexD10S:
> I saw your input had an extra space which causes the parser to be failed. I'll trim the input to avoid this issue.
Ah, I see! That makes sense, all good then.

PR #513, opened 2025-04-04
fix: benchmarking logger

chungquantin:
Initializing the logger in a global scope is not a good approach. Instead, I initialize a new logger on each new command.


AlexD10S (approved):
Thanks for the fix! I can now see the correct output when running `pop up network -f ./network.toml`, and the status still shows up properly when generating weights with `pop bench pallet`. Nicely done!

PR #517, opened 2025-04-08
feat: hashing

evilrobot-01:
Based on https://www.shawntabrizi.com/substrate-js-utilities/ and https://github.com/hack-ink/subalfred, adds a simple `hash` command as an additional utility to the CLI. It is feature gated behind `experimental` for now. It supports all the hashers typically used by the SDK (blake2, keccack, sha2, twox), without taking on any new dependencies.

### Installation
```shell
cargo install --path ./crates/pop-cli --locked --features experimental
```

### Usage 
There is a subcommand for each supported hasher, which expects the required length of the output along with input data in the form of a string, hex or path to a file to be hashed. The command structure is intended to be as simple as possible, whilst also allowing easier future extensibility. Whilst the commands could be combined with the output length (e.g. keccak-256), its the same amount of characters to type and providing the length separately makes it more natural in terms of how a hasher works imo. Possible values for length are included in `--help`.

Some of the commands also support an additional `--concat` option which can be used for storage key hashing.

```shell
pop hash blake2 256 ./runtime.wasm
pop hash twox 128 some_text
pop hash keccak 256 0x68656c6c6f
```

TODOs:
- [x] potentially include the length, in bytes, with the output 
- [ ] docs - a separate pr to add to pop-docs once approved
- [x] tests

[sc-3512]

### DIFF
--- a/crates/pop-cli/src/commands/hash.rs
+++ b/crates/pop-cli/src/commands/hash.rs
@@ -1,5 +1,6 @@
 // SPDX-License-Identifier: GPL-3.0
 
+use self::Command::*;
 use super::*;
 use anyhow::{anyhow, Result};
 use clap::{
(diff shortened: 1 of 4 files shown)


evilrobot-01:
It's experimental until some of the other useful commands in the listed resources land and the overall command structure can be stabilised.

crates/pop-cli/src/commands/hash.rs
```diff
@@ -0,0 +1,444 @@
...
+}
+
+impl Command {
+	/// Executes the command.
+	pub(crate) fn execute(&self) -> Result<()> {
+		let hash = self.hash()?;
+		let additional_info = format!("(Source: {}, Output: {} bytes)", self.data(), hash.len());
+		let output =
+			format!("{} {}", to_hex(&self.hash()?, false), console::style(additional_info).dim());
+		println!("{output}");
```
AlexD10S, line 80: For consistency can we use `cli.info` or `cli.success` instead?

AlexD10S (changes requested):
Nice work! The code is clean, well-organized, and very easy to review. The only change I’d propose is to replace the `println!` with our `CLI` module for the output, just to keep things consistent. Other than that, this looks good to approve and merge!

crates/pop-cli/src/commands/hash.rs
evilrobot-01: Pushed in 18b1408d

AlexD10S (approved):

PR #520, opened 2025-04-08
build(deps): eliminate unnecessary dependencies

evilrobot-01:
Eliminates unnecessary dependencies when building without default features. The number of dependencies when building with the `contract` feature only has been halved from ~1k to 500k. A handful were removed more generally by using `default-features = false`.

Also updates CI to ensure that builds with -`-no-default-features` are successful and remain so. Whilst the tool is primarily aimed towards contract and parachain development, allowing successful builds with no default features allows the introduction and testing of additional experimental features which are not specific to these two developer journeys.

[sc-3569]

### DIFF
--- a/crates/pop-parachains/src/bench/mod.rs
+++ b/crates/pop-parachains/src/bench/mod.rs
@@ -4,6 +4,7 @@ use crate::Error;
 use clap::Parser;
 use duct::cmd;
 use frame_benchmarking_cli::PalletCmd;
+pub use frame_benchmarking_cli::{BlockCmd, MachineCmd, OverheadCmd, StorageCmd};
 use sc_chain_spec::GenesisConfigBuilderRuntimeCaller;
 use serde::{Deserialize, Serialize};
 use sp_runtime::traits::BlakeTwo256;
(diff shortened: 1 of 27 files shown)


crates/pop-parachains/src/bench/mod.rs
```diff
@@ -4,6 +4,7 @@ use crate::Error;
 use clap::Parser;
 use duct::cmd;
 use frame_benchmarking_cli::PalletCmd;
+pub use frame_benchmarking_cli::{BlockCmd, MachineCmd, OverheadCmd, StorageCmd};
```
AlexD10S, line 7: Since this is public now consider adding a /// documentation comment

crates/pop-parachains/src/lib.rs
```diff
@@ -1,7 +1,8 @@
 // SPDX-License-Identifier: GPL-3.0
 
 #![doc = include_str!("../README.md")]
-mod bench;
+
+pub mod bench;
```
AlexD10S, line 5: Same as above, comment here and in `pub mod try_runtime;` below

AlexD10S (approved):
Thanks for the cleanup!
Everything looks good to me, I just left a couple of minor comments suggesting the addition of documentation comments for public modules. Good to go after that!

evilrobot-01:
Docs require some love. As this PR is only about resolving the dependencies, my suggestion is to tackle the docs separately and holistically in a single PR.

Issue #364, opened 2024-12-06
docs: add documentation for all errors and public modules


AlexD10S:
Add docs (`///`) to cover all error types (e.g https://github.com/r0gue-io/pop-cli/blob/main/crates/pop-common/src/errors.rs#L9) and public modules (e.g https://github.com/r0gue-io/pop-cli/blob/main/crates/pop-common/src/lib.rs#L1) across the various crates.

PR #521, opened 2025-04-09
docs: improve project documentation

evilrobot-01:
Addresses all issues reported by `cargo doc` and then also adds missing docs for public items.

[sc-3576]


AlexD10S (approved):
Nice, thanks for taking this. Some of these were necessary and had been pending for a while.

Closes https://github.com/r0gue-io/pop-cli/issues/364

PR #522, opened 2025-04-10
test: no default features

evilrobot-01:
A follow up to #520, ensuring that tests can be run successfully with no/default/all features enabled. Also updates CI workflow to check these cases.

Note: ideally implementations should be refactored into separate tests depending on feature where practical, instead of sprinkling `#[cfg(feature = "xx")] ` throughout. Hopefully this will come about organically over time now that the additional test checks are added to the CI workflow.


AlexD10S (commented):
Although this isn't introduced in this PR, I ran into errors when executing:
```
cargo test --lib --bins --no-default-features --features contract
```
or 
```
cargo test --lib --bins --no-default-features --features parachain
```
Looks like there are some errors with the dependencies when running with individual features. Might be worth fixing that up and updating the CI to test individual features instead of:
```
- name: Run unit tests
        run: cargo test --lib --bins
```

AlexD10S (approved):
Amazing, LGTM!

AlexD10S (approved):

evilrobot-01:
428a89a exposes the github token to those specific steps which need it only. 

It is already used in the CI at https://github.com/r0gue-io/pop-cli/blob/23c7b82dc8d4dd2026d6dab6bf1b668104758f3c/.github/workflows/ci.yml#L76

Additional context:

> At the start of each workflow job, GitHub automatically creates a unique GITHUB_TOKEN secret to use in your workflow. You can use the GITHUB_TOKEN to authenticate in the workflow job. Source: https://docs.github.com/en/actions/security-for-github-actions/security-guides/automatic-token-authentication#about-the-github_token-secret 

https://docs.github.com/en/actions/security-for-github-actions/security-guides/automatic-token-authentication#example-2-calling-the-rest-api also shows it being used in the same way as used within the `ApiClient`.

AlexD10S (approved):

PR #523, opened 2025-04-10
feat(up): add support for launching networks without network config files

evilrobot-01:
Improvements to the `pop up` command to enable the launching of networks with supported parachains without the need for network configuration files. A 'supported' parachain is simply one that is added to the `Parachain` enum within the `pop-parachains` crate, simply so that we can know how to source the binaries/chain-spec-generators for the chain to launch it as quickly as possible.

### Goal 
Support as much of the configuration in https://github.com/r0gue-io/pop-node/blob/main/networks/devnet.toml as possible:
- [x] optional para id specification, defaulting to some known static value 
- [x] port specification, randomly assigned if not specified
- [x] args: currently statically defined per parachain type, need to consider whether to support custom args for the first iteration
- [x] genesis state

Example usage:
```
pop up paseo -p asset-hub,pop -r stable2412-4
pop up kusama --port 8833 --parachain asset-hub:9944 -r stable2412-4
pop up polkadot --port 8833 --parachain asset-hub:9977,pop#3395:9944 -r stable2412-4
pop up westend -p asset-hub
```
The last arg explicitly specifies the relay binary version to work around current issues with `stable2503` and the `v1.4.3` release of Paseo. They would ordinarily not be required. Everything 'just works' on Westend.

Notes:
- Now uses positional path argument for specifying the network config file when using `pop up network`, eliminating the need for `--file/-f`.
- Supported chains: all system chains and Pop.
- The number of validators is automatically defined based on the specified number of parachains. This could be configurable but not sure its required?
- The number of collators is not currently configurable but this could be added. I personally dont see the need for more than one on a dev box, at least when building out and testing a solution.

TODOs:
- [x] fix: consistent ordering of parachains in output in cli output
- [x] logging: output versions used when --verbose enabled
- [x] specifying `pop` should auto-add asset-hub as a dependency, fund sovereign account on AH, notify user if AH not specified
- [x] westend
- [x] merging in applicable local binary versions when resolving which binary to use - e.g. stable2503 remains as prerelease, can be used by explicitly specifying but then defaults back to last release when run normally. It could include local binaries in the version list even if they dont appear in the current published release set. A similar poor experience is being prompted to revert back to the fallback version when rate limited on the GH api, despite having a newer version available locally.
- [x] add local ws endpoint to output for better devex
- [x] tests

[sc-3612]

### DIFF
--- a/crates/pop-cli/src/common/try_runtime.rs
+++ b/crates/pop-cli/src/common/try_runtime.rs
@@ -793,6 +793,7 @@ mod tests {
 						"MessageQueue",
 						"Messaging",
 						"Motion",
+						"MultiBlockMigrations",
 						"Multisig",
 						"NftFractionalization",
 						"Nfts",
@@ -810,6 +811,7 @@ mod tests {
 						"TransactionPayment",
 						"Treasury",
 						"Utility",
+						"WeightReclaim",
 						"XcmpQueue",
 					]
 					.iter()


crates/pop-common/src/git.rs
```diff
@@ -200,11 +200,14 @@ impl GitHub {
 	}
 
 	/// Fetch the latest releases of the GitHub repository.
-	pub async fn releases(&self) -> Result<Vec<Release>> {
+	///
+	/// # Arguments
+	/// * `prerelease` - Whether to include prereleases.
+	pub async fn releases(&self, prerelease: bool) -> Result<Vec<Release>> {
```
AlexD10S, line 206: Nice refactor

crates/pop-common/src/sourcing/mod.rs
```diff
@@ -241,6 +271,205 @@ impl GitHub {
...
+						.and_then(|pattern| pattern.version(v))
+						.unwrap_or(v)
+						.to_string()
+				});
+
+				// Extract versions from any cached binaries - e.g., offline or rate-limited.
+				for file in read_dir(cache).into_iter().flatten().filter_map(|f| {
+					f.ok().and_then(|f| f.file_name().into_string().ok()).filter(|n| {
+						n.starts_with(name) &&
+							// TODO: improve hacky workaround for naming clashes
```
AlexD10S, line 347: How are you thinking to refactor this workaround?

crates/pop-parachains/src/registry/system.rs
```diff
@@ -0,0 +1,159 @@
...
+/// See <https://docs.polkadot.com/polkadot-protocol/architecture/system-chains/asset-hub/> for more details.
+#[derive(Clone)]
+pub struct AssetHub(Parachain);
+impl AssetHub {
+	/// A new instance of the Asset Hub.
+	///
+	/// # Arguments
+	/// * `id` - The parachain identifier.
+	/// * `relay` - The relay chain.
+	pub fn new(id: Id, relay: Relay) -> Self {
```
AlexD10S, line 81: Is the id ever going to change for the system chains?

crates/pop-parachains/src/registry/pop.rs
```diff
@@ -0,0 +1,107 @@
...
+	sourcing::{traits::Source as SourceT, GitHub::ReleaseArchive, Source},
+	target,
+};
+use serde_json::{json, Map, Value};
+use sp_core::crypto::Ss58Codec;
+use std::collections::HashMap;
+
+/// Pop Network makes it easy for smart contract developers to use the power of Polkadot.
+#[derive(Clone)]
+pub struct Pop(Parachain);
```
AlexD10S, line 19: Just to clarify, for each new parachain we add support for, will we need to replicate this structure and create a dedicated file/module like this one?

AlexD10S (commented):
Great job,  this is a really useful feature that makes it much easier for developers to spin up complex local environments. It also opens the door to building a more interactive setup guide or UI, which could really help newcomers onboard more smoothly.

I noticed that in the code you’re introducing the term Parachain again for new structs and methods. Do you think this could be refactored or renamed after the inital refactored we started a month ago.

Just a couple of minor nitpicks:
- The description example: `pop up kusama --port 8833 --parachain asset-hub:9977,pop:9944` doesn't seem to work.  Running `pop up kusama --help` shows that `pop` isn’t listed as a supported parachain:
```
  -p, --parachain <PARACHAIN>
          The parachain(s) to be included. An optional parachain identifier and/or port can be affixed
          via #id and :port specifiers (e.g. `asset-hub#1000:9944`) [possible values: asset-hub,
          bridge-hub, coretime, people]
```
- The `#id:port` format (e.g., `asset-hub#1000:9944`) could be a bit confusing for newcomers. I don’t have a better idea to support a list, but longer-term, an interactive guide or wizard could make this much more approachable.

Once this is merged, it would be worth updating the documentation https://learn.onpop.io/appchains.

crates/pop-cli/src/commands/mod.rs
```diff
@@ -175,7 +177,15 @@ impl Command {
 					None => up::Command::execute(args).await.map(Up),
 					Some(cmd) => match cmd {
 						#[cfg(feature = "parachain")]
-						up::Command::Network(cmd) => cmd.execute().await.map(|_| Up(crate::common::Project::Network)),
+						up::Command::Network(cmd) => cmd.execute().await.map(|_| Up(Network)),
+						#[cfg(feature = "parachain")]
```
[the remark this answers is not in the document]
evilrobot-01: I think it was something along those lines - one would most probably use paseo locally since rococo disappeared and it is the public community testnet. Then one would ideally deploy to Kusama and then Polkadot. Westend was added specifically for revive, so anyone could launch Westend+AH instantly to experiment with contract deployment locally without needing to worry about faucets etc.

Happy to reorder based on any order you prefer.

crates/pop-cli/src/commands/up/network.rs
```diff
@@ -1,39 +1,62 @@
...
+pub(crate) struct ConfigFileCommand {
 	/// The Zombienet network configuration file to be used.
-	#[arg(short, long)]
-	file: String,
+	#[arg(value_name = "FILE", conflicts_with = "file")]
+	pub path: Option<PathBuf>,
+	/// [DEPRECATED] The Zombienet network configuration file to be used (will be removed in
+	/// v0.9.0).
+	#[arg(short = 'f', long = "file")]
+	#[deprecated(since = "0.8.0", note = "will be removed in v0.9.0")]
```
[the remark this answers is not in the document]
evilrobot-01: Addressed in 92ab011d

crates/pop-common/src/sourcing/mod.rs
evilrobot-01: May not be worth the effort tbh, will leave open and revisit before final review.

crates/pop-parachains/src/registry/pop.rs
evilrobot-01: The primary reason for adding a parachain manually here is to enable launching a network instantly, same as was defined in the enums. This effectively means a requirement for prebuilt cross-platform binaries and informing the CLI on how to source these for the chain in question, which may have different release versioning schemes. Using traits provides greater flexibility than the enums provide. Put another way, requirements now demand more than strum can provide - e.g. custom functions for ordering versions to determine the latest across local binaries and remote releases. Of course this only ever becomes useful as other chains are added.

The other aspect is defining requirements (such as Pop having AH as a dependency due to its token) so that these can be enacted automatically based on the launched network. A .toml file is always going to provide maximum flexibility, this rework looks to simplify common use cases, especially as it can eliminate a lot of complexity in the docs for launching a local network and the manually onboarding a parachain, thereby improving the experience.

evilrobot-01: PPS - the registry module could be extacted into its own crate in future to make it easier to extend.

crates/pop-cli/src/commands/up/network.rs
```diff
@@ -51,310 +74,547 @@ pub(crate) struct ZombienetCommand {
...
+	/// Automatically source all necessary binaries required without prompting for confirmation.
+	#[clap(short = 'y', long)]
+	skip_confirm: bool,
+}
+
+impl<const FILTER: u8> BuildCommand<FILTER> {
+	/// Executes the command.
+	pub(crate) async fn execute(&mut self, relay: Relay) -> anyhow::Result<()> {
+		clear_screen()?;
+		intro(format!(
```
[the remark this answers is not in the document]
evilrobot-01: Addressed in e377c9b4.

It required the addition of an error function for error logging, which has been dealt with as its own commit so it can easily be cherry-picked out into a separate PR if desired.

evilrobot-01:
>  It also opens the door to building a more interactive setup guide or UI, which could really help newcomers onboard more smoothly.

I'm not sure it warrants a UI tbh, at least not anytime soon. A UI makes sense for some contexts (e.g. call, where there are many possible calls and many parameters), but slapping a UI on something for the sake of it, when it can be improved to be simpler and therefore more powerful has more value. I would personally rather use shell history/auto-completes with simple commands, than be confronted with a complex UI that I need to interact with for a few minutes each time. 

> I noticed that in the code you’re introducing the term Parachain again for new structs and methods. Do you think this could be refactored or renamed after the inital refactored we started a month ago.

As discussed separately, will refactor to use 'rollup'.

> Just a couple of minor nitpicks:
>     * The description example: `pop up kusama --port 8833 --parachain asset-hub:9977,pop:9944` doesn't seem to work.  Running `pop up kusama --help` shows that `pop` isn’t listed as a supported parachain:

This is intentional. There is no deployment on Kusama, there is no reserved para on Kusama, there is no runtime specifically targeting Kusama.

>     * The `#id:port` format (e.g., `asset-hub#1000:9944`) could be a bit confusing for newcomers. 
Perhaps, but anyone that has familiarity with urls should know that ports are specified via [:port](https://developer.mozilla.org/en-US/docs/Web/API/URL/port) and fragment identifiers via [#hash](https://developer.mozilla.org/en-US/docs/Web/API/URL/hash). This general approach is already used in https://github.com/r0gue-io/pop-cli/blob/1c4797debe430de8fcedd9407743e437dbfc2614/crates/pop-cli/src/commands/up/network.rs#L39-L40 and is just an optional shorthand to define properties succinctly. An alternative might be json, but then you lose the simplicity.

crates/pop-cli/src/commands/up/network.rs
```diff
@@ -46,315 +68,553 @@ pub(crate) struct ZombienetCommand {
...
+	/// The version of the runtime to be used for system parachains, as per the release tag (e.g.
+	/// "v1.4.1"). See <https://github.com/polkadot-fellows/runtimes/releases> for more details.
+	#[arg(short = 'S', long)]
+	system_parachain_runtime: Option<String>,
+	/// The parachain(s) to be included. An optional parachain identifier and/or port can be
+	/// affixed via #id and :port specifiers (e.g. `asset-hub#1000:9944`).
+	#[arg(short, long, value_delimiter = ',', value_parser = SupportedParachains::<FILTER>::new())]
+	parachain: Option<Vec<Box<dyn Para>>>,
+	/// The port to be used for the first relay chain validator.
+	#[clap(short = 'P', long)]
```
evilrobot-01, line 144: This is maybe a little clumsy - as the relay is determined via a command (paseo, polkadot), there needs to be a way to specify the port for a validator. `pop up paseo:9944` doesnt seem great.

crates/pop-common/src/git.rs
```diff
@@ -127,7 +127,7 @@ impl Git {
 	/// Fetch the latest release from a repository
 	fn fetch_latest_tag(repo: &GitRepository) -> Option<String> {
 		let tags = repo.tag_names(None).ok()?;
-		parse_latest_tag(tags.iter().flatten().collect::<Vec<_>>())
+		parse_latest_tag(&tags.iter().flatten().collect::<Vec<_>>()).map(|t| t.to_string())
```
evilrobot-01, line 130: `parse_latest_tag` now returns a reference to one of the items provided in an attempt to be more efficient.

crates/pop-cli/src/commands/mod.rs
```diff
@@ -46,7 +48,7 @@ pub(crate) enum Command {
 	Call(call::CallArgs),
 	#[clap(aliases = ["u", "deploy"], about = about_up())]
 	#[cfg(any(feature = "parachain", feature = "polkavm-contracts", feature = "wasm-contracts"))]
-	Up(up::UpArgs),
+	Up(Box<up::UpArgs>),
```
evilrobot-01, line 51: Resolves clippy warning due to size.

crates/pop-common/src/sourcing/mod.rs
evilrobot-01: Refactored in 7d629ce4 to be more extensible/configurable.

crates/pop-cli/tests/parachain.rs
```diff
@@ -3,58 +3,78 @@
...
 use pop_parachains::Parachain;
-use std::{fs, path::Path, process::Command as Cmd};
+use std::{ffi::OsStr, fs, path::Path, process::Command, thread::sleep, time::Duration};
 use strum::VariantArray;
-use tokio::time::{sleep, Duration};
 
-/// Test the parachain lifecycle: new, build, up, call.
-#[tokio::test]
-async fn parachain_lifecycle() -> Result<()> {
-	let temp = tempfile::tempdir().unwrap();
```
evilrobot-01, line 16: Results in cache being ineffective, specifically in relation to the build of the template. Separated out into its own test.

crates/pop-cli/tests/parachain.rs
```diff
@@ -3,58 +3,78 @@
...
+// Test that all templates are generated correctly
+#[test]
+fn generate_all_the_templates() -> Result<()> {
+	let temp = tempfile::tempdir()?;
 	let temp_dir = temp.path();
-	// let temp_dir = Path::new("./"); //For testing locally
-	// Test that all templates are generated correctly
-	generate_all_the_templates(&temp_dir)?;
```
evilrobot-01, line 20: Separated out into its own test

crates/pop-cli/tests/parachain.rs
```diff
@@ -172,37 +193,20 @@ name = "collator-01"
...
-				&template.to_string(),
-				"--verify",
-			])
-			.assert()
-			.success();
-		assert!(temp_dir.join(parachain_name).exists());
-	}
-	Ok(())
+fn pop(dir: &Path, args: impl IntoIterator<Item = impl AsRef<OsStr>>) -> Command {
+	let mut command = Command::new(cargo_bin("pop"));
```
evilrobot-01, line 208: Use normal command rather than the one provided by assert_cmd in order to get maximum ouput from the test, useful in debugging CI issues. Integration tests can be flaky.

crates/pop-cli/tests/parachain.rs
```diff
@@ -121,26 +140,28 @@ default_command = "./target/release/parachain-template-node"
...
 
 	// `pop up network -f ./network.toml --skip-confirm`
-	let mut cmd = Cmd::new(cargo_bin("pop"))
-		.current_dir(&temp_parachain_dir)
-		.args(&["up", "network", "-f", "./network.toml", "--skip-confirm"])
-		.spawn()
-		.unwrap();
+	let mut command = pop(
+		&working_dir,
+		&["up", "network", "./network.toml", "-r", "stable2412", "--verbose", "--skip-confirm"],
```
evilrobot-01, line 151: 2503-1 doesnt produce blocks with the available paseo-local runtimes, resulting in the call steps failing.

crates/pop-cli/src/commands/mod.rs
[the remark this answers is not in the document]
AlexD10S: No, the logic makes sense.

crates/pop-cli/tests/parachain.rs
AlexD10S: I used `let temp = tempfile::tempdir()?`; because when running the test locally, the parachain folder isn’t cleaned up otherwise. But I see your point

AlexD10S (approved):
Thanks for addressing my points and answering my questions, everything looks good to me. Great work on this! I just left two comments, but they're minor nitpicks.

crates/pop-parachains/src/registry/pop.rs
evilrobot-01: @AlexD10S can this be resolved?

crates/pop-common/src/sourcing/mod.rs
```diff
@@ -241,6 +279,198 @@ impl GitHub {
...
+						.and_then(|pattern| pattern.version(v))
+						.unwrap_or(v)
+						.to_string()
+				});
+
+				// Extract versions from any cached binaries - e.g., offline or rate-limited.
+				for file in read_dir(cache).into_iter().flatten().filter_map(|f| {
+					f.ok()
+						.and_then(|f| f.file_name().into_string().ok())
+						.filter(|f| cache_filter(&f))
```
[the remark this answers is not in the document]
evilrobot-01: Addressed with f5534d43

crates/pop-parachains/src/up/mod.rs
```diff
@@ -1957,45 +2292,457 @@ command = "polkadot-parachain"
...
+							chain: None,
+							chain_spec_generator: None,
+						},
+					),
+				]
+				.into(),
+			)?;
+
+			let contents = adapted.dump_to_toml().unwrap();
+			println!("{contents}");
```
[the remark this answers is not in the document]
evilrobot-01: Completed in 0ea95b6c

crates/pop-cli/tests/parachain.rs
evilrobot-01: Can add an instruction at the end of the test to delete the folder, but then each time the test is run locally one needs to wait for a rebuild and then there is no caching.

evilrobot-01: Hmmm, as it stands the relevant directory is not being cached, presumably as it is buried within the directory structure. https://github.com/r0gue-io/pop-cli/actions/runs/15146706684/job/42584088437

It would therefore need updating on the CI workflow to include so I have reverted the change in 59a0cd9c. Waiting ages for CI to complete can be a chore, but so can having local disk space eaten up unknowingly.

AlexD10S (approved):

crates/pop-parachains/src/registry/mod.rs
```diff
@@ -0,0 +1,354 @@
...
+		],
+	);
+	registry
+};
+
+/// Returns the rollups registered for the provided relay chain.
+///
+/// # Arguments
+/// * `relay` - The relay chain.
+pub fn rollups(relay: &Relay) -> &'static [Box<dyn traits::Rollup>] {
```
chungquantin, line 101: Would it make sense for this method to be implemented for `Relay`? So we can have `Polkadot.rollups()`

tests/networks/paseo+people.toml
```diff
@@ -0,0 +1,20 @@
+# pop up network -f ./tests/networks/paseo+people.toml
```
chungquantin, line 1:
```suggestion
# pop up network ./tests/networks/paseo+people.toml
```

crates/pop-cli/tests/parachain.rs
```diff
@@ -121,26 +146,28 @@ default_command = "./target/release/parachain-template-node"
...
 [[parachains.collators]]
 name = "collator-01"
-"#,
-			random_port
+rpc_port = {random_port}
+"#
 		),
 	)?;
 
 	// `pop up network -f ./network.toml --skip-confirm`
```
chungquantin, line 154: Need to be updated

```
pop up network ./network.toml --skip-confirm
```

crates/pop-parachains/src/up/mod.rs
```diff
@@ -249,54 +230,44 @@ impl Zombienet {
...
 			}
-			if let Some(relay) = relay {
-				return Ok(relay);
-			}
+		}
+		if let Some(relay) = relay {
+			return Ok(relay);
 		}
 		// Otherwise use default
 		Ok(relay::default(version, runtime_version, chain, cache).await?)
```
chungquantin, line 273: Don't need to change in this PR but clippy is not happy with this: 

```
relay::default(version, runtime_version, chain, cache).await
```

crates/pop-parachains/src/up/mod.rs
chungquantin, line 1: When I worked on the benchmarking feature, I realized we have all tests in the same file with the implementation which makes it's quite large (>2500 lines for this file but > 1000 lines only for the tests). Don't need to update in this PR but I think we should refactor all commands in a separate PR following: 

```
up
  - mod.rs : For the implementation
  - tests.rs: For the tests
```

crates/pop-parachains/src/registry/pop.rs
```diff
@@ -0,0 +1,182 @@
...
+	}
+}
+
+impl SourceT for Pop {
+	type Error = Error;
+	/// Defines the source of a binary.
+	fn source(&self) -> Result<Source, Error> {
+		// Source from GitHub release asset
+		let binary = self.binary();
+		Ok(Source::GitHub(ReleaseArchive {
```
chungquantin, line 38: Including other binaries, we have different binaries with different configurations, maybe it's better to have a single file to mange all the configuration for the binary source? 

Not a blocker of this PR and can be tackled separately.

chungquantin (commented):
PR looks awesome. This feature improves the experience a lot because users don't need a network file to launch a network. I left a few nitpicks / improvements. 

The PR is quite big so there might be things overseen, I still need to test the functionality but other than that, I am happy to approve if everything works well.

crates/pop-common/src/sourcing/mod.rs
```diff
@@ -586,37 +856,114 @@ pub(super) mod tests {
...
+			assert!(temp_dir.path().join(format!("{item}-{version}")).exists());
 		}
 		Ok(())
 	}
 
+	#[tokio::test]
+	async fn resolve_from_github_release_archive_works() -> anyhow::Result<()> {
+		let owner = "r0gue-io".to_string();
+		let repository = "polkadot".to_string();
+		let version = "stable2503";
```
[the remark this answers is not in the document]
evilrobot-01: No. There are many instances across the codebase which currently use an outdated version number. Technically speaking, these type of calls should be mocked or moved to integration tests.

Please kindly resolve if you are happy.

crates/pop-cli/tests/parachain.rs
evilrobot-01: Ty, addressed in cde6195a

crates/pop-parachains/src/up/mod.rs
evilrobot-01: Ty, addressed in fa4c2582

crates/pop-parachains/src/registry/pop.rs
evilrobot-01: The point of the refactor was to allow contributions to add supported chains easily by adding a type and implementing the necessary traits. Such a module is therefore focused on one chain only, with it then being added to the global registry with a one liner.

If one extrapolates out to 10 chains, would having a single file with the config for 10 chains not ultimately result in refactoring them out into individual modules, which is currrently the case?

crates/pop-common/src/sourcing/mod.rs
```diff
@@ -251,7 +481,7 @@ impl GitHub {
 /// * `status` - Used to observe status updates.
 async fn from_archive(
 	url: &str,
-	contents: &[(&str, PathBuf)],
+	contents: &[(&str, PathBuf, bool)],
```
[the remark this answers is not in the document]
evilrobot-01: Added `ArchiveFileSpec` struct in a031f07d

crates/pop-parachains/src/up/mod.rs
chungquantin: Thank you. I'm happy to resolve it and we can tackle it later.

crates/pop-parachains/src/registry/pop.rs
chungquantin: I see, thank you.

crates/pop-common/src/polkadot_sdk.rs
```diff
@@ -1,41 +1,42 @@
...
+	)
+	.expect("Valid regex")
+});
+// Regex for v{major}.{minor}.{patch} format
+static VERSION: LazyLock<Regex> = LazyLock::new(|| {
+	Regex::new(r"v(?P<major>\d+)\.(?P<minor>\d+)\.(?P<patch>\d+)(-rc\d+)?").expect("Valid regex")
+});
+
+/// A tuple of version numbers.
+pub type Version = (u32, u32, u32);
```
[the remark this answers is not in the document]
AlexD10S: It seems like a different case, one is about getting the version of a binary, and this looks specific to the Polkadot SDK, right? Up to you, but if that’s the case, no need to reuse it. Keeping it separate makes sense.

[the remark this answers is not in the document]
chungquantin: This is a low impact refactor so up to you on wether it should be refactored or not. Feel free to resolve this.

chungquantin (approved):
LGTM! Thanks for applying my feedback. Functionalities work well. Can't wait to see this feature used in production. 

Edit: try runtime test is failing, probably due to the runtime used in the tests upgraded. We can fix it in a separate PR to avoid nuances.

AlexD10S (approved):

PR #527, opened 2025-04-16
chore: update required version of the frame-omni-bencher binary

chungquantin:
Update the version tests of the binaries `frame-omni-bencher` and `try-runtime`. The idea is, whatever versions above the threshold one are accepted and source binaries if the deprecated version of the binaries used.

If there is a break in the future version of the binary, we update the threshold version and change the code to support newer versions.

### DIFF
--- a/crates/pop-parachains/src/bench/binary.rs
+++ b/crates/pop-parachains/src/bench/binary.rs
@@ -83,8 +83,8 @@ mod tests {
 	async fn omni_bencher_generator_works() -> Result<(), Error> {
 		let temp_dir = tempdir()?;
 		let temp_dir_path = temp_dir.into_path();
-		let version = "polkadot-stable2412-4";
-		let binary = omni_bencher_generator(temp_dir_path.clone(), None).await?;
+		let version = "polkadot-stable2412";
+		let binary = omni_bencher_generator(temp_dir_path.clone(), Some(version)).await?;
 		assert!(matches!(binary, Binary::Source { name: _, source, cache }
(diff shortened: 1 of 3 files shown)


crates/pop-parachains/src/bench/binary.rs
```diff
@@ -83,7 +83,7 @@ mod tests {
 	async fn omni_bencher_generator_works() -> Result<(), Error> {
 		let temp_dir = tempdir()?;
 		let temp_dir_path = temp_dir.into_path();
-		let version = "polkadot-stable2412";
+		let version = "polkadot-stable2412-4";
 		let binary = omni_bencher_generator(temp_dir_path.clone(), None).await?;
```
AlexD10S, line 87: Should not be this specified in the line below: 
```
let binary = omni_bencher_generator(temp_dir_path.clone(), version).await?;
```
Otherwise the test will fails once it gets updated no?

chungquantin: > Should not be this specified in the line below:
> 
> ```
> let binary = omni_bencher_generator(temp_dir_path.clone(), version).await?;
> ```
> 
> Otherwise the test will fails once it gets updated no?

You're right. I update the PR to specify the version instead: https://github.com/r0gue-io/pop-cli/pull/527/commits/b9092ab643d8f016a5a2a58f0dfdd0d1458b74ee

AlexD10S (approved):
Thanks for addressing the feedback so quickly. LGTM ✅

PR #533, opened 2025-04-23
ci: update init to use latest rust-cache action

evilrobot-01:
Noticed some caching issues in the CI runs due to sunsetting of old GitHub cache actions (https://github.blog/changelog/2025-03-20-notification-of-upcoming-breaking-changes-in-github-actions/#decommissioned-cache-service-brownouts), so updating to the latest version of the rust-cache action.

Also clears additional space.


AlexD10S (approved):
LGTM

PR #534, opened 2025-05-05
fix: test create_snapshot with relaychain URL

chungquantin:
Fetching states from Pop Network on PASEO does not work as expected. Updating the URL to test with Paseo Relaychain. This is a temporary fix to unblock the CI.


chungquantin:
Close to fix with integration-tests

PR #537, opened 2025-05-13
fix: update branch used to fetch ink! v5 templates

AlexD10S:
This PR fixes a bug reported in the ink! Telegram group, where ink! v5 templates were not loading correctly due to the code fetching from the wrong branch. It was basically generating the template from ink! v6 branch instead of the ink! v5.
The fix updates the logic to fetch templates from the correct branch for ink! v5.

Additionally, a test has been added to ensure this issue doesn’t happen again by validating the correct version after creating a contract.

This PR also bumps the version to `0.8.1` to publish the release with the fix.

### DIFF
--- a/crates/pop-contracts/src/new.rs
+++ b/crates/pop-contracts/src/new.rs
@@ -1,6 +1,8 @@
 // SPDX-License-Identifier: GPL-3.0
 
-use crate::{errors::Error, templates::V6_CONTRACTS_BRANCH, utils::canonicalized_path, Contract};
+use crate::{errors::Error, utils::canonicalized_path, Contract};
+#[cfg(feature = "v6")]
+use crate::templates::V6_CONTRACTS_BRANCH;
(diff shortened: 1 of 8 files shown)


crates/pop-contracts/src/new.rs
```diff
@@ -68,6 +70,9 @@ fn create_template_contract(
 	let template_repository = template.repository_url()?;
 	// Clone the repository into the temporary directory.
 	let temp_dir = ::tempfile::TempDir::new_in(std::env::temp_dir())?;
+	#[cfg(feature = "v5")]
+	Git::clone(&Url::parse(template_repository)?, temp_dir.path(), None)?;
```
AlexD10S, line 74: For ink! v5 fetch it from main: https://github.com/use-ink/ink-examples
For ink! v6 from: https://github.com/use-ink/ink-examples/tree/v6.x

PR #538, opened 2025-05-15
build(deps): consolidate sp-weights dependency

evilrobot-01:
Received an error when using `cargo metadata` due to two instances of the `sp-weights` dependency with the exact same version.


al3mart (approved):
Nice catch!

PR #539, opened 2025-05-16
test(parachains/bench): fix expected error in load_pallet_extrinsics test

evilrobot-01:
Fixes the current failing test by updating the expected value.


AlexD10S (approved):
LGTM!

chungquantin (approved):
Thanks for the fix!

PR #540, opened 2025-05-16
ci: optimize polkavm contract jobs

evilrobot-01:
The polkavm contract jobs are currently running all tests in the project rather than only those which actually have the specified features. That is, parachain jobs are being run in parallel unnecessarily.

This PR simply constrains these jobs to only run over the `pop-cli` and `pop-contracts` crates where these features exist.


AlexD10S (approved):
This change makes a lot of sense, thanks for taking care of it!

PR #547, opened 2025-06-17
feat(bench): support benchmarking multiple pallets

chungquantin:


AlexD10S (commented):
The `update_excluded_extrinsics_works` unit test is failing: https://github.com/r0gue-io/pop-cli/actions/runs/15891493397/job/44815303172

PR #578, opened 2025-07-28
feat: add interactive mode to 'pop new' command

### LINKED ISSUE #575: `pop new`: interactive
Then the interactive UI asks you for whether to create a new contract or chain.

forever8896:
Makes the 'command' field optional in NewArgs, allowing users to run 'pop new' without subcommands to get a beautiful interactive selection with emojis and engaging descriptions.

- 🌐 Parachain: Build your own blockchain with custom logic, token economics, and governance
- 🧩 Pallet: Create reusable blockchain modules to add features like tokens, voting, or custom logic
- 📝 Smart Contract: Write decentralized applications with ink! for the Polkadot ecosystem

Includes updated test patterns and display formatting for the optional command.

Resolves #575

image

### DIFF
--- a/crates/pop-cli/src/commands/new/mod.rs
+++ b/crates/pop-cli/src/commands/new/mod.rs
@@ -1,9 +1,9 @@
 // SPDX-License-Identifier: GPL-3.0
 
-use clap::{Args, Subcommand};
-use std::fmt::{Display, Formatter, Result};
 use crate::cli::{traits::Cli as _, Cli};
 use anyhow::Result as AnyhowResult;
+use clap::{Args, Subcommand};
+use std::fmt::{Display, Formatter, Result};
 
 #[cfg(any(feature = "polkavm-contracts", feature = "wasm-contracts"))]
 pub mod contract;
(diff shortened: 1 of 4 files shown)


crates/pop-cli/src/commands/new/mod.rs
```diff
@@ -62,6 +64,80 @@ impl Display for Command {
...
+	Cli.intro("🚀 Welcome to Pop CLI!")?;
+	
+	let mut prompt = cliclack::select("What would you like to create?".to_string());
+	
+	// Add available options based on features
+	#[cfg(feature = "parachain")]
+	{
+		prompt = prompt.item(
+			"parachain", 
+			"🌐 Parachain",
```
AlexD10S, line 78: Would you mind using Chain instead of Parachain here? There's a PR coming in soon https://github.com/r0gue-io/pop-cli/pull/564 that renames it, so updating this now would save an extra step later

crates/pop-cli/src/commands/new/mod.rs
```diff
@@ -62,6 +64,80 @@ impl Display for Command {
...
+	
+	let mut prompt = cliclack::select("What would you like to create?".to_string());
+	
+	// Add available options based on features
+	#[cfg(feature = "parachain")]
+	{
+		prompt = prompt.item(
+			"parachain", 
+			"🌐 Parachain", 
+			"Build your own blockchain with custom logic, token economics, and governance"
```
AlexD10S, line 79: I wouldn't be that specific with token economics and governance in this message: 
> Build your own custom  chain

crates/pop-cli/src/commands/new/mod.rs
```diff
@@ -62,6 +64,80 @@ impl Display for Command {
...
+			"Create reusable blockchain modules to add features like tokens, voting, or custom logic"
+		);
+	}
+	
+	#[cfg(any(feature = "polkavm-contracts", feature = "wasm-contracts"))]
+	{
+		prompt = prompt.item(
+			"contract", 
+			"📝 Smart Contract", 
+			"Write decentralized applications with ink! for the Polkadot ecosystem"
```
AlexD10S, line 93: Same as above, wouldn't be that specific here with the Polkadot ecosystem:
> Write ink! smart contracts

crates/pop-cli/src/commands/new/mod.rs
```diff
@@ -62,6 +64,80 @@ impl Display for Command {
...
+	{
+		prompt = prompt.item(
+			"parachain", 
+			"🌐 Parachain", 
+			"Build your own blockchain with custom logic, token economics, and governance"
+		);
+		prompt = prompt.item(
+			"pallet", 
+			"🧩 Pallet", 
+			"Create reusable blockchain modules to add features like tokens, voting, or custom logic"
```
AlexD10S, line 84: Same as above, I wouldn't be that specific with the examples. Shorter:
`Create reusable and customizable chain modules`

crates/pop-cli/src/commands/new/mod.rs
AlexD10S, line 78: Just for consistency with the other messages, could you remove the emojis from the names?

AlexD10S (commented):
Great to see you contributing here, thanks for helping improve Pop CLI!
To fix the CI, just run:
```
cargo +nightly fmt --all
```
That will automatically take care of the formatting.

I also left a few comments on the output messages to keep things consistent with the rest of the CLI.

crates/pop-cli/src/commands/new/mod.rs
forever8896: Should I change the value as well as the visual display? As changing the value would break my PR functionality, but it would work on the merge of the #564

forever8896: image

Then I believe its done!

PR #592, opened 2025-08-15
Add command to convert from/to Ethereum addresses

### LINKED ISSUE #566: feat(utils): add utility to convert Ethereum (H160) address to Substrate (SS58) address
With the Polkadot Hub supporting EVM compatibility, it would be helpful to provide a quick way to convert an Ethereum-style H160 address (e.g. a deployed contract address) into its corresponding Substrate SS58 format.

This could be especially useful for contract developers who are testing on Polkadot Hub.

There are already some community UIs that do this:
- https://ggwpez.github.io/evm-to-ss58/
- https://eth-ksm-convert.vercel.app/
- https://www.shawntabrizi.com/substrate-js-utilities/

It would be great to support this conversion directly in pop-cli, perhaps via a simple command.

Moliholy:
Closes #566

This PR address the `convert` command to initially convert fromEthereum address format to Substrate format and viceversa.

However, and merely as a proposal, I think this command should be a gateway to the same (or more!) functionality as the one we can find in the [Substrate js-utilities](https://www.shawntabrizi.com/substrate-js-utilities/).

Also, this `convert` command could be perfectly isolated as a separate feature. Would be fine for me to address that, but firstly it'd be good go have some feedback.

## Usage

```sh
# Convert from Substrate address to Ethereum address:
pop convert address 13dKz82CEiU7fKfhfQ5aLpdbXHApLfJH5Z6y2RTZpRwKiNhX
0x742d35cc6634c0532925a3b844bc454e4438f44e
```

```sh
# Convert from Ethereum address to Substrate address with prefix 0 (Polkadot).
pop convert address 0x742d35Cc6634C0532925a3b844Bc454e4438f44e          
13dKz82CEiU7fKfhfQ5aLpdbXHApLfJH5Z6y2RTZpRwKiNhX
```

```sh
# Convert from Ethereum address to Substrate address with prefix 42 (Generic).
pop convert address 0x742d35Cc6634C0532925a3b844Bc454e4438f44e 42   
5Eh2qnm8NwCeDnfBhm2aCfoSffBAeMk914NUs8UDGLuoY6qg
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,6 +31,7 @@ env_logger = { version = "0.11.7", default-features = false }
 flate2 = "1.0.30"
 git2 = { version = "0.18", default-features = true, features = ["vendored-openssl"] }
 glob = { version = "0.3.1", default-features = false }
+hex = { version = "0.4.3", default-features = false }
 log = { version = "0.4.20", default-features = false }
 mockito = { version = "1.4.0", default-features = false }
 tar = { version = "0.4.40", default-features = false }
--- a/crates/pop-cli/Cargo.toml
+++ b/crates/pop-cli/Cargo.toml
@@ -20,6 +20,7 @@ console.workspace = true
 dirs.workspace = true
 duct.workspace = true
 env_logger.workspace = true
+hex.workspace = true
 os_info.workspace = true
 reqwest.workspace = true
 serde = { workspace = true, features = ["derive"] }
--- a/crates/pop-cli/src/commands/convert.rs
+++ b/crates/pop-cli/src/commands/convert.rs
@@ -1,0 +1,121 @@
+// SPDX-License-Identifier: GPL-3.0
+
+use self::Command::*;
+use super::*;
+use crate::cli::traits::Cli;
+use anyhow::Result;
+use clap::Args;
+use hex;
+use regex::Regex;
+use sp_core::crypto::{AccountId32, Ss58Codec};
+
+const ETHEREUM_ADDRESS_REGEX: &str = "^0x[0-9a-fA-F]{40}$";
+const EE_BYTE: u8 = 0xEE;
+const DEFAULT_POLKADOT_SS58_PREFIX: u16 = 0;
+
+fn convert_address(address: &str, ss58_prefix: Option<u16>) -> Result<String> {
+    let eth_regex = Regex::new(ETHEREUM_ADDRESS_REGEX)?;
+
+    if eth_regex.is_match(address) {
+        let mut raw_bytes = hex::decode(&address[2..])?;
+        raw_bytes.extend_from_slice(&[EE_BYTE; 12]);
+
+        // Convert H256 to AccountId32 first
+        let account_id = AccountId32::new(raw_bytes[..].try_into()?);
+        let version = ss58_prefix.unwrap_or(DEFAULT_POLKADOT_SS58_PREFIX);
+        let ss58_address = account_id.to_ss58check_with_version(version.into());
+        Ok(ss58_address)
+
+    } else {
+        // Try to decode SS58 address
+        let account_id = AccountId32::from_ss58check(address)?;
+        let bytes: [u8; 32] = account_id.into();
+
+        // Verify last 12 bytes are 0xEE
+        if !bytes[20..].iter().all(|&b| b == EE_BYTE) {
+            return Err(anyhow::anyhow!("Invalid address: last 12 bytes must be 0xEE"));
+        }
+
+        // Take first 20 bytes and format as hex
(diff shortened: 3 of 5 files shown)


Cargo.toml
```diff
@@ -31,6 +31,7 @@ env_logger = { version = "0.11.7", default-features = false }
 flate2 = "1.0.30"
 git2 = { version = "0.18", default-features = true, features = ["vendored-openssl"] }
 glob = { version = "0.3.1", default-features = false }
+hex = { version = "0.4.3", default-features = false }
```
AlexD10S, line 34: Instead of adding a new library can we use `sp_core`?
```
use sp_core::bytes::{from_hex, to_hex};
```

crates/pop-cli/Cargo.toml
```diff
@@ -64,12 +65,13 @@ subxt-signer.workspace = true
 default = ["chain", "telemetry", "wasm-contracts"]
 contract = ["wasm-contracts"]
 contracts = ["polkavm-contracts"]
-experimental = ["hashing"]
+experimental = ["hashing", "convert"]
```
AlexD10S, line 68: So cool that you added it under the experimental feature flag as `hashing`. We can leave it like this for this PR, but as you also mentioned in the PR description  we should merge both so the "utility" command becomes something like:

```
pop convert address 0x742d35Cc6634C0532925a3b844Bc454e4438f44e
pop convert hash twox 128 some_text
```

For context, this was the idea behind putting `hashing` under `experimental`: [https://github.com/r0gue-io/pop-cli/pull/517#issuecomment-2785593196](https://github.com/r0gue-io/pop-cli/pull/517?utm_source=chatgpt.com#issuecomment-2785593196)

crates/pop-cli/src/commands/convert.rs
```diff
@@ -0,0 +1,139 @@
+// SPDX-License-Identifier: GPL-3.0
+
+use self::Command::*;
+use super::*;
+use crate::cli::traits::Cli;
+use anyhow::Result;
+use clap::Args;
+use hex;
```
AlexD10S, line 8: As above, we probably can use:
`use sp_core::bytes::{from_hex, to_hex}; `

AlexD10S (commented):
Excellent job! I really like the feature, I believe it will be widely used. The DevEx is very smooth, the user doesn’t have to specify the input format, the CLI already recognizes it.
The code is solid and is well tested. The only comment I left was about the new library that was imported, which might not be necessary.

I do have one question from testing: when I try with Alice’s address

```
pop convert address 5GrwvaEF5zXb26Fz9rcQpDWS57CtERHpNehXCPcNoHGKutQY
```
I get the following error:
```
Error: Invalid address: last 12 bytes must be 0xEE
```

Cargo.toml
Moliholy: Absolutely right, I didn't realize of that when exploring the codebase. Done in [130b650](https://github.com/r0gue-io/pop-cli/pull/592/commits/130b650bcb46a12de6e4085d8d6ccf685f37aacf).

crates/pop-cli/Cargo.toml
Moliholy: > For context, this was the idea behind putting hashing under experimental

Heh, I didn't know that, but it makes total sense. If so, it'd be good to create issues for the remaining functionality that should land in that command. But I'll leave that up to the team to discuss the details of that.

Anyhow, I think the community would find valuable not to be forced to leave the terminal to a web page or random source to perform these kind of operations.

Moliholy:
@AlexD10S some minor detail regarding this development: 

Screenshot 2025-08-29 at 00 23 20

The output, as you can see, is not purely the resulting address. For this very particular case (and hashing, since we are on it) in my opinion the output should be totally clean, similar to the result of an standard `println!` statement. Otherwise, results cannot be piped, hence making life harder when integrating pop-cli in shell scripts.

I didn't do it here, as I wanted to respect the similar development for hashing, but it might be worth to create an issue so that this does not get forgotten.

crates/pop-cli/Cargo.toml
```diff
@@ -63,8 +63,9 @@ subxt-signer.workspace = true
 default = ["chain", "telemetry", "wasm-contracts"]
 contract = ["wasm-contracts"]
 contracts = ["polkavm-contracts"]
-experimental = ["hashing"]
```
AlexD10S, line 66: It needs rebase here after merging https://github.com/r0gue-io/pop-cli/commits/main/

AlexD10S (approved):
LGTM! 🚀
One follow-up: we’ll also need to document this new command in the docs repo: https://github.com/r0gue-io/pop-docs

PR #606, opened 2025-09-02
feat: make `pop up <contract>` interactive

### LINKED ISSUE #502: feat: guide UI in pop up contract
Currently, UI guidance exists within various commands.
 Introducing a similar UI guidance system specifically for ` pop up` contracts will improve the DevEx. 

The implementation will reuse existing logic for parsing metadata, similar to the `pop call contract` functionality. The primary difference will be parsing and prompting for parameters required in `constructors` instead of `message` parameters.

moliholy:
Closes #502 

This PR makes the `pop up path/to/my/contract` command interactive.

Since I was on it, I also removed some redundant code, which got encapsulated in utility functions.

## Example

Screenshot 2025-09-03 at 01 27 54

### DIFF
--- a/crates/pop-cli/src/common/contracts.rs
+++ b/crates/pop-cli/src/common/contracts.rs
@@ -142,8 +142,11 @@ mod tests {
 	use crate::cli::MockCli;
 	use duct::cmd;
 	use pop_common::{find_free_port, set_executable_permission};
-	use pop_contracts::{is_chain_alive, run_contracts_node};
-	use std::fs::{self, File};
+	use pop_contracts::{extract_function, is_chain_alive, run_contracts_node, FunctionType};
+	use std::{
+		env,
+		fs::{self, File},
+	};
 	use url::Url;
 
 	#[test]
@@ -165,6 +168,21 @@ mod tests {
 		File::create(contract_path.join(format!("target/ink/{}.contract", name)))?;
 		assert!(has_contract_been_built(Some(&path.join(name))));
 		Ok(())
+	}
+
+	#[tokio::test]
+	async fn request_contract_function_args_works() -> anyhow::Result<()> {
+		let mut current_dir = env::current_dir().expect("Failed to get current directory");
+		current_dir.pop();
+		let mut cli = MockCli::new()
+			.expect_input("Enter the value for the parameter: init_value", "true".into());
+		let function = extract_function(
+			current_dir.join("pop-contracts/tests/files/testing.json"),
+			"new",
+			FunctionType::Constructor,
+		)?;
+		assert_eq!(request_contract_function_args(&function, &mut cli)?, vec!["true"]);
+		cli.verify()
 	}
 
 	#[tokio::test]


crates/pop-contracts/src/utils/metadata.rs
```diff
@@ -131,11 +133,8 @@ fn get_contract_functions(
 /// # Arguments
 /// * `path` -  Location path of the project or contract artifact.
 /// * `message` - The label of the contract message.
-pub fn get_message<P>(path: P, message: &str) -> Result<ContractFunction, Error>
-where
-	P: AsRef<Path>,
-{
-	get_messages(path.as_ref())?
+pub fn get_message(path: &Path, message: &str) -> Result<ContractFunction, Error> {
```
AlexD10S, line 136: Curious about the motivation for this change, was it mainly for simplicity? The previous implementation accepted anything convertible to a `Path` (`&str`, `String`, `&Path`, `PathBuf`), which felt a bit more flexible.

crates/pop-cli/src/common/contracts.rs
```diff
@@ -89,6 +89,33 @@ pub fn has_contract_been_built(path: Option<&Path>) -> bool {
...
 
+/// Requests and collects function arguments from the user via CLI interaction.
+///
+/// # Arguments
+/// * `function` - The contract function containing argument definitions.
+/// * `cli` - Command line interface implementation for user interaction.
+///
+/// # Returns
+/// A vector of strings containing the user-provided argument values.
+pub fn request_contract_function_args(
```
AlexD10S, line 100: This function can have its own unit test that tests the CLI output. 
As an example the one below:
https://github.com/r0gue-io/pop-cli/blob/main/crates/pop-cli/src/common/contracts.rs#L146

crates/pop-cli/src/commands/call/contract.rs
```diff
@@ -300,19 +300,7 @@ impl CallContractCommand {
...
-				.placeholder(&format!("Type required: {}", arg.type_name));
-
-			// Set default input only if the parameter type is `Option` (Not mandatory)
-			if arg.type_name.starts_with("Option<") {
-				input = input.default_input("");
-			}
-			contract_args.push(input.interact()?);
-		}
-		self.args = contract_args;
+		self.args = request_contract_function_args(&message, cli)?;
```
AlexD10S, line 303: Small clippy warning

AlexD10S (commented):
Really enjoyed reviewing this PR, super clean and easy to follow 👏. I only left a small note about a unit test, and GitHub triggers a clippy warning.

I tested the behaviour and the DevEx feels great, nice work! 🎉

One thought: a contract could technically have multiple constructors, but probably not worth addressing now. Since ink! v6 is moving toward Solidity compatibility with a single constructor, the current approach (using the `default`) seems like the best DevEx. And if needed, users can always specify a different constructor with the `--constructor` flag.

crates/pop-cli/src/commands/call/contract.rs
Moliholy: Shouldn't this fail in the CI? I am actually seeing many things that could be autofixed with clippy 🤔  that are not yelling in the CI.

Moliholy: Should be fixed in [d31d86f](https://github.com/r0gue-io/pop-cli/pull/606/commits/d31d86f806ed1ed7337f2a078ed302ae36f1d71a).

Moliholy: In my opinion, yes. Otherwise we leave clippy issues around and never bother to fix them.

PR #614, opened 2025-09-04
feat(cli): clearly highlight writing operations when calling smart contracts

### LINKED ISSUE #611: Distinguish between read and write operations in contract calls
Currently both read and write operations look the same in the list.

## Example

Image

Here we can see that `flip` is a write operation, as it mutates the storage. It will broadcast an extrinsic and won't return any data other than the consumed weight.

On the other hand, the `get` is an actual call that does not cost any fee and returns the value of the boolean stored in the contract.

I propose to differentiate them by:

1) Placing an icon/emoji on the left depending on whether it's a read or write operation.
2) Showcasing at the beginning of the right-most description of the call that, in case of a write operation, it will cost fee (or simply that it is a write operation).

Moliholy:
Closes #611

This PR showcases those operations in smart contracts that imply sending an extrinsic from those that merely read from the storage.

## Screenshots

fb46c62a-da17-4618-bebb-65d39bb82877
fb46c62a-da17-4618-bebb-65d39bb82877

### DIFF
--- a/crates/pop-cli/src/cli.rs
+++ b/crates/pop-cli/src/cli.rs
@@ -771,7 +771,7 @@ pub(crate) mod tests {
 			// Check expectations
 			if let Some(items) = self.items_expectation.as_mut() {
 				let item = (label.to_string(), hint.to_string());
-				assert!(items.contains(&item), "`{item:?}` item does not satisfy any expectations");
+				assert!(items.contains(&item), "`{item:?}` item does not satisfy any expectations.\nAvailable expectations:\n{items:#?}");
 				items.retain(|x| *x != item);
 			}
 			// Collect if specified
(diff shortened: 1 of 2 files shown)


crates/pop-cli/src/cli.rs
```diff
@@ -771,7 +771,7 @@ pub(crate) mod tests {
 			// Check expectations
 			if let Some(items) = self.items_expectation.as_mut() {
 				let item = (label.to_string(), hint.to_string());
-				assert!(items.contains(&item), "`{item:?}` item does not satisfy any expectations");
+				assert!(items.contains(&item), "`{item:?}` item does not satisfy any expectations.\nAvailable expectations:\n{items:#?}");
```
AlexD10S, line 774: That's very useful for us, thanks!

AlexD10S (approved):
Nice!
I tested it on a contract with a long list of messages, and it’s really useful for quickly seeing which ones write to state and which ones are read-only.

PR #617, opened 2025-09-05
Optimize CI with already built docker image

Moliholy:


Moliholy:
@AlexD10S this is an alternative PR to #615.

Issue #579, opened 2025-07-29
Improve input `None` as contract argument


AlexD10S:
Currently for a smart contract that has an `Option` parameter, is not intuitive how to add `None`.
Example: 
```rs
 #[ink(storage)]
    pub struct TestContract {
        name: String,
        otherContract: Option<H160>,
    }

    impl TestContract {
        #[ink(constructor)]
        pub fn new(name: String,  otherContract: Option<H160>) -> Self {
            Self { name, otherContract }
        }
```
The `None` value has to be indicated with `""` e.g: `pop up --args \"name\" ""` . This is because cargo-contract library is implemented that way, to improve the DevEx why not let the user use None and convert it under the hood to "".

Also the experience to input a String has to be improved too, to don't use `\"name\"`: https://github.com/r0gue-io/pop-cli/issues/408

Issue #408, opened 2025-02-11
Strings are not escpade.


frankdeno:
### Hola

**Problem**: Pop CLI on contract call doesn't escape Quotes.
**Current solution**: copy  command and add escape characters manually.
**Expected**: It escapes characters.

#### Screenshot

![Screenshot](https://i.ibb.co/vvqHSm5z/image.png)

#### Replicate

1. `git clone https://gitlab.com/frankbevr/006-DeShare.git` 
2. `cd 006-DeShare/contract/`
3. `pop up contract -a 10 5GrwvaEF5zXb26Fz9rcQpDWS57CtERHpNehXCPcNoHGKutQY` 
4. Copy contract addess
5. `pop call contract -p ./target/ink/de_share.contract` 
6. Arguments are a number and a string.
7. Read: Anyhow error: Expected a String value

Thanks for tooling!

### Cheers

AlexD10S:
Hi @frankdeno ! Nice to see you testing pop-cli, and thanks for reporting the issue.

For some reason, I can't access the contract from that link. Could you provide the code here or at least the function you're trying to call?

Also, if you have the latest version of `pop-cli`, you can try simply running `pop call contract` and follow the interactive UI to call the contract.

frankdeno:
@AlexD10S 

### Hoi Alex 👋 

Huch^^ 
I forget to make it public. 
https://gitlab.com/frankbevr/006-DeShare
Now it should work. 

Thanks for letting me know. I will try that out. Ha, I remember, yes I did that, but then I have to always give the path, so I decided to use the argument. I guess. Memory is kind of blurry even that's just some days ago 🙈 

### SideQuestion:

While writing, something just rushed thruh my head. What is the best approach to create something like

```mermaid
classDiagram
    class Car {
        +id: string
        +model: string
        +status: string
    }

    class Customer {
        +id: string
        +name: string
    }

    class Rental {
        +id: string
        +car: Car
        +customer: Customer
        +startDate: string
        +endDate: string
    }

 Car  -- Rental
 Customer --  Rental

``` 
Basically how do I hammer a object (f.e. Car) im my storage (f.e. Rental) without breaking myself a bone.  : D
In solidity it is quite straightforward but in ink! it always gets quite ugly  🙈  Especially when it starts with Mapping<id, Car> or something like that. 

Refrences: 
[...]

PR #618, opened 2025-09-05
fix: process user input for Strings and None values

AlexD10S:
This PR improves the handling of user-provided arguments for contract calls:

- Strings: User input is automatically normalized to ensure values are wrapped in double quotes (if not already).
Closes https://github.com/r0gue-io/pop-cli/issues/408

- None values: User input "None" is automatically normalized to an empty string ("").
Closes https://github.com/r0gue-io/pop-cli/issues/579


Moliholy (approved):
LGTM! Very clean and straightforward PR ✨

PR #620, opened 2025-09-05
feat(cli): do not ask to run with sudo for every call

### LINKED ISSUE #610: Do not ask for sudo confirmation
When executing a call in a standard scenario users are not meant to use sudo, and presumably only a tiny percentage of users will want to use sudo.

Hence, it makes little sense that the client is constantly asking for whether users want to execute a call with sudo or not. I propose to remove the question, and only if the user specifies the `--sudo` call and then check whether the chain supports it or not.

Moliholy:
Closes #610

This PR avoids asking every time to the user wether the call should be run with sudo or not.

Instead, it simply asks the user to confirm its decision if the `--root` flag is specified.

Example:

Screenshot 2025-09-05 at 17 22 53

### DIFF
--- a/crates/pop-cli/src/commands/call/chain.rs
+++ b/crates/pop-cli/src/commands/call/chain.rs
@@ -200,8 +200,8 @@ impl CallChainCommand {
 
 			// If chain has sudo prompt the user to confirm if they want to execute the call via
 			// sudo.
-			if !self.skip_confirm {
-				self.configure_sudo(chain, cli)?;
+			if self.sudo {
+				self.check_sudo(chain, cli)?;
 			}
 
 			let (use_wallet, suri) = self.determine_signing_method(cli)?;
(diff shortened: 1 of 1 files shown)


crates/pop-cli/src/commands/call/chain.rs
```diff
@@ -289,26 +289,20 @@ impl CallChainCommand {
...
-						)
-						.initial_value(false)
-						.interact()?;
-				},
+			Ok(_) => {
+				self.sudo = cli
+					.confirm(
+						"Are you sure you want to dispatch this function call with `Root` origin?",
+					)
+					.initial_value(false)
```
AlexD10S, line 299: Nitpick: When the user passes `--sudo`, the confirmation prompt should default to `Yes`. 

Also, if `self.skip_confirm` is true, we should not prompt at all. This is important for example for a CI, where `--skip-confirm` avoids blocking on interactive input.

AlexD10S (approved):
LGTM!!

PR #622, opened 2025-09-08
feat(cli): follow logs of contracts node before termination

### LINKED ISSUE #555: Kill the ink-node automatically
Found here: https://learn.onpop.io/contracts/guides/deploy-your-contract-locally/deploy-locally-on-a-solochain

> Note: The contracts node will be running in the background of your machine. If at anytime you need to kill the process you can do the following:
Find the process ID by running the following command: lsof -i :9944
Kill the process by running the kill command and passing in the ID of the process:
kill -9 3537

I think whenever we spin up the ink-node for the users, the terminal should change to an interactive mode. If the user than ends it with ctrl-c the ink-node is automatically killed as well.

moliholy:
Closes #555

This PR allows users to follow the logs instead of detaching the contracts node. This way users have access to the logs and can keep interacting with the node.

## Example

Screenshot 2025-09-08 at 22 33 03

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -42,7 +42,7 @@ tracing-subscriber = { version = "0.3.19", default-features = false }
 
 # networking
 reqwest = { version = "0.12", default-features = false, features = ["default-tls", "json", "multipart", "stream"] }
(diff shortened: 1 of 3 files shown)


AlexD10S (approved):
Nice one!

PR #627, opened 2025-09-09
chore: remove cargo deny job from the CI

AlexD10S:
Remove the `deny` job from CI: it’s producing frequent  failures due to new upstream advisories on transitive deps, even when none new dependency has been added.

I keep `deny.toml` in the repo for local checks and future re-enablement with a quick explanation on how to use it in the README.


Moliholy:
@AlexD10S you can run a separate workflow that can only be run manually, if you want.

PR #637, opened 2025-09-20
Remove build command from readme that doesn't build

talhadaar:
There are several items with the same name that are build only under certain feature tags. Building with --all-features flag causes build to fail as all items will similar names come into scope.


AlexD10S (approved):
Sorry for the delay, we were investigating why the CI was failing, but it’s unrelated to this PR. Thanks a lot for the fix and for contributing to `pop-cli`!

PR #640, opened 2025-09-25
chore: `pop build spec` improvements

### LINKED ISSUE #642: Make all pop commands shell-compatible
Currently there are cases where parameters in pop commands do take a default value, yet the program asks users for their preferred choice regarding the value of the parameter. This fact prevents shell scripts to effectively use the tool, which is inconvenient. This is particularly notorious when using boolean parameters with a default value (typically `false`). For example, in `pop build spec` you cannot run it without `genesis-state` and you specify `--genesis-state=false` then the tool keeps asking you whether to generate the genesis state or not.

For cases like this one we should use an `Option<bool>`, which leads to three variants:

1) `None` -> the user did not select anything regarding the parameter -> ask the user.
2) `Some(false)` -> the user explicitly disabled the functionality -> do not ask.
3) `Some(true)` -> the user explicitly enabled the functionality -> do not ask.

This may happen in other pop commands, so this is the parent issue for all those cases where we identify this issue. Once all commands are addressed, we may close this issue.

moliholy:
Closes #642.

This PR enhances the `pop build spec` so that it can:

- Specify the base directory (`path` argument) instead of assuming it is the current directory. This allow the `pop build spec` to work from any directory in the system. By default the current directory keeps being used if not specified.
- Changes clap arguments to allow the command to be executed in a non-interactive manner. This way the command can be embedded in bash scripts, for example.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,3 +1,2 @@
 @AlexD10S
 @al3mart
-@evilrobot-01
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,8 +19,10 @@ env:
   CARGO_TERM_COLOR: always
   GITHUB_ACTOR: pop-cli
   CARGO_INCREMENTAL: 1
-  RUST_VERSION: 1.68
   RUST_BACKTRACE: 1
+  # It is important to always use the same flags. Otherwise, the cache will not work.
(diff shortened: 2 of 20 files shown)


AlexD10S (approved):
LGTM!

Moliholy:
@AlexD10S mind to merge? I don't have permission to do it if the build is failing.

AlexD10S (approved):

Issue #607, opened 2025-09-03
Make polkavm the default


Daanvdplas:
Potentially remove wasm functionality.

AlexD10S:
> [@Daanvdplas](https://github.com/Daanvdplas) we'll have to wait for a stable release of `cargo-contract@6.0.0` before doing so.

Exactly. We already have a draft PR pointing to the latest changes in `ink!` and `cargo-contract`: https://github.com/r0gue-io/pop-cli/pull/584. 
The plan is to update it and merge it once a stable release is available on crates.io and then deprecate the WASM functionality.

PR #645, opened 2025-09-28
feat: using ink! `v6.0.0 alpha.4`

AlexD10S:
This branch allows us to use `pop-cli` with the latest ink! `v6.0.0-alpha.4`. 
Due to dependency changes and conflicts, it was simpler to remove the feature flag and maintain support for a single contract type in this PR (v6): v6 which closes https://github.com/r0gue-io/pop-cli/issues/607

In draft because` v6.0.0 alpha.4` needs to be publish in crates.io before merging this PR


crates/pop-cli/src/commands/build/contract.rs
```diff
@@ -10,6 +10,8 @@ pub struct BuildContract {
 	pub(crate) path: PathBuf,
 	/// Build profile: `true` for release mode, `false` for debug mode.
 	pub(crate) release: bool,
+	/// Which specification specification to use for contract metadata.
```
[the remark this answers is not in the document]
AlexD10S: Fixed in https://github.com/r0gue-io/pop-cli/pull/645/commits/3eb6afd115621e027ff130c3eea15a87712a7029

.github/workflows/coverage.yml
```diff
@@ -68,7 +68,7 @@ jobs:
       - name: Generate code coverage
         run: |
           cargo llvm-cov nextest --workspace --lib --bins --no-default-features       \
-          --features "default chain telemetry v5 wasm-contracts wallet-integration"   \
+          --features "default chain telemetry contract wallet-integration"   \
```
moliholy: No need to specify `--features` at all 😄

crates/pop-cli/src/commands/call/contract.rs
```diff
@@ -191,12 +190,11 @@ impl CallContractCommand {
 		cli.warning("NOTE: contract has not yet been built.")?;
 		let spinner = spinner();
 		spinner.start("Building contract in RELEASE mode...");
-		let result = match build_smart_contract(&project_path, true, Verbosity::Quiet) {
+		let result = match build_smart_contract(&project_path, true, Verbosity::Quiet, None) {
```
moliholy: Is it correct not to pass the metadata spec here?

moliholy: Same here: is it correct not to pass the metadata here?

AlexD10S: `cargo-contract` doesn’t support deploying or calling contracts that use Solidity metadata yet.
In this specific case, where is building the contract only before pop call, I think it makes sense to use the default (ink!) metadata for the build process.

moliholy: Perfect then 👌

.github/workflows/coverage.yml
AlexD10S: I see, you are right actually. Removed https://github.com/r0gue-io/pop-cli/pull/645/commits/036cd90e1f7c6d832381fd16720839c30735dd1c

tsenovilla (approved):
This is great!

Too many changes but mostly related to the flags removal :)

crates/pop-common/src/lib.rs
```diff
@@ -100,6 +101,7 @@ pub fn target() -> Result<&'static str, Error> {
 ///
 /// A new Command instance configured to run the pop binary with the specified arguments
 pub fn pop(dir: &Path, args: impl IntoIterator<Item = impl AsRef<OsStr>>) -> Command {
+	#[allow(deprecated)]
```
moliholy: This seems to be way harder to get rid of, I'm afraid... the equivalent macro does not seem to work. I think we can leave it for later 😓

PR #651, opened 2025-09-30
feat(fork): add fork to pop cli

al3mart:
Adds `fork` to pop cli.

Initially implemented in two different modules: `chain` and `contract` as it might be valuable in the future.
A priori forking a contract might not seem to provide extra value over forking a whole chain as if the chain is host to certain contracts these are brought with the chain' state.

But there might be space for optimizations such that we only pick certain state from certain contracts addresses explicitly provided by the user.

Although the initial path to implementation is forking live chains, I've gone ahead and include the command structure for now as it was not any extra effort.

### DIFF
--- a/crates/pop-cli/src/commands/fork/mod.rs
+++ b/crates/pop-cli/src/commands/fork/mod.rs
@@ -1,0 +1,44 @@
+// SPDX-License-Identifier: GPL-3.0
+
+use clap::{Args, Subcommand};
+use std::fmt::{Display, Formatter, Result};
+use url::Url;
+
+#[cfg(feature = "chain")]
+pub(crate) mod chain;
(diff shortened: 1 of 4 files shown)


crates/pop-cli/src/commands/fork/mod.rs
```diff
@@ -0,0 +1,44 @@
...
+#[cfg(any(feature = "polkavm-contracts", feature = "wasm-contracts"))]
+pub(crate) mod contract;
+
+/// Arguments of the fork command.
+#[derive(Args)]
+#[command(args_conflicts_with_subcommands = true)]
+pub(crate) struct ForkArgs {
+	/// Entrypoint of the fork command.
+	#[command(subcommand)]
+	pub(crate) command: Command,
```
al3mart, line 18: could be an Option and if no subcommand is provided to fork default to `fork chain`

PR #657, opened 2025-10-07
feat: always build contracts with pop up

### LINKED ISSUE #623: Always build the contract before deploying with `pop up`
Currently the contract is only built if no `target` directory is found. I propose to always build it.

Moliholy:
Closes #623.

This PR adds a `skip-build` command to `pop up contract`.

- If `false`, it will always rebuild the contract, even if it was already built. This is the default behaviour.
- If `true`, it won't build the contract in advance, UNLESS it wasn't built already. If so, it'll be built, ignoring the `--skip-build` flag.


AlexD10S (approved):
LGTM!
Just a nitpick, should we add the --skip-build flag in the integration test to avoid rebuilding it? https://github.com/r0gue-io/pop-cli/blob/main/crates/pop-cli/tests/contract.rs#L110

PR #658, opened 2025-10-07
feat(cli): allow to choose a chain from a list

### LINKED ISSUE #616: Offer the user a list of well-known chain RPCs to connect to when using `pop call`
Image

Instead of being the user the one specifying the RPC, Pop could provide a per-relay-chain list of chains with well-maintained URLs to connect to.

This list could be maintained as a JSON file (to be discussed) by the team and be placed in a public repository to enable external contributors to participate.

Moliholy:
Closes #616.

This PR allows users to select a chain to call from a list of well-known RPCs.


Moliholy:
@AlexD10S just added the `supports_contracts` flag to check whether pallet-revive is included in the runtime or not 😎 .

crates/pop-cli/src/common/chain.rs
```diff
@@ -20,14 +24,38 @@ pub(crate) async fn configure(
...
+				let mut prompt = cli.select("Select a chain (type to filter):");
+				for (pos, node) in chains.iter().enumerate() {
+					if filter_fn(node) {
+						prompt = prompt.item(pos, &node.name, "");
+					}
+				}
+				let selected = prompt.filter_mode().interact()?;
+				let providers = &chains[selected].providers;
+				providers
+					.choose(&mut rand::rng())
```
AlexD10S, line 55: We do something similar here for pseudo-random URL rotation: https://github.com/r0gue-io/pop-cli/blob/main/crates/pop-chains/src/deployer_providers.rs#L132
Using that approach could help us drop the rand dependency.

crates/pop-cli/src/commands/call/chain.rs
```diff
@@ -608,6 +613,65 @@ fn parse_function_name(name: &str) -> Result<String, String> {
 	Ok(name.to_ascii_lowercase())
 }
 
+/// Represents a node in the network with its RPC endpoints and chain properties.
+#[allow(dead_code)]
+#[derive(Debug)]
+pub(crate) struct RPCNode {
```
AlexD10S, line 619: Does this logic makes more sense to have it in a common folder? I can see this implemented for contracts at some point

AlexD10S (commented):
Awesome feature!
Thanks for adding the `supportsContracts` flag, that should make it easier to bring this feature to call contracts too.
Are we planning to implement it for `pop call contracts` once we only support `pallet_revive` no? Might be good to open an issue as a reminder.

Left a couple of minor comments only.

crates/pop-cli/src/commands/call/chain.rs
Moliholy: On a second thought, it makes total sense to move it. Done in [e3cd4f3](https://github.com/r0gue-io/pop-cli/pull/658/commits/e3cd4f384c8e219c05a4d40d49261b2de0df02ff).

Moliholy:
@AlexD10S issues created! https://github.com/r0gue-io/pop-cli/issues/661

PR #671, opened 2025-10-14
feat: build the chain with runtime

### LINKED ISSUE #672: Make `pop build` compatible with parachains without node

moliholy:
Closes #672.

This PR:

- Uses the `ChainSpecBuilder` to build node-based parachains, as well as runtime-based parachains that work with `polkadot-omni-node`.
- Allows to specify the `--features` flag for features other than `runtime-benchmarks` and `try-runtime`.
- Refactors the logic to use the new `find_runtime_dir` function throughout the whole project to standarize user interaction.
- Adds/adapts the corresponding tests.

### DIFF
--- a/crates/pop-cli/src/commands/build/mod.rs
+++ b/crates/pop-cli/src/commands/build/mod.rs
@@ -9,7 +9,7 @@ use clap::{Args, Subcommand};
 use contract::BuildContract;
 use duct::cmd;
 use pop_common::Profile;
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 #[cfg(feature = "chain")]
 use {
 	chain::BuildChain,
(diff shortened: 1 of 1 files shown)


crates/pop-cli/src/common/builds.rs
```diff
@@ -55,6 +54,43 @@ pub fn ensure_node_binary_exists(
...
+/// # Returns
+/// The chain spec builder for the node or the runtime.
+#[cfg(feature = "chain")]
+pub fn create_chain_spec_builder(
+	path: &Path,
+	profile: &Profile,
+	default_bootnode: bool,
+	cli: &mut impl Cli,
+) -> anyhow::Result<ChainSpecBuilder> {
+	let default_node_path = path.join("node");
```
AlexD10S: Is this folder always has to be called `node`?

moliholy: Yes, indeed.

PR #673, opened 2025-10-14
feat: upgrade to rust 1.90

### LINKED ISSUE #674: Upgrade pop-cli to the latest Rust version

moliholy:
Closes #674.

This PR upgrades pop-cli to the latest stable rust release -> 1.90.

Most of the changes are related to formatting or clippy issues.

### DIFF
--- a/crates/pop-common/src/sourcing/mod.rs
+++ b/crates/pop-common/src/sourcing/mod.rs
@@ -2,6 +2,7 @@
 
 use crate::{Git, Release, SortedSlice, Status, api, git::GITHUB_API_CLIENT};
 pub use binary::*;
+use derivative::Derivative;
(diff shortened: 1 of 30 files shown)


crates/pop-common/src/sourcing/mod.rs
```diff
@@ -160,7 +161,8 @@ impl Source {
 }
 
 /// A binary sourced from GitHub.
-#[derive(Clone, Debug, PartialEq)]
+#[derive(Clone, Debug, Derivative)]
```
AlexD10S: I was initially confused about why we needed to introduce the new `derivative` dependency, but I see now that without it we’d have to manually implement `PartialEq`. Just confirming that’s the reason for adding it, right?

Issue #663, opened 2025-10-08
feat(pop new): scaffold UI templates


AlexD10S:
Add an optional `--with-frontend` flag to `pop new contract|chain` to scaffold a front-end in the same workspace under a `frontend` folder. 

### Phase 1
Support a curated set of community templates, some examples:

**Contracts**: 
- [inkathon](https://github.com/scio-labs/inkathon)
- [typink](https://github.com/dedotdev/typink)

**Chain:**
- [create-polkadot-dapp](https://github.com/paritytech/create-polkadot-dapp)
- [polkadot-nextjs-starter](https://github.com/niklasp/polkadot-nextjs-starter) 
- [create-dot-app](https://github.com/preschian/create-dot-app)

With the command: 
```sh
pop new contract|chain <name> --with-frontend[=<provider>]  # If provider omitted, open an interactive picker.
```

Pop CLI will pull the frontend templates as we do for chain and contract templates and include it into a `frontend` folder.

#### Run the frontend template
```sh
pop up frontend
```
Pop CLI will detect required tooling. If something’s missing, it prompts to install:
Eg. for inkathon template: https://github.com/scio-labs/inkathon/blob/main/README.md#quickstart It will check Node and bun are installed and if so will run: `cd frontend && bun run dev`

PR #676, opened 2025-10-16
feat: scaffold UI templates

AlexD10S:
Closes https://github.com/r0gue-io/pop-cli/issues/663

New flag `--with-frontend` in `pop new` that supports frontend for contracts and chains. Currently it supports a curated set of community templates:
**Contracts**:  [inkathon](https://github.com/scio-labs/inkathon), [typink](https://github.com/dedotdev/typink)
**Chain:** [create-dot-app](https://github.com/preschian/create-dot-app)

### How to test
```bash
# Install with 
cargo install --git https://github.com/r0gue-io/pop-cli.git --branch  feat/scaffold-ui-templates --locked

# Interactive will prompt you for frontend template
pop new chain / pop new contract
# Or just CLI
pop new chain my-chain --with-frontend
pop new contract flipper --with-frontend
# To specify the template with the CLI
pop new chain my-chain --with-frontend=create-dot-app 
pop new contract flipper --with-frontend=typink
```
It prompts the user to install the needed dependencies if not installed: A version >= 20  of `node` & `bun` for inkathon.
It also also includes `--frontend` in `pop install` to cover the installation of these dependencies:
```
pop install -y --frontend
```

And it provides with a command to easily start the frontend dev server:
```
pop up frontend
```

Ideally, `pop up` would also spin up the local chain (or deploy the contract) and start the frontend dev server in one go. But that's trickier than it looks, because both spinning up the chain/contract and running the frontend are long-running processes that block indefinitely.

### Future improvements:
- inkathon template can show some problems, currently investigating it https://github.com/scio-labs/inkathon/issues/88
- Possibility to include more community templates, e.g [create-polkadot-dapp](https://github.com/paritytech/create-polkadot-dapp) for parachain templates in the future.


crates/pop-cli/src/commands/install/frontend.rs
```diff
@@ -0,0 +1,160 @@
...
+			.confirm("📦 NodeJS v20+ is required. Install/upgrade now via nvm?")
+			.initial_value(true)
+			.interact()?
+		{
+			return Err(anyhow!(
+				"🚫 You have cancelled the installation process. NodeJS v20+ is required. Install from https://nodejs.org and re-run."
+			));
+		}
+	}
+	install_nvm(cli).await?;
```
AlexD10S: https://nodejs.org/en/download

crates/pop-cli/src/commands/install/frontend.rs
```diff
@@ -0,0 +1,160 @@
...
+				"📦 Do you want to proceed with the installation of the following package: bun ?",
+			)
+			.initial_value(true)
+			.interact()?
+		{
+			return Err(anyhow!("🚫 You have cancelled the installation process."));
+		}
+	}
+	// Install Bun (macOS/Linux official script)
+	cmd("bash", vec!["-lc", r#"curl -fsSL https://bun.sh/install | bash"#]).run()?;
```
AlexD10S: https://bun.com/docs/installation

moliholy: There is a `run_external_script` function for this.

al3mart (commented):
Not a very proper review but some comments based on some test runs. 

- When passing `--with-frontend` we are still prompting the user to choose whether they want to generate the frontend. We should skip the question and present the options.

- If there is only one available option, then don't even present options and let's run with it :)

- I had troubles using `inkathon` to generate a contract with a template. Concrete error below but the `frontend` directory was not generated.

```bash
◆  Generation complete
│  
▲  NOTE: the resulting contract is not guaranteed to be audited or reviewed for security vulnerabilities.
│  
bun add v1.2.9 (9a329c04)

installed polkadot-api@1.20.1 with binaries:
 - papi
 - polkadot-api

182 packages installed [5.74s]

Blocked 1 postinstall. Run `bun pm untrusted` for details.
error: Script not found "create-inkathon-app@latest"
◆  Next Steps:
│  ● cd into "./flipper_2" and enjoy hacking! 🚀
│  ● Use `pop build` to build your contract.
│  ● Use `pop up contract` to deploy your contract to a live network.
│  ● Frontend template created inside ./flipper_2. To run it locally, use: `pop up frontend`. Navigate to the `frontend` folder to start customizing it for your contract. 
│  
└  Need help? Learn more at https://learn.onpop.io
```

cc @wottpal Might you know what issue we are hitting here with inkathon ?

crates/pop-cli/src/commands/up/frontend.rs
```diff
@@ -0,0 +1,166 @@
...
+		))?;
+		return Ok(None);
+	}
+	Ok(Some(path))
+}
+
+/// Decide which command to use to run locally the frontend and run it.
+///
+/// # Arguments
+/// * `target` - Location where the smart contract will be created.
```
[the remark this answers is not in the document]
AlexD10S: Good point fixed in https://github.com/r0gue-io/pop-cli/pull/676/commits/1fcf48a252611034b470e20d6c99e66932deb801

crates/pop-cli/src/commands/new/frontend.rs
```diff
@@ -0,0 +1,134 @@
...
+		}
+		prompt = prompt.item(template, template.name(), template.description().trim());
+	}
+	Ok(prompt.interact()?.clone())
+}
+
+/// Create a new frontend.
+///
+/// # Arguments
+/// * `target` - Location where the smart contract will be created.
```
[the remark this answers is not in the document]
AlexD10S: Good catch, fixed https://github.com/r0gue-io/pop-cli/pull/676/commits/1fcf48a252611034b470e20d6c99e66932deb801

crates/pop-cli/src/commands/install/frontend.rs
```diff
@@ -0,0 +1,162 @@
...
+	cli::{self, traits::Confirm},
+	commands::install::run_external_script,
+	common::binary::SemanticVersion,
+};
+use anyhow::{Result, anyhow};
+use duct::cmd;
+use std::path::PathBuf;
+
+const MIN_NODE_VERSION: u8 = 20;
+const NVM_INSTALL_SCRIPT: &str = "https://raw.githubusercontent.com/nvm-sh/nvm/v0.40.3/install.sh";
```
moliholy: I think this should be more sophisticated by always fetching the latest release. For that we should do some digging as we currently do with polkadot-sdk releases. Anyway, not a stopper for this PR to be merged, but rather something for the backlog.

crates/pop-cli/src/commands/install/frontend.rs
```diff
@@ -0,0 +1,162 @@
...
+/// # Arguments
+/// * `skip_confirm`: If true, skip user confirmation prompts.
+/// * `cli`: Command line interface.
+pub async fn ensure_bun(skip_confirm: bool, cli: &mut impl cli::traits::Cli) -> Result<PathBuf> {
+	if let Some(path) = which("bun") {
+		return Ok(PathBuf::from(path));
+	}
+	if !skip_confirm &&
+		!cli.confirm(
+			"📦 Do you want to proceed with the installation of the following package: bun ?",
```
moliholy: Extra space before the question mark.

crates/pop-cli/src/commands/install/frontend.rs
```diff
@@ -0,0 +1,162 @@
...
+	}
+	Ok(())
+}
+
+fn has(bin: &str) -> bool {
+	cmd("which", vec![bin]).read().is_ok()
+}
+
+fn which(bin: &str) -> Option<String> {
+	cmd("which", vec![bin]).read().ok().map(|s| s.trim().to_string())
```
moliholy: I don't think there's a need to trim here. I definitely don't think the `which` command includes unnecessary whitespaces in its output.

crates/pop-cli/src/commands/install/frontend.rs
```diff
@@ -0,0 +1,162 @@
...
+		let result = which("sh");
+		assert!(result.is_some() || result.is_none()); // Just verify no panic
+		let result = which("this_binary_should_not_exist_12345");
+		assert!(result.is_none());
+	}
+
+	#[tokio::test]
+	async fn ensure_bun_returns_path_when_bun_exists() {
+		// Only test if bun exists: check we get a path
+		if which("bun").is_some() {
```
moliholy: This is a conditional test. Tests should not rely on the system having a binary or not.

crates/pop-cli/tests/contract.rs
```diff
@@ -80,9 +80,13 @@ async fn contract_lifecycle() -> Result<()> {
...
 	generate_all_the_templates(temp_dir)?;
 	// pop new contract test_contract (default)
-	let mut command = pop(temp_dir, ["new", "contract", "test_contract", "--template", "standard"]);
+	let mut command = pop(
+		temp_dir,
+		["new", "contract", "test_contract", "--template", "standard", "--with-frontend=typink"],
+	);
 	assert!(command.spawn()?.wait()?.success());
 	assert!(temp_dir.join("test_contract").exists());
+	assert!(temp_dir.join("test_contract").join("frontend").exists());
```
moliholy: We should test `pop up frontend` here too.

@AlexD10S but that does not return immediately, right? If so it's going to be hard to test 🤔

crates/pop-common/src/templates/frontend.rs
```diff
@@ -0,0 +1,165 @@
...
+	)]
+	CreateDotApp,
+}
+
+impl Template for FrontendTemplate {}
+
+impl FrontendTemplate {
+	/// Get the command to create a new project using this template.
+	pub fn command(&self) -> Option<&str> {
+		self.get_str("Command")
```
moliholy: Can't we `unwrap()` here?

moliholy:
Just finished the review! Spotted minor stuff and the vast majority of the logic looks good! Excellent job!

crates/pop-cli/src/commands/new/contract.rs
```diff
@@ -55,7 +70,22 @@ impl NewContractCommand {
 		}
 
 		let template = self.template.clone().unwrap_or_default();
-		let contract_path = generate_contract_from_template(name, path, &template, &mut cli)?;
+		let mut frontend_template: Option<FrontendTemplate> = None;
+		if let Some(frontend_arg) = &self.with_frontend {
+			frontend_template =
+				if frontend_arg == "prompt" {
```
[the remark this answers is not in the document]
moliholy: @AlexD10S rather than using a magic constant, can't we accept `None` as an equivalent to the current `"prompt"`?

crates/pop-cli/src/commands/install/frontend.rs
```diff
@@ -0,0 +1,162 @@
...
+use crate::{
+	cli::{self, traits::Confirm},
+	commands::install::run_external_script,
+	common::binary::SemanticVersion,
+};
+use anyhow::{Result, anyhow};
+use duct::cmd;
+use std::path::PathBuf;
+
+const MIN_NODE_VERSION: u8 = 20;
```
moliholy: I didn't update my node version the first time I run the command and got this warning:

Image

Maybe we should set the minimum to 24?

crates/pop-cli/src/commands/install/frontend.rs
AlexD10S: True, https://github.com/r0gue-io/pop-cli/pull/676/commits/849199d0d98ec36684454b4e1160b582c8497a9e

crates/pop-cli/src/commands/install/frontend.rs
```diff
@@ -0,0 +1,162 @@
...
+	#[test]
+	fn has_works() {
+		assert!(has("sh"));
+		assert!(!has("this_binary_should_not_exist_12345"));
+	}
+
+	#[test]
+	fn which_works() {
+		let result = which("sh");
+		assert!(result.is_some() || result.is_none()); // Just verify no panic
```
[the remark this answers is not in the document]
AlexD10S: You are right is not good, we can check only `is_some()`: https://github.com/r0gue-io/pop-cli/pull/676/commits/849199d0d98ec36684454b4e1160b582c8497a9e

crates/pop-common/src/templates/frontend.rs
AlexD10S: We have been trying to avoid unwrap(), once getting the command in the pop-cli we throw an error if no command has been configured instead:
```
let command = template
		.command()
		.ok_or_else(|| anyhow::anyhow!("no command configured for {:?}", template))?;
```

moliholy: My point is that we can be 100% `.unwrap()` won't fail there. Am I correct? I mean, all variants of the enum have the property `Command` defined, so in principle there is no room for failure.

PR #681, opened 2025-10-17
fix: `pop install` command with rustup

### LINKED ISSUE #683: Improve pop install command
Currently the `pop install` workflow has a few flaws:

- It asks users to confirm rustup's installation.
- Once installed, it is not correctly added to the PATH variable.
- Installation of apt packages in debian systems is not interactive.
- The CI does not check rustup installation, as it is already installed by the time `pop install` is run.

moliholy:
Closes #683.

This PR fixes the installation code by polishing how rustup is installed:

- Making debian packages installation non-interactive by setting the corresponding environment variables.
- Fixing the rustup installation, which was broken in case rustup was not previously installed (which was never checked since in the CI we always installed previously rustup).

I confirm the whole flow has been successfully tested in a docker with `ubuntu:latest` by running the following commands:

```
export NONINTERACTIVE=1 && \
export DEBIAN_FRONTEND=noninteractive && \
apt-get update && \
apt-get install git curl gcc build-essential -y && \
/bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh)" && \
eval "$(/home/linuxbrew/.linuxbrew/bin/brew shellenv)" \
cp /workdir/target/debug/pop /usr/local/bin &&
pop --version && \
unset DEBIAN_FRONTEND && \
pop install -y
```

In normal conditions we'd run `brew install r0gue-io/pop-cli/pop` (which yes, works on linux systems too) instead of copying it, but I already had the pop binary compiled for ubuntu in order to test the installation script with my local code.

### DIFF
--- a/.github/workflows/install.yml
+++ b/.github/workflows/install.yml
@@ -67,17 +67,12 @@ jobs:
           brew update
           brew uninstall --force cmake || true
           brew install cmake openssl protobuf
-      - name: Install Rust
-        run: curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
       - name: Install Pop
         run: |
-          . "$HOME/.cargo/env"
           cargo install --locked --no-default-features --features contract,chain --path ./crates/pop-cli
           cp $(which pop) /usr/local/bin/pop
-          rustup self uninstall -y
+          brew uninstall rustup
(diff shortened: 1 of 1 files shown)


AlexD10S (approved):
LGTM!

Issue #685, opened 2025-10-21
fix: Improve some missing cases with `pop build spec`


tsenovilla:
Follows: 
- #647 

While adding this feature to Paseo CI (https://github.com/paseo-network/runtimes/issues/285) I realized it's not supporting well CI integrations in codebases like Paseo. As it doesn't contain a folder purely called "runtime" anywhere, `pop` will correctly find the runtimes but prompt the user to pick one, which is a pain for CIs.

Additionally, we need to add args to allow modifying the "name" and "id" fields of the spec

PR #687, opened 2025-10-21
fix: Allow `pop build spec` to specify the runtime path as argument

tsenovilla:
Closes:
- #685 

This PR will:
- Allow use `pop build spec` on CIs (eg, Paseo) by letting the user pass the runtime dir path as argument. This wasn't possible before as `find_runtime_dir` automatically prompts the user if there's not a "runtime" folder on the call path (which is the case for Paseo) <- This isn't nice for CIs.
- Allow specifying name and Id in the generated chain spec.
- Get rid of `find_workspace_toml` function as it's not correctly supporting not prefixed inputs like "system-chains/bridge-hub-paseo". I detected this a while ago when I was working on #524 and added a fixed version to [this crate](https://docs.rs/rustilities/latest/rustilities/), which was gonna be added to Pop-CLI on that PR. Let's add it now and also leverage we have it to get rid of `add_crate_to_workspace`, reducing pop codebase's size.

### DIFF
--- a/crates/pop-chains/src/lib.rs
+++ b/crates/pop-chains/src/lib.rs
@@ -20,6 +20,8 @@ mod generator;
 mod new_chain;
 /// Tools for creating new runtime pallets.
 mod new_pallet;
+/// Provides functionality for running runtime-only parachains.
+pub mod omni_node;
 /// A registry of parachains.
 pub mod registry;
 /// Relay chain interaction and management.
--- a/crates/pop-chains/src/omni_node.rs
+++ b/crates/pop-chains/src/omni_node.rs
@@ -1,0 +1,152 @@
+use pop_common::{
+	Error,
+	git::GitHub,
+	polkadot_sdk::sort_by_latest_semantic_version,
+	sourcing::{
+		ArchiveFileSpec, Binary,
+		GitHub::*,
+		Source,
+		filters::prefix,
+		traits::{
+			Source as SourceT,
+			enums::{Source as _, *},
+		},
+	},
+	target,
+};
+use std::path::PathBuf;
+use strum_macros::EnumProperty;
+
+/// CLI enum for managing Polkadot Omni Node binary sources and configuration.
+/// Provides repository information and binary specifications for fetching and managing the node.
+#[derive(Debug, EnumProperty, PartialEq)]
+pub enum PolkadotOmniNodeCli {
+	#[strum(props(
+		Repository = "https://github.com/r0gue-io/polkadot",
+		Binary = "polkadot-omni-node",
+		TagPattern = "polkadot-{version}",
+		Fallback = "stable2506-2"
+	))]
+	/// Polkadot Omni Node binary. Used to bootstrap parachains without node.
+	PolkadotOmniNode,
+}
+
+impl SourceT for PolkadotOmniNodeCli {
+	type Error = Error;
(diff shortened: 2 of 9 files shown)


tsenovilla:
Tried functionality on [Paseo](https://github.com/paseo-network/runtimes) with:
```
pop build spec \
--profile release \
--name "Paseo Bridge Hub Local" \
--id "paseo-bridge-hub-local" \
--para-id 1002 \
--type local \
--relay paseo-local \
--protocol-id bh-pas \
--default-bootnode=false \
--genesis-code=true \
--genesis-state=true \
--deterministic=false \
--properties ss58Format=0,tokenDecimals=10,tokenSymbol=PAS \
--chain local_testnet \
--runtime system-parachains/bridge-hub-paseo \
--output ./chain-specs/local/new_spec.json
```
and compared the generated raw file with [the old one](https://github.com/paseo-network/runtimes/blob/main/chain-specs/local/bridge-hub-paseo-local.json). It looks super good, just one thing caught my attention:
- The added properties are always strings, so for example `"decimals":10` became `"decimals":"10"`. After investigating a bit if this might be problematic I found out that it's OK, so don't need to update that on POP.


We should be able to include this on Paseo workflows soon ;)

moliholy:
Works like a charm!

Many thanks for making this command shine ✨

crates/pop-cli/src/commands/build/spec.rs
```diff
@@ -151,8 +151,8 @@ pub struct BuildSpecCommand {
 	#[arg(long, value_enum)]
 	pub(crate) profile: Option<Profile>,
 	/// Parachain ID to be used when generating the chain spec files.
-	#[arg(short, long)]
-	pub(crate) id: Option<u32>,
+	#[arg(long = "para-id")]
+	pub(crate) para_id: Option<u32>,
```
AlexD10S: This is a breaking change, any scripts that generate the build spec using `id` will fail. 
Does it make sense to rename the field below `id` to `network_id `instead? Otherwise we’ll need to make it clear in the next release.

tsenovilla: Good catch

moliholy: @AlexD10S there are two ids in the JSON:

- `id` (string): the chain readable id, like "dev", "polkadot" or alike.
- `parachainId` (number): the parachain's numeric ID.

IMO, pop's `id` should have never been named as such, as it conflicts with the JSON's `id` parameter. After changes in this PR, now both parameters are aligned with the JSON, so I'd say the breaking change is justified.

tsenovilla: Definitely everything's opinionated :)

I can't agree more with this:
> when users compare the parameters they are passing to the client to the resulting JSON they are going to get inevitably confused, which is not user-friendly

Hence my vote is to stay with the breaking change

AlexD10S: Cool then let's use `para_id` then and include it in the release notes as a breaking change.

One more nitpick here: we use the `short` in all fields consistently around pop, I see you removed it from `id` https://github.com/r0gue-io/pop-cli/pull/687#discussion_r2452049052 because it doesn't really make sense  and I agree. But let's have add a short option here (`-i`?).

AlexD10S (approved):

PR #699, opened 2025-10-29
feat: pop up chain spec cmd

moliholy:
This PR:

- Updates zombienet to 0.4.1.
- Adds the `--raw` flag to `pop build spec` to build the raw chain spec.
- Switches the directory when executing a zombienet network. This allows to run commands from the parachain root folder that will be used to generate the chainspec.

All these changes are related to adding support for omni-node-based chains.


Daanvdplas (approved):
LGTM!

PR #711, opened 2025-11-04
feat: remove network configuration from base-parachain

moliholy:
As discussed in https://github.com/r0gue-io/base-parachain/pull/69#pullrequestreview-3416035048, this PR removes the `network.toml` file from the base-parachain template.

### DIFF
--- a/crates/pop-contracts/src/node.rs
+++ b/crates/pop-contracts/src/node.rs
@@ -142,6 +142,7 @@ pub async fn run_ink_node(
 	let mut command = Command::new(binary_path);
 	command.arg("-linfo,runtime::contracts=debug");
 	command.arg(format!("--rpc-port={}", port));
+	command.arg("--tmp");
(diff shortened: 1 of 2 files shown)


crates/pop-chains/src/new_chain.rs
```diff
@@ -73,9 +73,6 @@ pub fn instantiate_standard_template(
 		&target.join("node/src/chain_spec.rs"),
 		chainspec.render().expect("infallible").as_ref(),
 	)?;
-	// Add network configuration
```
AlexD10S: For devs using the templates in a version 2506 or below this, this is a breaking change that it won't create a network.toml
Because we have the version in https://github.com/r0gue-io/pop-cli/blob/main/crates/pop-chains/src/new_chain.rs#L46 we can check the version and generate the network.toml if is an older version

PR #712, opened 2025-11-04
fix: chain & contract manual interaction

Daanvdplas:
Closes #710

PR #713, opened 2025-11-04
fix: actually use already passed contract message arguments

moliholy:
This PR fixes an issue by which passed arguments when calling a contract were omitted.


Daanvdplas:
https://github.com/r0gue-io/pop-cli/pull/712

moliholy:
Closes in favor of #712.

PR #714, opened 2025-11-04
fix: contract gas input devex

### LINKED ISSUE #612: Improve weight messaging in `pop call` command
Image

Few things here:

1) In the `Enter the gas limit` question:
    - It is not "gas", it's the `ref_time` component of the weight, i.e. the maximum number of picoseconds of execution for a given extrinsic. The name is misleading, and does not have a 1:1 parallelism with the typical Ethereum "gas" devs are used to. I propose renaming if it to something to "execution time" or something alike. Of course, open to discussion.
    - I wonder... how often users do NOT estimate the call? If users do always estimate, I think it'd make sense to only display this question if the user provides a flag for it, or even in the most extreme case: not specifying it for simplicity.

2) In the `Enter the proof size limit` question:
    - If the previous question was left blank... will users really fill this one? I think it should be skipped in that case.
    - Same point applies here: it should only be specified if a flag is passed and automatically estimated.

Daanvdplas:
Closes #612

### DIFF
--- a/crates/pop-cli/src/commands/call/contract.rs
+++ b/crates/pop-cli/src/commands/call/contract.rs
@@ -58,7 +58,7 @@ pub struct CallContractCommand {
 	/// If not specified it will perform a dry-run to estimate the proof size required.
 	/// DEPRECATED use `--manual-weight <REF_TIME> <PROOF_SIZE>` instead, will be removed in
 	/// v0.13.0.
-	#[arg(short = 'P', long, conflicts_with = "manual_weight", requires = "gas_limit")]
+	#[arg(short = 'P', long, conflicts_with = "manual_weight", requires = "gas")]
 	proof_size: Option<u64>,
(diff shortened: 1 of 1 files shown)


crates/pop-cli/src/commands/call/contract.rs
```diff
@@ -51,12 +51,24 @@ pub struct CallContractCommand {
 	/// Maximum amount of gas to be used for this command.
 	/// If not specified it will perform a dry-run to estimate the gas consumed for the
 	/// call.
-	#[arg(name = "gas", short, long)]
+	/// DEPRECATED use `--manual-weight <REF_TIME> <PROOF_SIZE>` instead, removed in v0.13.0.
+	#[arg(name = "gas", short, long, conflicts_with = "manual_weight", requires = "proof_size")]
```
moliholy: This one (and the other deprecated ones) should be marked with `#[deprecated]`.

PR #716, opened 2025-11-05
chore: Bump rustilities

tsenovilla:
Just a PR to use a newer version of rustilities that doesn't necessarily depends on rust 1.90


AlexD10S (approved):
Thanks for this quick fix!
For context reported in https://github.com/r0gue-io/pop-docs/issues/59

PR #718, opened 2025-11-05
chore: update passet-hub spec

al3mart:
Updates PAsset Hub spec to a more recent version.


AlexD10S (approved):
Thanks for the quick update, I have tested and works well

PR #723, opened 2025-11-06
feat: rich info for telemetry

moliholy:
This PR allows telemetry to have very complete data of all commands being used by serializing as JSON the commands being run, even at the end of the workflow when they have already been populated in case the user did not specify all possible parameters.

Sensitive data has been excluded from this serialization.

## Screenshot

Statistics will look like this from now on:

Screenshot 2025-11-06 at 18 23 27


Daanvdplas:
Not sure if you have the time but if it is easy, can you look at #635 for additional things?

PR #725, opened 2025-11-10
fix(pop up contract): signer prompted if not provided

### LINKED ISSUE #721: `pop up` doesn't ask for who or how to sign the transaction

### LINKED ISSUE #732: pop up flow inconsistent
If I deploy the ink-node and the contract using `pop up`, and I provide the --suri flag with //Alice or I use the 
--use-wallet flag, am I then asked to continue interacting with the contract for both scenarios? 
  
Scenario 1: Using --suri //Alice

  YES - You ARE prompted to continue interacting with the contract.

  The flow follows this path in /Users/R0GUE/Documents/Code/R0GUE/pop-cli/crates/pop-cli/src/commands/up/contract.rs:
  1. Line 246: resolve_signer() is called (no prompt needed since suri is already provided)
  2. Line 249: Skips the wallet flow since use_wallet is false
  3. Lines 395-407: After successful deployment, it calls:
  Cli.success(COMPLETE)?;
  self.keep_interacting_with_node(&mut Cli, contract_address).await?;
  4. Lines 410-428: The keep_interacting_with_node method prompts:
  "Do you want to keep making calls to the contract?"

  Scenario 2: Using --use-wallet

  NO - You are NOT prompted to continue interacting with the contract.

  The flow follows this path:
  1. Line 246: resolve_signer() is called (no prompt needed since use_wallet is true)
  2. Line 249-338: Enters the wallet flow
  3. Line 336-338: After successful deployment:
  Cli.outro(COMPLETE)?;
  terminate_nodes(&mut Cli, processes).await?;
  return Ok(());  // Returns early!
  4. The method returns without calling keep_interacting_with_node

This should be fixed

Daanvdplas:
When deploying a contract the signer was no prompted if not provided and alice was always used. This fixes that.

Closes #721 
Closes #732


crates/pop-cli/src/common/contracts.rs
```diff
@@ -186,6 +186,39 @@ pub(crate) async fn map_account(
 	Ok(())
 }
 
+/// Resolves the signer for contract operations (deployment or calls).
```
moliholy: I don't like this function as it is right now. Reasons:

- Takes assumptions on parameters passed in a given workflow, yet the function itself is supposed to be agnostic.
- Mutates received parameters. That might have been an option in C's old days when you could not return tuples or enums, but I definitely wouldn't do that with Rust's capabilities.

My proposal is to make this function internal inside `UpContractCommand` and take a mutable reference to itself, and only them mutate its state, instead of delegating that functionality to a foreign function.

Daanvdplas:
Can we also check whether with the --skip-confirm flag the user is not prompted for confirmation on executing the message @moliholy

PR #735, opened 2025-11-12
feat: rust 1.91.1

moliholy:
This PR:

- Bumps rust version to 1.91.1.
- Removes the need to specify different versions of pop components. This makes easier to bump pop versions, as less changes are required.
- Includes the `telemetry` feature in the `install` workflow.
- Adds the `--locked` flag in all CI executions/builds.


al3mart (approved):
Makes sense to me 👍

PR #750, opened 2025-11-14
fix: allow mcp to fetch the contract address upon deployment

### LINKED ISSUE #747: Fix pop up output for mcp
We deploy contracts via pop up <contract …>, and the CLI only prints the contract address inside display_contract_info, which stops a cliclack spinner (`crates/pop-cli/src/commands/up/contract.rs:378-386,544-562`). That spinner uses indicatif, and when the command runs without a TTY (the MCP uses Command::output, so no tty) indicatif suppresses all spinner rendering—meaning the contract address never reaches stdout or stderr. Other commands like pop up ink-node log their status through Cli.info/Cli.outro, which emit even without a TTY, so the MCP sees those. To fix the deployment case we need to output the contract address via a channel that works in non‑TTY contexts (e.g., Cli.info, println!, or a structured stdout line) right after display_contract_info in both the normal and wallet flows; otherwise automation will never be able to capture the address.

moliholy:
Closes #747 

This PR allows the mcp to easily fetch the generated contract address upon deployment.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -72,8 +72,8 @@ symlink = { version = "0.1", default-features = false }
 serde_json = { version = "1.0", default-features = false, features = ["preserve_order"] }
 serde = { version = "1.0", default-features = false, features = ["derive"] }
 srtool-lib = { version = "0.13.2", default-features = false }
-zombienet-configuration = { version = "0.4.2", default-features = false }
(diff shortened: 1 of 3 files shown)


AlexD10S (approved):
Thanks for this fix! LGTM

PR #752, opened 2025-11-14
fix: remove --manual-weight flag

### LINKED ISSUE #738: pop up gas / proof size arguments inconsistent with pop call
https://github.com/r0gue-io/pop-cli/pull/714

The manual-weight flag should also be implemented for pop up.

moliholy:
Closes #738

This PR removes the `--manual-weight` flag, which is inconsistent with the `pop up` flow.


AlexD10S (approved):
Thanks for removing this!

PR #753, opened 2025-11-14
fix(contracts): tx always wait for finalization

al3mart:
This way we return the events as soon as the tx is in a block. Improves devex highly


AlexD10S (approved):
Great! LGTM

PR #759, opened 2025-11-14
fix: ask to execute deployment of contract in interactive flow

### LINKED ISSUE #758: `pop up` flow small bug
When I run `pop up` in my contract it doesn't ask me to execute the call and thus will not be deployed and return the contract address, so it looks I deployed the contract but the address not returned.

Daanvdplas:
Closes #758


AlexD10S (approved):
LGTM

PR #778, opened 2025-11-18
fix: parachain_lifecycle integration test

moliholy:
This PR fixes the broken `parachain_lifecycle` integration test.

This error started to occur right after we released the newest version of the parachain template that does not include the runtime, as it was expecting a node.


AlexD10S (changes requested):
The test is failing when testing the `pop call chain` because the chain doesn't seems to be running after the `pop up`.
The version we are launching is hardcoded in the version `stable2412` which doesn't have the changes you applied: https://github.com/r0gue-io/pop-cli/blob/main/crates/pop-cli/tests/chain.rs#L195 removing it should fix it, or if we want to be sure is the one that uses the omninode changes use the `2506.1`

AlexD10S (approved):
Amazing! Thanks for fixing this

Issue #773, opened 2025-11-17
feat: Stop using chain spec generator for Paseo


tsenovilla:
Now that Paseo's releases include local chain specs (eg, https://github.com/paseo-network/runtimes/releases/tag/v2.0.1) and the chain spec generator was removed from Paseo, we can start fetching the files directly from Paseo's releases instead of fetching the chain spec generator binary

PR #783, opened 2025-11-19
feat: Allow sourcing plain files (not just binaries) and remove Paseo chain specs generator in favor of chain specs

tsenovilla:
Closes:
- #773 

By:

- Allowing source plain files, not just binaries. This extends the cache concept of `pop` and might be used by other features in the future.
- Using the released chain specs for Paseo instead of the former chain spec generator. This implies that in the future we can get rid of this repo as well: https://github.com/r0gue-io/paseo-runtimes (when we stop giving support to some old `pop` versions).


crates/pop-common/src/sourcing/archive.rs
```diff
@@ -61,40 +77,66 @@ impl Binary {
...
-			Self::Source { name, cache, .. } => {
+			Self::Source { name, cache, archive_type, .. } => {
 				// Determine whether a specific version is specified
-				self.version()
-					.map_or_else(|| cache.join(name), |v| cache.join(format!("{name}-{v}")))
+				self.version().map_or_else(
+					|| cache.join(name),
+					|v| {
+						// For files,
+						if *archive_type == ArchiveType::File {
```
moliholy: I'd use match expressions in cases like this one.

PR #786, opened 2025-11-20
fix: do not delete ink-node logs if detached

moliholy:
This PR fixes a small bug by which ink-node and eth-rpc log files were pruned when the process finished.


AlexD10S (approved):
LGTM!

Issue #793, opened 2025-11-25
feat: Contract verification


tsenovilla:
Tracking issue for contract verification feature. 

This feature basically consist of three steps:

1. `pop build --verifiable` <- allowing users to build their contracts in a deterministic why such that it can be reproduced. `cargo contract` already includes this feature, so we should adapt it.
2. `pop publish` <- allowing users to publish all stuff needed for verification (source code, metadata, contract address, ...) to some place where everyone can pick it and verify its authenticity (to be discussed how to do it because we have a couple of options).
3. `pop verify` <- allowing users to verify a contract, either published or locally. Again, `cargo contract` already includes it so we need to adapt it for `pop`.

Daanvdplas:
A little late to the party, but should we change `pop build --verifiable` to `pop build --deterministic`? This makes it consistent with the building a runtime deterministically

Daanvdplas:
What do you mean with `we should be consistent not with the runtime commands but with the contracts commands.`? We are defining the contract commands now, why not be consistent?

The functionality is the same as far as I understand, building a binary in a deterministic way, why have two different names for the flags?

PR #797, opened 2025-11-25
feat: Contract verification

tsenovilla:
Closes:

- #793

### DIFF
--- a/crates/pop-cli/src/commands/verify.rs
+++ b/crates/pop-cli/src/commands/verify.rs
@@ -43,7 +43,7 @@ pub(crate) struct VerifyCommand {
 
 impl VerifyCommand {
 	pub(crate) async fn execute(&self, cli: &mut impl Cli) -> Result<()> {
-		cli.intro("Start verifying your contract, this may take a bit ⏳")?;
+		cli.intro("Contract verification started. This might take a bit⏳")?;
 
 		let project_path = ensure_project_path(self.path.clone(), self.path_pos.clone());
 
--- a/crates/pop-cli/tests/contract.rs
+++ b/crates/pop-cli/tests/contract.rs
@@ -299,7 +299,7 @@ async fn verifiable_contract_lifecycle() -> Result<()> {
 
 	let temp = tempfile::tempdir()?;
 	let temp_dir = temp.path();
-	//let temp_dir = Path::new("./"); //For testing locally
+
 	// pop new contract test_contract (default)
 	let mut command = pop(temp_dir, ["new", "contract", "test_contract", "--template", "standard"]);
 	assert!(command.spawn()?.wait().await?.success());


crates/pop-cli/src/commands/build/contract.rs
```diff
@@ -1,17 +1,21 @@
...
 
 /// Configuration for building a smart contract.
 pub struct BuildContract {
 	/// Path of the contract project.
 	pub(crate) path: PathBuf,
-	/// Build profile: `true` for release mode, `false` for debug mode.
-	pub(crate) release: bool,
+	/// Build profile: `Release` for release mode, `Debug` for debug mode, `Verifiable` for
+	/// deterministic, release mode.
+	pub(crate) build_mode: BuildMode,
```
Daanvdplas: Could you check that this is also possible in this way for chain building?

crates/pop-cli/src/commands/build/mod.rs
```diff
@@ -46,8 +50,8 @@ pub(crate) struct BuildArgs {
 	#[cfg(feature = "chain")]
 	pub command: Option<Command>,
 	/// Directory path with flag for your project [default: current directory]
+	#[arg(long, visible_alias = "manifest-path")]
```
Daanvdplas: The code already resolves a manifest from either a directory or a Cargo.toml so can we not drop the special-casing and let --path work in verifiable mode too?

crates/pop-cli/src/commands/build/mod.rs
```diff
@@ -89,6 +93,15 @@ pub(crate) struct BuildArgs {
...
 	pub(crate) metadata: Option<MetadataSpec>,
+	/// Whether to build in a way that the contract is verifiable. For verifiable contracts, use
+	/// --manifest-path instead of --path to directly point to your contracts Cargo.toml
+	#[clap(long, conflicts_with_all = ["release", "profile"], help_heading = CONTRACT_HELP_HEADER)]
+	#[cfg(feature = "contract")]
+	pub(crate) verifiable: bool,
+	/// Custom image for verifiable builds
+	#[clap(long, requires = "verifiable", help_heading = CONTRACT_HELP_HEADER)]
+	#[cfg(feature = "contract")]
+	pub(crate) image: Option<String>,
```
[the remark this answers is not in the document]
tsenovilla: Not really. `image` is used to specify which docker image we should use (the image name), `tag` for specifying the version of the image used. So `pop build --deterministic` can specify the tag but not the image (always using srtool). Here you can create your own docker image and use it (and then verification will be run against that image), defaulting to `use-ink` one if not specified

crates/pop-cli/src/commands/build/mod.rs
tsenovilla: Already discussed here: https://github.com/r0gue-io/pop-cli/pull/800#discussion_r2576317611

Basically if we use the default `useink` image (which we should, as otherwise we enforce users to create their own images), we need to pass `manifest-path` there. However https://github.com/r0gue-io/pop-cli/pull/812 uses a new version of cargo contract containing a new generic implementation (https://github.com/use-ink/cargo-contract/pull/2177) for what's passed to the image, so we can update this now! 

Thanks for bringing it back cause this part of the PR was discussed before the `cargo-contract` one even existed, and I totally forgot about it :) I'll update this branch when the `pop verify` PR is merged

crates/pop-cli/tests/contract.rs
```diff
@@ -291,6 +292,116 @@ async fn contract_lifecycle() -> Result<()> {
...
 }
 
+#[tokio::test]
+async fn verifiable_contract_lifecycle() -> Result<()> {
+	let endpoint_port = resolve_port(None);
+	let default_endpoint: &str = &format!("ws://127.0.0.1:{}", endpoint_port);
+
+	let temp = tempfile::tempdir()?;
+	let temp_dir = temp.path();
+	//let temp_dir = Path::new("./"); //For testing locally
```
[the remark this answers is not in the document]
tsenovilla: https://github.com/r0gue-io/pop-cli/pull/797/commits/76aa6d29514a39bcc6ce0c948df4dd3028086717

crates/pop-cli/src/commands/verify.rs
```diff
@@ -0,0 +1,84 @@
...
+		let success_message = if let (Some(endpoint), Some(address)) = (&self.url, &self.address) {
+			format!(
+				"The contract deployed on {} at address {} is successfully verified ✅",
+				endpoint, address
+			)
+		} else {
+			"The contract is successfully verified ✅".to_string()
+		};
+
+		let _ = cli.success(success_message);
```
moliholy: This should throw an error if fails (which never does).

Also, this is the end of the command, so probably it should be `outro`.

moliholy (approved):
Amazing job 🚀

crates/pop-cli/src/commands/verify.rs
moliholy: No need. Anyway this won't be usable 😢

PR #813, opened 2025-12-03
refactor: remove the code to receive the contract address from wallet signing portal

AlexD10S:
We previously added a patch to manually calculate the smart contract address in the **wallet signing portal** when the `Instantiated` event didn't exist. Receiving the address in `pop-cli`.
The event was reintroduced in https://github.com/paritytech/polkadot-sdk/pull/8789
Since the address can now be obtained directly from the events in `pop-cli`, the manual calculation is no longer necessary in the wallet signing portal and can be removed.

Is basically removing undoing this commit where this was introduced: https://github.com/r0gue-io/pop-cli/commit/3b7e05967910dd0075688c9aa78e5b071672d1d5

The changes in the **wallet signing portal** that needs to be reviewed too are in: https://github.com/r0gue-io/pop-wallet-signing-portal/pull/12


AlexD10S:
> @AlexD10S if using a polkadot-sdk version older than 2509 pallet-revive won't have that commit, isn't it?

How did you see the 2509? PassetHub I believe has the 2507 and it has the `Instantiated` event

moliholy (approved):
Then let's go ahead!

PR #833, opened 2025-12-08
feat(fork): RPC Client Wrapper

al3mart:
Addresses #819 

## Description

Create `ForkRpcClient` - a wrapper that provides the RPC methods needed for lazy-loading storage from a live chain. This is the foundational component that enables connecting to any Polkadot SDK chain and fetching data on-demand.

### API Implemented

```rust
pub struct ForkRpcClient {
    client: RpcClient,
    legacy: LegacyRpcMethods<PolkadotConfig>,
    endpoint: Url,
}

impl ForkRpcClient {
    /// Connect to a live chain
    pub async fn connect(endpoint: &Url) -> Result<Self, RpcClientError>;

    /// Get the endpoint URL
    pub fn endpoint(&self) -> &Url;

    /// Get the latest finalized block hash
    pub async fn finalized_head(&self) -> Result<H256, RpcClientError>;

    /// Get block header by hash
    pub async fn header(&self, hash: H256) -> Result<Header, RpcClientError>;

    /// Get single storage value at block
    pub async fn storage(&self, key: &[u8], at: H256) -> Result<Option<Vec<u8>>, RpcClientError>;

    /// Get multiple storage values (batch)
    pub async fn storage_batch(&self, keys: &[Vec<u8>], at: H256) -> Result<Vec<Option<Vec<u8>>>, RpcClientError>;

    /// Get storage keys by prefix (paginated)
    pub async fn storage_keys_paged(
        &self,
        prefix: &[u8],
        count: u32,
        start_key: Option<&[u8]>,
        at: H256
    ) -> Result<Vec<Vec<u8>>, RpcClientError>;

    /// Get runtime metadata
    pub async fn metadata(&self, at: H256) -> Result<Vec<u8>, RpcClientError>;

    /// Get runtime code (:code storage key)
    pub async fn runtime_code(&self, at: H256) -> Result<Vec<u8>, RpcClientError>;

    /// Get chain name (for CLI display)
    pub async fn system_chain(&self) -> Result<String, RpcClientError>;

    /// Get system properties (token decimals, symbols, etc.)
    pub async fn system_properties(&self) -> Result<SystemProperties, RpcClientError>;
}
```

### RPC Methods Used

| Method | Purpose |
|--------|---------|
| `chain_getFinalizedHead` | Get latest finalized block hash |
| `chain_getHeader` | Get block header |
| `state_getStorage` | Fetch single storage value |
| `state_queryStorageAt` | Batch fetch storage values |
| `state_getKeysPaged` | Enumerate storage keys by prefix |
| `state_getMetadata` | Get runtime metadata |
| `system_chain` | Get chain name |
| `system_properties` | Get token decimals, symbols, etc. |

## Tasks

- [x] Create `crates/pop-fork/` crate with `Cargo.toml`
- [x] Create `crates/pop-fork/src/lib.rs` and `crates/pop-fork/src/rpc.rs`
- [x] Implement `ForkRpcClient::connect()`
- [x] Implement `finalized_head()`
- [x] Implement `header()`
- [x] Implement `storage()` for single key fetch
- [x] Implement `storage_batch()` for batch fetching
- [x] Implement `storage_keys_paged()` for key enumeration
- [x] Implement `metadata()`
- [x] Implement `runtime_code()` helper
- [x] Implement `system_chain()` (added: useful for CLI output)
- [x] Implement `system_properties()` (added: useful for displaying token info)
- [x] Add integration tests connecting to Paseo testnet

---

# Why Legacy RPCs?

We use subxt's `LegacyRpcMethods` (`state_*`, `chain_*`) rather than the newer
`chainHead_v1_*` or `archive_v1_*` specifications because:

1. **Universal support**: Legacy RPCs work with all Polkadot SDK nodes. The newer specs may not
be available on all endpoints.

2. **Simplicity**: Legacy RPCs use request/response patterns. The new specs require subscription
    lifecycle management (follow/unfollow, pin/unpin) which adds complexity for our use case of
    querying a specific historical block.

 3. **Precedent**: Tools like [chopsticks](https://github.com/AcalaNetwork/chopsticks) use legacy RPCs for fetching from upstream chains.

Note: subxt marks legacy methods as "not advised" but they remain widely used. This decision should be revisited if the ecosystem moves away from legacy RPCs.

### DIFF
--- a/crates/pop-fork/src/rpc.rs
+++ b/crates/pop-fork/src/rpc.rs
@@ -59,42 +59,16 @@
 //! cargo nextest run -p pop-fork --features integration-tests --test rpc -j 1
 //! ```
 
-use crate::strings::rpc::{methods, storage_keys};
+use crate::{
+	error::rpc::RpcClientError,
+	strings::rpc::{methods, storage_keys},
+};
 use subxt::{
 	SubstrateConfig,
 	backend::{legacy::LegacyRpcMethods, rpc::RpcClient},
 	config::substrate::H256,
 };
-use thiserror::Error;
 use url::Url;
-
-/// Errors that can occur when interacting with the RPC client.
-#[derive(Debug, Error)]
-pub enum RpcClientError {
-	/// Failed to connect to the RPC endpoint.
-	#[error("Failed to connect to {endpoint}: {message}")]
-	ConnectionFailed { endpoint: String, message: String },
-	/// RPC request failed.
-	#[error("RPC request `{method}` failed: {message}")]
-	RequestFailed {
-		/// The RPC method that failed.
-		method: &'static str,
-		/// The error message.
-		message: String,
-	},
-	/// RPC request timed out.
-	#[error("RPC request `{method}` timed out")]
-	Timeout {
-		/// The RPC method that timed out.
-		method: &'static str,
-	},
-	/// Invalid response from RPC.
-	#[error("Invalid RPC response: {0}")]
-	InvalidResponse(String),
-	/// Storage key not found (this is different from empty storage).
-	#[error("Required storage key not found: {0}")]
-	StorageNotFound(String),
-}
 
 /// RPC client wrapper for fork operations.
 ///
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -95,6 +95,20 @@ jobs:
           - macos-latest
     runs-on: ${{ matrix.os }}
     steps:
+      - name: Free up space on runner (Linux)
+        if: matrix.os == 'ubuntu-latest'
+        shell: bash
+        run: |
+          set -xeuo pipefail
+          sudo rm -rf /usr/local/lib/android &
+          sudo rm -rf /usr/share/dotnet &
+          sudo rm -rf /usr/share/swift &
+          if command -v docker &> /dev/null; then
+            sudo docker image prune -af &
+          fi
+          sudo apt-get clean &
+          wait
+
       - uses: actions/checkout@v4
 
       - name: Install stable toolchain
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,7 @@ debian/debhelper-build-stamp
 debian/files
 debian/pop-cli.substvars
 debian/.debhelper
+
+
+# Diesel
+.diesel_lock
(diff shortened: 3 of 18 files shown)


crates/pop-fork/src/rpc.rs
```diff
@@ -0,0 +1,326 @@
...
+				change_set.changes.into_iter().map(|(k, v)| {
+					let key_bytes = k.0.to_vec();
+					let value_bytes = v.map(|v| v.0.to_vec());
+					(key_bytes, value_bytes)
+				})
+			})
+			.collect();
+
+		// Return values in the same order as input keys
+		let values = keys.iter().map(|key| changes.get(key).cloned().flatten()).collect();
```
tsenovilla: values can be arbitrarily big, these clones may be more expensive than what we can imagine. I'd just remove the ownership from the map and use always the same memory blob.

```suggestion
		let values = keys.iter().map(|key| changes.remove(key)).collect();
```

crates/pop-fork/src/rpc.rs
```diff
@@ -0,0 +1,326 @@
...
+//! This separation exists because:
+//!
+//! 1. Tests spawn local test nodes (ink-node), requiring binary downloads and process management.
+//! 2. Concurrent test execution can cause race conditions during node binary downloads.
+//! 3. CI runs these tests with `-j 1` (sequential) to avoid resource contention.
+//!
+//! To run the tests locally:
+//! ```bash
+//! cargo nextest run -p pop-fork --features integration-tests --test rpc -j 1
+//! ```
```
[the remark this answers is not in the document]
al3mart: Resolving as per https://github.com/r0gue-io/pop-cli/pull/833#discussion_r2619799941.

crates/pop-fork/src/rpc.rs
```diff
@@ -0,0 +1,346 @@
...
+	///
+	/// # Arguments
+	/// * `keys` - Slice of storage keys to fetch
+	/// * `at` - The block hash to query state at
+	///
+	/// # Returns
+	/// A vector of optional values, in the same order as the input keys.
+	pub async fn storage_batch(
+		&self,
+		keys: &[Vec<u8>],
```
al3mart: I've changed the type of keys in https://github.com/r0gue-io/pop-cli/pull/840/changes#diff-efcd9fbee39f89e8765e88df7710ed298e46fed96a31c31e0672d5df66a95c2fR187

PR#840

Issue #820, opened 2025-12-08
1.2 fork: Storage Cache (SQLite)


al3mart:
# 1.2 Storage Cache (SQLite)

## Summary

Implement a persistent SQLite-based cache for storing fetched storage values, enabling fast restarts and reducing RPC calls to live chains.

## Branch

_note that my user al3mart may be used in the branch names for the shake of having a full story. Please do check if the branches below are the correct ones_

Base: `feat/pop-fork`

## Location

`crates/pop-common/src/fork/cache.rs`

## Description

Create `StorageCache` - a SQLite-backed persistent cache for storage values fetched from live chains. This is critical for good UX since it allows:
- Fast restarts without re-fetching all data
- Reduced load on public RPC endpoints
- Offline access to previously fetched data

Uses `sqlx` for async-native database access, which integrates cleanly with the tokio-based async architecture of the fork infrastructure.

### API to Implement

```rust
use sqlx::{SqlitePool, sqlite::SqlitePoolOptions};

pub struct StorageCache {
    pool: SqlitePool,
}

impl StorageCache {
    /// Open or create cache database at path
    pub async fn open(path: &Path) -> Result<Self>;

    /// Open in-memory cache (for testing)
    pub async fn in_memory() -> Result<Self>;

    /// Get cached storage value
    /// Returns:
    ///   Some(Some(value)) = cached with value
    ///   Some(None) = cached as empty (storage key exists but has no value)
    ///   None = not in cache (unknown)
    pub async fn get(&self, block_hash: H256, key: &[u8]) -> Result<Option<Option<Vec<u8>>>>;

    /// Cache a storage value
    pub async fn set(&self, block_hash: H256, key: &[u8], value: Option<&[u8]>) -> Result<()>;

    /// Batch get (for efficiency)
    pub async fn get_batch(&self, block_hash: H256, keys: &[Vec<u8>])
        -> Result<Vec<Option<Option<Vec<u8>>>>>;

    /// Batch set (for efficiency)
    pub async fn set_batch(&self, block_hash: H256, entries: &[(Vec<u8>, Option<Vec<u8>>)])
        -> Result<()>;

    /// Cache block metadata
    pub async fn cache_block(&self, hash: H256, number: u32, header: &[u8]) -> Result<()>;

    /// Get cached block
    pub async fn get_block(&self, hash: H256) -> Result<Option<BlockInfo>>;
}

pub struct BlockInfo {
    pub hash: H256,
    pub number: u32,
[...]

PR #834, opened 2025-12-08
feat(fork): Storage Cache (SQLite)

al3mart:
Addresses #820 

## Description

Create `StorageCache` - a SQLite-backed persistent cache for storage values fetched from live chains. This is critical for good UX since it allows:
- Fast restarts without re-fetching all data
- Reduced load on public RPC endpoints
- Offline access to previously fetched data

Uses `sqlx` for async-native database access, which integrates cleanly with the tokio-based async architecture of the fork infrastructure.

### API to Implement

```rust
use sqlx::{SqlitePool, sqlite::SqlitePoolOptions};

pub struct StorageCache {
    pool: SqlitePool,
}

impl StorageCache {
    /// Open or create cache database at path
    pub async fn open(path: &Path) -> Result<Self>;

    /// Open in-memory cache (for testing)
    pub async fn in_memory() -> Result<Self>;

    /// Get cached storage value
    /// Returns:
    ///   Some(Some(value)) = cached with value
    ///   Some(None) = cached as empty (storage key exists but has no value)
    ///   None = not in cache (unknown)
    pub async fn get(&self, block_hash: H256, key: &[u8]) -> Result<Option<Option<Vec<u8>>>>;

    /// Cache a storage value
    pub async fn set(&self, block_hash: H256, key: &[u8], value: Option<&[u8]>) -> Result<()>;

    /// Batch get (for efficiency)
    pub async fn get_batch(&self, block_hash: H256, keys: &[Vec<u8>])
        -> Result<Vec<Option<Option<Vec<u8>>>>>;

    /// Batch set (for efficiency)
    pub async fn set_batch(&self, block_hash: H256, entries: &[(Vec<u8>, Option<Vec<u8>>)])
        -> Result<()>;

    /// Cache block metadata
    pub async fn cache_block(&self, hash: H256, number: u32, header: &[u8]) -> Result<()>;

    /// Get cached block
    pub async fn get_block(&self, hash: H256) -> Result<Option<BlockInfo>>;
}

pub struct BlockInfo {
    pub hash: H256,
    pub number: u32,
    pub header: Vec<u8>,
    pub parent_hash: H256,
}
```

### Database Schema

```sql
CREATE TABLE IF NOT EXISTS storage (
    block_hash BLOB NOT NULL,
    key BLOB NOT NULL,
    value BLOB,
    is_empty BOOLEAN NOT NULL DEFAULT FALSE,  -- Distinguishes "empty" from "not cached"
    PRIMARY KEY (block_hash, key)
);

CREATE INDEX IF NOT EXISTS idx_storage_block ON storage(block_hash);

CREATE TABLE IF NOT EXISTS blocks (
    hash BLOB PRIMARY KEY,
    number INTEGER NOT NULL,
    header BLOB NOT NULL,
    parent_hash BLOB NOT NULL
);

CREATE INDEX IF NOT EXISTS idx_blocks_number ON blocks(number);
```

## Tasks

- [x] Add `sqlx` dependency to `pop-fork/Cargo.toml` with `runtime-tokio` and `sqlite` features
- [x] Create `crates/pop-fork/src/cache.rs`
- [x] Implement database initialization with schema
- [x] Implement `StorageCache::open()` and `in_memory()`
- [x] Implement `get()` and `set()` for single values
- [x] Implement `get_batch()` and `set_batch()` for efficiency
- [x] Implement `cache_block()` and `get_block()`
- [x] Handle the three-state return (cached-with-value, cached-empty, not-cached)
- [x] Add unit tests for all CRUD operations
- [x] Test cache persistence (write, close, reopen, read)
- [x] Test concurrent access with connection pool

## Implementation Notes

The implementation matches the spec with these improvements:
- `cache_block()` accepts additional `parent_hash` parameter for complete block info
- Added `clear_block()` method to remove cached data for a specific block
- `BlockInfo` includes `parent_hash` field

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -95,6 +95,20 @@ jobs:
           - macos-latest
     runs-on: ${{ matrix.os }}
     steps:
+      - name: Free up space on runner (Linux)
+        if: matrix.os == 'ubuntu-latest'
+        shell: bash
+        run: |
+          set -xeuo pipefail
+          sudo rm -rf /usr/local/lib/android &
+          sudo rm -rf /usr/share/dotnet &
+          sudo rm -rf /usr/share/swift &
+          if command -v docker &> /dev/null; then
+            sudo docker image prune -af &
+          fi
+          sudo apt-get clean &
+          wait
+
       - uses: actions/checkout@v4
 
       - name: Install stable toolchain


crates/pop-fork/src/cache.rs
```diff
@@ -0,0 +1,564 @@
...
+		Ok(())
+	}
+
+	/// Get multiple cached storage values in a batch.
+	///
+	/// Returns results in the same order as the input keys.
+	pub async fn get_batch(
+		&self,
+		block_hash: H256,
+		keys: &[Vec<u8>],
```
tsenovilla:
```suggestion
		keys: &[&[u8]],
```

We're using `key.as_slice()` inside the function, so we can directly pass slices as input and avoid forcing memory allocation when this is called, and is more flexible.

Also `get` uses `key: &[u8]`, so we should be consistent with that

crates/pop-fork/src/cache.rs
```diff
@@ -0,0 +1,564 @@
...
+		let placeholders: Vec<_> = keys.iter().map(|_| "?").collect();
+		let query = format!(
+			"SELECT key, value, is_empty FROM storage WHERE block_hash = ? AND key IN ({})",
+			placeholders.join(", ")
+		);
+
+		let mut query_builder = sqlx::query(&query).bind(block_hash.as_bytes());
+
+		for key in keys {
+			query_builder = query_builder.bind(key.as_slice());
```
tsenovilla: Just to apply if the previous suggestion is also applied

```suggestion
			query_builder = query_builder.bind(key);
```

crates/pop-fork/src/cache.rs
```diff
@@ -0,0 +1,564 @@
...
+		Ok(keys.iter().map(|key| cache_map.get(key).cloned()).collect())
+	}
+
+	/// Cache multiple storage values in a batch.
+	///
+	/// Uses a transaction for efficiency.
+	pub async fn set_batch(
+		&self,
+		block_hash: H256,
+		entries: &[(Vec<u8>, Option<Vec<u8>>)],
```
tsenovilla:
```suggestion
		entries: &[(&[u8], Option<&[u8]>)],
```

Same reasons as in `get_batch`

tsenovilla (changes requested):
Amazing job with all the `pop fork` research Ale! This PR is a perfect first step to get this awesome feature for `pop-cli`. Left a few comments, but overall I like the design and the approach

crates/pop-fork/src/cache.rs
```diff
@@ -0,0 +1,564 @@
...
+			.await?;
+
+		// Create tables
+		sqlx::query(CREATE_TABLES_SQL).execute(&pool).await?;
+
+		Ok(Self { pool })
+	}
+
+	/// Open an in-memory cache (for testing).
+	pub async fn in_memory() -> Result<Self, CacheError> {
```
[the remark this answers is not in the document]
tsenovilla: Good point. I'd anyway let it behind `test` until we need it somewhere else, but I'm fine if you prefer to keep it outside (specially if you already have an use case in mind for other PRs related to this feature)

tsenovilla (approved):
Looks amazing now! Thanks for addressing it

crates/pop-fork/src/cache.rs
```diff
@@ -0,0 +1,595 @@
...
+			return Ok(vec![]);
+		}
+
+		// Build a SELECT query with dynamic IN clause for batch retrieval.
+		// This fetches all requested keys in a single round-trip to the database,
+		// which is more efficient than individual queries when fetching many keys.
+		// Example: SELECT ... WHERE block_hash = ? AND key IN (?, ?, ?)
+		let placeholders: Vec<_> = keys.iter().map(|_| "?").collect();
+		let query = format!(
+			"SELECT key, value, is_empty FROM storage WHERE block_hash = ? AND key IN ({})",
```
moliholy: Although perfectly valid, I am pretty against using raw SQL syntax if possible, favouring ORMs. Not a strong opinion though, just a personal preference. @al3mart did you consider using https://crates.io/crates/diesel?

moliholy: Reasoning: if the Rust model changes but not the SQL syntax, it all fails. If using diesel (for instance), you have compile-time guarantees that models match.

crates/pop-fork/src/cache.rs
```diff
@@ -0,0 +1,581 @@
...
+#[derive(Clone)]
+pub struct StorageCache {
+	pool: SqlitePool,
+}
+
+impl StorageCache {
+	/// Open or create a cache database at the specified path.
+	///
+	/// Creates the parent directory if it doesn't exist.
+	pub async fn open(path: &Path) -> Result<Self, CacheError> {
```
moliholy: Why not opening directly in RAM memory? I find pretty annoying when chopstics creates sqlite files and some more metadata. Is it so much memory consumed that we cannot afford to have it there?

crates/pop-fork/src/cache.rs
tsenovilla: Fair enough! My point is that even if ORMs nail it, there's also a cost associated to use them. So even if the SQL sentences can get messy, the Rust type-system would keep us safe if we use the macro. But you're right that if it evolves to super complex stuff this approach might be not enough. My concern is if introducing the ORM at this point makes sense or it's better to iterate on the whole fork feature and decide if we need something like this when the time comes (don't misunderstand me, I'm talking just about this concrete PR where I think it's not needed at all, if we introduce further queries somewhere else I'd introduce the ORM asap).

That being said, I'm totally fine with going with the ORM from now as well :)

crates/pop-fork/src/cache.rs
tsenovilla: Good point :) There's still the issue of re-fetching in every restart if we clean the RAM up when halting the command. It's true it shouldn't take long if the ddbb is small but Idk, I'd still keep the files as that allows us, for instance, to run pop fork without internet. Is there any reason for which you're against storing them inside the cache dir?? From your first comment I think the main issue is to push the files in the calling dir as chopsticks does (totally annoying btw hahah)

moliholy: Well, I'd change the description to maximize comfort for the user, to be honest 😓 

Anyway the state will have to be fetched if using the latest block.

al3mart: In any case, let's expose the in-memory option also without the test cfg :+1: Should be alright :)

PR #837, opened 2025-12-11
feat: replace sqlx with diesel

moliholy:
This PR updates the logic to remove sqlx and instead use diesel.

### DIFF
--- a/crates/pop-fork/src/cache.rs
+++ b/crates/pop-fork/src/cache.rs
@@ -12,10 +12,7 @@ use crate::{
 };
 use bb8::CustomizeConnection;
 use diesel::{
-	OptionalExtension,
-	prelude::*,
-	result::{DatabaseErrorKind, Error as DieselError},
-	sqlite::SqliteConnection,
+	OptionalExtension, prelude::*, result::Error as DieselError, sqlite::SqliteConnection,
 };
 use diesel_async::{
 	AsyncConnection, AsyncMigrationHarness, RunQueryDsl,
@@ -27,7 +24,7 @@ use diesel_async::{
 };
 use diesel_migrations::{EmbeddedMigrations, MigrationHarness, embed_migrations};
 use std::{
-	collections::HashMap,
+	collections::{HashMap, HashSet},
 	future::Future,
 	ops::{Deref, DerefMut},
 	path::Path,
@@ -191,14 +188,11 @@ impl StorageCache {
 	///
 	/// Handles acquiring the connection from either the pool or single mutex.
 	/// The connection is automatically returned to the pool or unlocks the mutex when dropped.
-	async fn get_conn(&self) -> Result<ConnectionGuard<'_>, DieselError> {
+	async fn get_conn(&self) -> Result<ConnectionGuard<'_>, CacheError> {
 		match &self.inner {
 			StorageConn::Pool(pool) => {
 				let conn = pool.get().await.map_err(|e| {
-					DieselError::DatabaseError(
-						DatabaseErrorKind::UnableToSendCommand,
-						Box::new(e.to_string()),
-					)
+					CacheError::Connection(ConnectionError::BadConnection(e.to_string()))
 				})?;
 				Ok(ConnectionGuard::Pool(conn))
 			},
@@ -300,6 +294,11 @@ impl StorageCache {
 			return Ok(vec![]);
 		}
 
+		let mut seen = HashSet::with_capacity(keys.len());
+		if keys.iter().any(|key| !seen.insert(key)) {
+			return Err(CacheError::DuplicatedKeys);
+		}
+
 		use crate::schema::storage::columns as sc;
 		let mut conn = self.get_conn().await?;
 
@@ -334,6 +333,11 @@ impl StorageCache {
 	) -> Result<(), CacheError> {
 		if entries.is_empty() {
 			return Ok(());
+		}
+
+		let mut seen = HashSet::with_capacity(entries.len());
+		if entries.iter().any(|(key, _)| !seen.insert(key)) {
+			return Err(CacheError::DuplicatedKeys);
 		}
 
 		// Use a transaction to batch all inserts together.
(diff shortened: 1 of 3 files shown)


crates/pop-fork/migrations/00000000000001_create_storage_and_blocks/up.sql
al3mart: I think the name could use some more 0s :P

Any reason for this ?

crates/pop-fork/src/cache.rs
```diff
@@ -54,50 +55,86 @@ pub struct BlockInfo {
...
 	pub async fn in_memory() -> Result<Self, CacheError> {
-		let pool = SqlitePoolOptions::new()
-			.max_connections(MEMORY_POOL_CONNECTIONS)
-			.connect(SQLITE_MEMORY_URL)
-			.await?;
-
-		sqlx::query(CREATE_TABLES_SQL).execute(&pool).await?;
-
-		Ok(Self { pool })
+		Self::open(None).await
```
al3mart: Nice!

crates/pop-fork/src/models.rs
```diff
@@ -0,0 +1,20 @@
...
+	pub(crate) key: Vec<u8>,
+	pub(crate) value: Option<Vec<u8>>,
+	pub(crate) is_empty: bool,
+}
+
+#[derive(Insertable, Clone)]
+#[diesel(table_name = blocks)]
+pub(crate) struct NewBlockRow {
+	pub(crate) hash: Vec<u8>,
+	pub(crate) number: i32,
```
al3mart: As @tsenovilla hinted previously we might want to use i64 instead.

Polkadot SDK blocks are u32, but we will always fit them in i64 and have safe conversions.
After the previous feedback I went with a `try_into`, but it might be worth reconsidering.

crates/pop-fork/src/cache.rs
```diff
@@ -133,22 +186,54 @@ impl StorageCache {
...
+		// Insert or update the cached storage entry with simple retry on lock contention.
+		use crate::schema::storage::columns as sc;
+		let new = NewStorageRow {
+			block_hash: block_hash.as_bytes().to_vec(),
+			key: key.to_vec(),
+			value: value.map(|v| v.to_vec()),
+			is_empty: value.is_none(),
+		};
+		let mut attempts = 0;
+		loop {
```
al3mart: A few clones per retry. Let's check if optimizing here could be interesting.

crates/pop-fork/src/cache.rs
```diff
@@ -215,23 +303,74 @@ impl StorageCache {
...
+			.map(|(k, v)| NewStorageRow {
+				block_hash: block_hash.as_bytes().to_vec(),
+				key: (*k).to_vec(),
+				value: v.map(|vv| vv.to_vec()),
+				is_empty: v.is_none(),
+			})
+			.collect();
+
+		let mut attempts = 0;
+		loop {
```
al3mart: A few clones per retry. Let's check if optimizing here could be interesting.

crates/pop-fork/src/cache.rs
```diff
@@ -108,17 +145,33 @@ impl StorageCache {
...
-		Ok(row.map(|r| {
-			let is_empty: bool = r.get("is_empty");
-			if is_empty { None } else { Some(r.get("value")) }
-		}))
+		use crate::schema::storage::columns as sc;
+		let bh = block_hash.as_bytes().to_vec();
+		let k = key.to_vec();
+		let row: Option<(Option<Vec<u8>>, bool)> = match &self.inner {
+			StorageConn::Pool(pool) => {
+				let mut conn = pool.get().await?;
```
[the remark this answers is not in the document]
al3mart: I'm checking these suggestions again and they might be problematic on retries...

I think a macro would somewhat better if we want to take this route. An equivalent macro of the above would be something like:

```rust
macro_rules! with_conn {
      ($self:expr, |$conn:ident| $body:expr) => {{
          match &$self.inner {
              StorageConn::Pool(pool) => {
                  let mut $conn = pool.get().await?;
                  $body
              },
              StorageConn::Single(m) => {
                  let mut __guard = m.lock().await;
                  let $conn = &mut *__guard;
                  $body
              },
          }
      }};
  }
```

crates/pop-fork/src/cache.rs
```diff
@@ -54,50 +55,86 @@ pub struct BlockInfo {
...
+				std::fs::create_dir_all(parent)?;
+			}
+			let url = path.display().to_string();
+
+			// Run migrations on a temporary async connection first
+			{
+				let mut conn = SyncConnectionWrapper::<SqliteConnection>::establish(&url).await?;
+				// Apply pragmatic settings for better concurrency on file databases
+				diesel::sql_query("PRAGMA journal_mode=WAL;").execute(&mut conn).await?;
+				diesel::sql_query("PRAGMA busy_timeout=5000;").execute(&mut conn).await?;
```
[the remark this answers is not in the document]
tsenovilla: Agree, we can add a connection customizer to the pool builder to get this behavior: https://docs.rs/bb8/0.9.0/bb8/struct.Builder.html#method.connection_customizer

crates/pop-fork/src/models.rs
al3mart: Let's do it :) Thanks

crates/pop-fork/src/cache.rs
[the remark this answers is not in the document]
tsenovilla: Fixed in https://github.com/r0gue-io/pop-cli/pull/837/commits/8e7ab243a064f9bc82d35701fe5265fdf0e42980

crates/pop-fork/src/cache.rs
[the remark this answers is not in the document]
tsenovilla: Fixed in https://github.com/r0gue-io/pop-cli/pull/837/commits/2b9fef484fedc49cbf9aa04aff4cb898d2cfb24d

Retries aren't a problem at all, just keep the `with_conn` calls inside the loop as you'd do with the macro.

tsenovilla:
Ok @al3mart I just pushed some commits addressing both the feedback you shared yesterday and the one I spotted during my review. Mind to review them again? I left all conversations opened, pointing to the commit where they were solved. Feel free to close them if you're happy with changes or ping me back otherwise.

I'm ok with this PR except for the `clone` stuff, but I wanna know your opinion before applying any change

crates/pop-fork/src/cache.rs
```diff
@@ -5,100 +5,217 @@
...
+	/// Handles acquiring the connection from either the pool or single mutex.
+	/// The connection is automatically returned to the pool or unlocks the mutex when dropped.
+	async fn get_conn(&self) -> Result<ConnectionGuard<'_>, DieselError> {
+		match &self.inner {
+			StorageConn::Pool(pool) => {
+				let conn = pool.get().await.map_err(|e| {
+					DieselError::DatabaseError(
+						DatabaseErrorKind::UnableToSendCommand,
+						Box::new(e.to_string()),
+					)
```
al3mart: Are we mapping to the correct error ?

I see what we have could make sense when a query has failed because we weren't able to acquire the connection.

But if we fail on a regular `let mut conn = self.get_conn().await?;` which we have a few, we will be using an odd error for it.

Maybe we should move from `DieselError` to `CacheError` 🤔 , what do you think @tsenovilla ?

crates/pop-fork/src/models.rs
```diff
@@ -0,0 +1,31 @@
+use crate::schema::{blocks, storage};
+use diesel::{Insertable, Queryable, Selectable};
+
+#[derive(Insertable, Clone)]
+#[diesel(table_name = storage)]
+pub struct NewStorageRow<'a> {
```
al3mart: Could we change visibility to `pub(crate)` instead ? Do we need to expose this at this time ?

crates/pop-fork/src/models.rs
```diff
@@ -0,0 +1,31 @@
...
+	pub block_hash: &'a [u8],
+	pub key: &'a [u8],
+	pub value: Option<&'a [u8]>,
+	pub is_empty: bool,
+}
+
+/// Block row for insertions (uses borrowed data to avoid allocations)
+#[derive(Insertable, Clone)]
+#[diesel(table_name = blocks)]
+pub struct NewBlockRow<'a> {
```
al3mart: Same as above.

crates/pop-fork/src/cache.rs
```diff
@@ -163,45 +300,34 @@ impl StorageCache {
...
+		for (key, value, empty) in rows {
+			let value = if empty { None } else { value };
 			cache_map.insert(key, value);
 		}
 
 		// Return values in the same order as input keys.
 		// Keys not found in cache_map (not in DB) return None (not cached).
 		// Keys found return Some(value) where value is None for empty or Some(bytes) for data.
-		Ok(keys.iter().map(|key| cache_map.get(*key).cloned()).collect())
+		Ok(keys.iter().map(|key| cache_map.remove(*key)).collect())
```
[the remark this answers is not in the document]
tsenovilla: This should be actually done in the function input, even when we used `get` duplicated keys shouldn't be a possibility I think. Good catch! I believe we can make the function fail if duplicated keys are passed

al3mart (approved):
🔥 LFG

Issue #821, opened 2025-12-08
1.3 fork: Remote Storage Layer


al3mart:
# 1.3 Remote Storage Layer

## Summary

Implement the remote storage layer that fetches storage on-demand from live chains with automatic caching.

## Branch

_note that my user `al3mart` is used in the branch names before for the shake of having a full story. Please do check if the branches below are the correct ones_

Base: `al3mart/pop-fork-storage-cache` (or `feat/pop-fork` if #819  and #820 are merged)

## Location

`crates/pop-fork/src/storage/remote.rs`

## Description

Create `RemoteStorageLayer` - the component that fetches storage from a live chain via RPC, automatically caching results. This implements the "lazy loading" pattern where storage is only fetched when needed.

### API to Implement

```rust
pub struct RemoteStorageLayer {
    rpc: Arc<ForkRpcClient>,
    cache: Arc<StorageCache>,
    block_hash: H256,
}

impl RemoteStorageLayer {
    pub fn new(
        rpc: Arc<ForkRpcClient>,
        cache: Arc<StorageCache>,
        block_hash: H256
    ) -> Self;

    /// Get storage value (cache-first, then RPC)
    pub async fn get(&self, key: &[u8]) -> Result<Option<Vec<u8>>>;

    /// Batch get with caching (efficiently handles partial cache hits)
    pub async fn get_batch(&self, keys: &[Vec<u8>]) -> Result<Vec<Option<Vec<u8>>>>;

    /// Get keys by prefix (paginated)
    pub async fn keys_paged(
        &self,
        prefix: &[u8],
        count: u32,
        start: Option<&[u8]>
    ) -> Result<Vec<Vec<u8>>>;
}
```

### Logic Flow for `get()`

```
1. Check cache for (block_hash, key)
   ├─ Found (Some(value) or None) → return cached result
   └─ Not found → continue

2. Fetch from RPC: state_getStorage(key, block_hash)

3. Cache the result (even if None/empty)

4. Return the value
```

### Logic Flow for `get_batch()`

```
1. Check cache for all keys
   ├─ Collect hits (return directly)
   └─ Collect misses (need to fetch)

2. Batch fetch misses from RPC

3. Cache all fetched values

4. Merge and return in original order
```

## Tasks

- [ ] Create `crates/pop-fork/src/storage/` directory
- [ ] Create `crates/pop-fork/src/storage/mod.rs`
- [ ] Create `crates/pop-fork/src/storage/remote.rs`
[...]

PR #840, opened 2025-12-12
feat(fork::remote): Add remote storage layer

al3mart:
# PR: Remote Storage Layer

Addresses #821

## Description

Create `RemoteStorageLayer` - the component that fetches storage from a live chain via RPC, automatically caching results. This implements the "lazy loading" pattern where storage is only fetched when needed.

### Architecture

```text
┌─────────────────────────────────────────────────────────────────┐
│                    RemoteStorageLayer                           │
│                                                                 │
│   get(key) ─────► Cache Hit? ──── Yes ────► Return cached value │
│                        │                                        │
│                        No                                       │
│                        │                                        │
│                        ▼                                        │
│                 Fetch from RPC                                  │
│                        │                                        │
│                        ▼                                        │
│                 Store in cache                                  │
│                        │                                        │
│                        ▼                                        │
│                 Return value                                    │
└─────────────────────────────────────────────────────────────────┘
```

### Implemented API

```rust
#[derive(Clone)]
pub struct RemoteStorageLayer {
    rpc: ForkRpcClient,
    cache: StorageCache,
    block_hash: H256,
}

impl RemoteStorageLayer {
    /// Create a new remote storage layer
    pub fn new(rpc: ForkRpcClient, cache: StorageCache, block_hash: H256) -> Self;

    /// Get the block hash this layer is querying
    pub fn block_hash(&self) -> H256;

    /// Get a reference to the underlying RPC client
    pub fn rpc(&self) -> &ForkRpcClient;

    /// Get a reference to the underlying cache
    pub fn cache(&self) -> &StorageCache;

    /// Get a storage value, fetching from RPC if not cached
    pub async fn get(&self, key: &[u8]) -> Result<Option<Vec<u8>>, RemoteStorageError>;

    /// Get multiple storage values, fetching uncached keys from RPC
    pub async fn get_batch(&self, keys: &[&[u8]]) -> Result<Vec<Option<Vec<u8>>>, RemoteStorageError>;

    /// Prefetch all keys matching a prefix (for warming the cache)
    pub async fn prefetch_prefix(&self, prefix: &[u8], page_size: u32) -> Result<usize, RemoteStorageError>;
}

#[derive(Debug, Error)]
pub enum RemoteStorageError {
    #[error("RPC error: {0}")]
    Rpc(#[from] RpcClientError),
    #[error("Cache error: {0}")]
    Cache(#[from] CacheError),
}
```

### Logic Flow for `get()`

```
1. Check cache for (block_hash, key)
   ├─ Found (Some(value) or None) → return cached result
   └─ Not found → continue

2. Fetch from RPC: state_getStorage(key, block_hash)

3. Cache the result (even if None/empty)

4. Return the value
```

### Logic Flow for `get_batch()`

```
1. Check cache for all keys
   ├─ Collect hits (return directly)
   └─ Collect misses (need to fetch)

2. Batch fetch misses from RPC

3. Cache all fetched values

4. Merge and return in original order
```

## Tasks

- [x] Create `crates/pop-fork/src/remote.rs`
- [x] Create `crates/pop-fork/src/error/remote.rs` with `RemoteStorageError`
- [x] Implement `RemoteStorageLayer::new()`
- [x] Implement `get()` with cache-first logic
- [x] Implement `get_batch()` with partial cache hit handling
- [x] Implement `prefetch_prefix()` for key enumeration and value prefetching
- [x] Add `integration-tests` feature flag to `Cargo.toml`
- [x] Create integration tests in `tests/remote.rs`
- [x] Verify caching works correctly (fetch once, subsequent calls hit cache)
- [x] Export new types from `lib.rs`

### API Changes

| Original Spec | Implementation | Reason |
|---------------|----------------|--------|
| `Arc<ForkRpcClient>` | `ForkRpcClient` | Both `ForkRpcClient` and `StorageCache` already use internal reference counting, so `Clone` is sufficient without `Arc` wrapper |
| `Arc<StorageCache>` | `StorageCache` | Same as above |
| `get_batch(&[Vec<u8>])` | `get_batch(&[&[u8]])` | Reduces unnecessary memory allocations when callers already have byte slices |
| `keys_paged()` | `prefetch_prefix()` | Combined key enumeration with value fetching for practical use case |

### Additional Features

- Added `block_hash()`, `rpc()`, and `cache()` accessor methods for flexibility
- `prefetch_prefix()` returns the count of prefetched keys for progress tracking

## Implementation Notes

- **Three-state cache semantics preserved**: The layer correctly handles:
  - `None` from cache → fetch from RPC
  - `Some(None)` from cache → return `None` (cached empty)
  - `Some(Some(v))` from cache → return `Some(v)` (cached value)

## Dependencies

- Client Wrapper (`ForkRpcClient`) ✓
- Storage Cache (`StorageCache`) ✓

### DIFF
--- a/crates/pop-fork/src/rpc.rs
+++ b/crates/pop-fork/src/rpc.rs
@@ -454,7 +454,7 @@ mod tests {
 			let client = ForkRpcClient::connect(&endpoint).await.unwrap();
 			let hash = client.finalized_head().await.unwrap();
 
-			let keys = vec![
+			let keys = [
 				hex::decode(SYSTEM_NUMBER_KEY).unwrap(),
 				hex::decode(SYSTEM_PARENT_HASH_KEY).unwrap(),
 			];
@@ -530,7 +530,7 @@ mod tests {
 			let hash = client.finalized_head().await.unwrap();
 
 			// Mix of existing and non-existing keys
-			let keys = vec![
+			let keys = [
 				hex::decode(SYSTEM_NUMBER_KEY).unwrap(), // exists
 				vec![0xff; 32],                          // doesn't exist
 			];


crates/pop-fork/src/cache.rs
```diff
@@ -49,7 +49,7 @@ const MAX_LOCK_RETRIES: u32 = 30;
 ///
 /// Enables fast restarts without re-fetching all data from live chains
 /// and reduces load on public RPC endpoints.
-#[derive(Clone)]
+#[derive(Clone, Debug)]
```
tsenovilla: Why do we need this to be `Debug`??

crates/pop-fork/src/remote.rs
```diff
@@ -0,0 +1,289 @@
...
+	/// - Only fetches uncached keys from RPC
+	/// - Caches all fetched values (including empty ones)
+	/// - Returns results in the same order as input keys
+	pub async fn get_batch(
+		&self,
+		keys: &[&[u8]],
+	) -> Result<Vec<Option<Vec<u8>>>, RemoteStorageError> {
+		if keys.is_empty() {
+			return Ok(vec![]);
+		}
```
tsenovilla: I was thinking that we might need to check for duplicated keys here, then realized that this is already checked below when obtaining `cached_results`.

But the same applies for empty keys right? I mean `self.cache.get_storage_batch` already checks this

crates/pop-fork/src/remote.rs
```diff
@@ -0,0 +1,289 @@
...
+		fn assert_debug<T: std::fmt::Debug>() {}
+		assert_debug::<RemoteStorageLayer>();
+	}
+
+	#[test]
+	fn remote_storage_error_is_send_sync() {
+		// Compile-time check that RemoteStorageError is Send + Sync
+		fn assert_send_sync<T: Send + Sync>() {}
+		assert_send_sync::<RemoteStorageError>();
+	}
```
tsenovilla: Are this one needed? I mean the compiler will complain if we try to use those types as if they implemented a trait they dont

crates/pop-fork/tests/remote.rs
```diff
@@ -0,0 +1,158 @@
...
+//!
+//! Run with: `cargo nextest run -p pop-fork --features integration-tests --test remote`
+
+#![cfg(feature = "integration-tests")]
+
+use pop_fork::{ForkRpcClient, RemoteStorageLayer, StorageCache};
+use url::Url;
+
+/// Paseo testnet public RPC endpoint.
+const PASEO_ENDPOINT: &str = "wss://rpc.ibp.network/paseo";
```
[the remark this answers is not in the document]
tsenovilla: Related: https://github.com/r0gue-io/pop-cli/issues/846

crates/pop-fork/src/cache.rs
al3mart: Yeah, if I remember correctly this was only for tests. Will look into it :)

crates/pop-fork/tests/remote.rs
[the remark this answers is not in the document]
al3mart: Yeah, I talked with Alex previously and we agreed that for the time being we will use TestNode instances.

crates/pop-fork/src/cache.rs
al3mart: While I think your point is valid. I was xchecking with https://rust-lang.github.io/api-guidelines/checklist.html?highlight=Debug#rust-api-guidelines-checklist and it seems that implementing Debug for public types it is considered a good practice.

crates/pop-fork/tests/remote.rs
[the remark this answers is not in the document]
al3mart: [1cb0e1d](https://github.com/r0gue-io/pop-cli/pull/840/commits/1cb0e1ddbfd2acfb284af060a1f842354424b2ce)

crates/pop-fork/src/remote.rs
al3mart: [5e87a0c](https://github.com/r0gue-io/pop-cli/pull/840/commits/5e87a0cacb2c4bc6283202b97bd4a6edd41a0dc3)

Issue #823, opened 2025-12-08
1.5 fork: Storage Provider Trait


al3mart:
# 1.5 Storage Provider Trait

## Summary

Define the `StorageProvider` trait that provides a unified interface for all storage layer implementations.

## Branch

Base: `feat/pop-fork`

## Location

`crates/pop-fork/src/storage/mod.rs`

## Description

Create the `StorageProvider` trait that abstracts over different storage implementations (remote, local, cached). This trait is used by the runtime executor to read storage without knowing the underlying implementation.

### API to Implement

```rust
use async_trait::async_trait;

#[async_trait]
pub trait StorageProvider: Send + Sync + Clone {
    /// Get storage value by key
    async fn get(&self, key: &[u8]) -> Result<Option<Vec<u8>>>;

    /// Batch get - default implementation calls get() in parallel
    /// Note: Uses &[&[u8]] to avoid unnecessary allocations
    async fn get_batch(&self, keys: &[&[u8]]) -> Result<Vec<Option<Vec<u8>>>> {
        futures::future::try_join_all(keys.iter().map(|k| self.get(k))).await
    }

    /// Check if key was explicitly deleted (for local layers)
    fn is_deleted(&self, key: &[u8]) -> bool {
        false
    }
}
```

### Implementations

```rust
// Remote storage layer
#[async_trait]
impl StorageProvider for RemoteStorageLayer {
    async fn get(&self, key: &[u8]) -> Result<Option<Vec<u8>>> {
        // Cache-first, then RPC
        self.get(key).await
    }

    async fn get_batch(&self, keys: &[&[u8]]) -> Result<Vec<Option<Vec<u8>>>> {
        // Optimized batch with partial cache hits
        self.get_batch(keys).await
    }
}

// Local storage layer
#[async_trait]
impl<P: StorageProvider + Clone> StorageProvider for LocalStorageLayer<P> {
    async fn get(&self, key: &[u8]) -> Result<Option<Vec<u8>>> {
        // Check local mods, then cascade to parent
        // ...
    }

    fn is_deleted(&self, key: &[u8]) -> bool {
        // Check if explicitly deleted or matches deleted prefix
        // ...
    }
}
```

## Tasks

- [ ] Add `async-trait` dependency to `pop-fork/Cargo.toml`
- [ ] Define `StorageProvider` trait in `storage/mod.rs`
- [ ] Implement `StorageProvider` for `RemoteStorageLayer`
[...]

Issue #822, opened 2025-12-08
1.4 fork: Local Storage Layer (Modifications)


al3mart:
# 1.4 Local Storage Layer (Modifications)

## Summary

Implement the local storage layer that tracks modifications on top of a parent storage provider without mutating the original.

## Branch

```
al3mart/pop-fork-local-storage
```

Base: `feat/pop-fork`

## Location

`crates/pop-fork/src/storage/local.rs`

## Description

Create `LocalStorageLayer` - a copy-on-write storage layer that tracks modifications. When building new blocks, changes are written to this layer while the underlying remote storage remains unchanged. This enables:
- Building blocks without modifying cached state
- Easy rollback of failed transactions
- Nested layers for transaction isolation

### API to Implement

```rust
/// Note: Parent is stored directly (not Arc-wrapped) because StorageProvider
/// implementations should already handle internal sharing via Clone.
/// LocalStorageLayer derives Clone when P: Clone.
#[derive(Clone)]
pub struct LocalStorageLayer<P: StorageProvider + Clone> {
    parent: P,
    modifications: Arc<RwLock<HashMap<Vec<u8>, Option<Vec<u8>>>>>,
    deleted_prefixes: Arc<RwLock<Vec<Vec<u8>>>>,
}

impl<P: StorageProvider + Clone> LocalStorageLayer<P> {
    /// Create a new layer on top of a parent
    pub fn new(parent: P) -> Self;

    /// Get storage value (check local first, then parent)
    pub async fn get(&self, key: &[u8]) -> Result<Option<Vec<u8>>>;

    /// Set a storage value (local modification)
    /// Note: Uses &[u8] to avoid forcing callers to allocate
    pub fn set(&self, key: &[u8], value: Option<&[u8]>);

    /// Delete all keys with prefix
    pub fn delete_prefix(&self, prefix: &[u8]);

    /// Check if a key was explicitly deleted
    pub fn is_deleted(&self, key: &[u8]) -> bool;

    /// Get all modifications as a diff
    pub fn diff(&self) -> Vec<(Vec<u8>, Option<Vec<u8>>)>;

    /// Merge another layer's modifications into this one
    pub fn merge(&self, other: &LocalStorageLayer<impl StorageProvider + Clone>);

    /// Create a child layer for nested modifications
    pub fn child(&self) -> LocalStorageLayer<Self>;
}
```

### Logic Flow for `get()`

```
1. Check local modifications
[...]

PR #855, opened 2025-12-18
feat: Local Storage Layer

tsenovilla:
Closes
-  #822 

This PR introduces a LocalStorageLayer that can be used to keep track of local modifications without affecting the underlying remote storage and cache. 

It directly uses RemoteStorageLayer right now, but will probably become generic after #823

### DIFF
--- a/crates/pop-fork/src/local.rs
+++ b/crates/pop-fork/src/local.rs
@@ -757,6 +757,7 @@ mod tests {
 	async fn set_stores_value() {
 		let ctx = create_test_context().await;
 		let layer = create_layer(&ctx);
+		let block = layer.get_latest_block_number();
 
 		let key = b"key";
 		let value = b"value";
@@ -764,7 +765,7 @@ mod tests {
 		layer.set(key, Some(value)).unwrap();
 
 		// Verify via get
-		let result = layer.get(ctx.block_number, key).await.unwrap();
+		let result = layer.get(block, key).await.unwrap();
 		assert_eq!(result, Some(Arc::new(value.as_slice().to_vec())));
 	}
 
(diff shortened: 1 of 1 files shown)


crates/pop-fork/src/local.rs
```diff
@@ -181,15 +211,14 @@ impl LocalStorageLayer {
 			return Ok(cached.map(Arc::new));
 		}
 
-		// Case 3: Block before or at fork point - query remote directly
-		// First, get the block hash from the blocks table
-		let block = self.parent.cache().get_block_by_number(block_number).await?;
+		// Case 3: Block before or at fork point - fetch and cache block if needed
```
al3mart: We should only hit this if the user were to query for historical data prior to the fork height. 

If query is at fork point point the state should already exist. Am I right ?

crates/pop-fork/src/remote.rs
```diff
@@ -242,6 +242,53 @@ impl RemoteStorageLayer {
...
+	/// * `Ok(Some(block_row))` - Block was fetched and cached successfully
+	/// * `Ok(None)` - Block number doesn't exist
+	/// * `Err(_)` - RPC or cache error
+	///
+	/// # Caching Behavior
+	/// - Fetches block hash and data from block number using `chain_getBlockHash` and
+	///   `chain_getBlock`
+	/// - Caches block metadata (hash, number, parent_hash, header) in the cache
+	/// - If block is already cached, this will update the cache entry
+	pub async fn fetch_and_cache_block_by_number(
```
al3mart: Let's change the visibility here to pub(crate) ?

crates/pop-fork/src/local.rs
al3mart: This does a better job at explaining what I had in mind: https://github.com/r0gue-io/pop-cli/pull/855#discussion_r2642939717

Issue #824, opened 2025-12-08
2.1 fork: Runtime Executor (smoldot Integration)


al3mart:
# 2.1 Runtime Executor (smoldot Integration)

## Summary

Implement the runtime executor using smoldot to execute Substrate runtime calls against forked state.

## Branch

Base: `al3mart/pop-fork-storage-trait` (or after Phase 1 is merged)

## Location

`crates/pop-fork/src/executor.rs`

## Description

Create `RuntimeExecutor` - a wrapper around smoldot's executor that runs Substrate runtime calls (like `Core_version`, `BlockBuilder_apply_extrinsic`, etc.) against our lazy-loaded storage. This is the core component that enables executing transactions on the forked state.

### Why smoldot?

smoldot already implements all ~50 Substrate host functions:
- Storage operations (get, set, clear, exists)
- Cryptographic operations (sr25519, ed25519, ecdsa)
- Hashing (blake2, keccak, sha2, twox)
- Memory allocation
- Logging and debugging

We intercept storage operations to route them through our `StorageProvider`.

### API to Implement

```rust
pub struct RuntimeExecutor {
    runtime_code: Vec<u8>,
    heap_pages: u32,
}

pub struct RuntimeCallResult {
    pub output: Vec<u8>,
    pub storage_diff: Vec<(Vec<u8>, Option<Vec<u8>>)>,
}

impl RuntimeExecutor {
    /// Create executor with runtime WASM code
    pub fn new(runtime_code: Vec<u8>) -> Result<Self>;

    /// Execute a runtime call
    pub async fn call<S: StorageProvider>(
        &self,
        method: &str,
        args: &[u8],
        storage: &S,
    ) -> Result<RuntimeCallResult>;
}
```

### Execution Flow

```rust
pub async fn call<S: StorageProvider>(
    &self,
    method: &str,
    args: &[u8],
    storage: &S,
) -> Result<RuntimeCallResult> {
    let prototype = HostVmPrototype::new(Config {
        module: &self.runtime_code,
        heap_pages: self.heap_pages,
        // ...
    })?;

    let mut vm = prototype.run(method, args)?;
    let mut storage_diff = Vec::new();

    loop {
        match vm {
            HostVm::ReadyToRun(r) => {
                vm = r.run();
            }

            // Intercept storage reads
            HostVm::ExternalStorageGet(req) => {
                let key = req.key().as_ref().to_vec();
                let value = storage.get(&key).await?;
[...]

PR #869, opened 2025-12-22
feat(pop-fork): Runtime Executor (smoldot Integration) 

al3mart:
Addresses: https://github.com/r0gue-io/pop-cli/issues/824

# Runtime Executor

## Summary

This PR implements the runtime executor using smoldot for executing Substrate runtime calls against forked chain state, along with significant improvements to the storage layer infrastructure.

- Adds `RuntimeExecutor` - a smoldot-based executor for running Substrate runtime calls
- Adds `next_key` support for key enumeration during runtime execution
- Simplifies block lookup logic by removing redundant RPC calls

### New Files

| File | Description |
|------|-------------|
| `crates/pop-fork/src/executor.rs` | Runtime executor using smoldot (~720 lines) |
| `crates/pop-fork/src/error/executor.rs` | Executor-specific error types |
| `crates/pop-fork/src/strings/executor.rs` | String constants for executor module |

### RuntimeExecutor

A wrapper around smoldot's executor that runs Substrate runtime calls against lazy-loaded storage.

```rust
let executor = RuntimeExecutor::new(runtime_code, None)?;
let result = executor.call("Core_version", &[], &storage).await?;
```

**Handles all required smoldot callbacks:**
- `StorageGet` - Reads from LocalStorageLayer with child trie support
- `NextKey` - Key enumeration for prefix iteration
- `SignatureVerification` - With mock modes for testing (None, MagicSignature, AlwaysValid)
- `LogEmit` - Captures runtime logs
- `ClosestDescendantMerkleValue` - Falls back to smoldot calculation
- `Offchain*` - Offchain worker support (timestamp, random seed, storage)

**Configuration options:**
- `SignatureMockMode` - Control signature verification behavior for testing
- `ExecutorConfig` - Configure log levels, unresolved imports, proof size

### Key Enumeration

New `next_key` support for both storage layers:

```rust
// LocalStorageLayer - filters deleted prefixes
let next = local.next_key(prefix, current_key).await?;

// RemoteStorageLayer - fetches from RPC
let next = remote.next_key(block_hash, prefix, current_key).await?;
```

## Test Plan

- [x] Unit tests for `RuntimeExecutor` (magic signature validation, child key prefixing, config defaults)
- [x] Unit tests for `get_keys_by_prefix` and `count_keys_by_prefix`
- [x] Unit tests for `increment_prefix` helper
- [x] Integration with existing LocalStorageLayer tests
- [x] Integration test: execute `Core_version` against live chain state
- [x] Integration test: execute `Metadata_metadata`

## Dependencies

- `smoldot = "0.21.0"` (new workspace dependency)
- `hex` moved from dev-dependencies to dependencies

### DIFF
--- a/crates/pop-fork/src/executor.rs
+++ b/crates/pop-fork/src/executor.rs
@@ -92,7 +92,7 @@ struct ArcBytes(Arc<Vec<u8>>);
 
 impl AsRef<[u8]> for ArcBytes {
 	fn as_ref(&self) -> &[u8] {
-		&self.0.as_ref()
+		self.0.as_ref()
 	}
 }
 


crates/pop-fork/src/local.rs
```diff
@@ -221,6 +220,63 @@ impl LocalStorageLayer {
...
+	/// # Behavior
+	/// 1. Queries the parent layer for the next key
+	/// 2. Skips keys that match deleted prefixes
+	/// 3. Does not consider locally modified keys (they are transient)
+	///
+	/// # Note
+	/// This method currently delegates directly to the parent layer.
+	/// Locally modified keys are not included in key enumeration since
+	/// they represent uncommitted changes.
+	pub async fn next_key(
```
tsenovilla: maybe adding a test for this function could be helpful?

crates/pop-fork/src/executor.rs
```diff
@@ -0,0 +1,1021 @@
...
+							req.key().as_ref().iter().copied(),
+						)
+					} else {
+						req.key().as_ref().to_vec()
+					};
+
+					// Check local changes first
+					if let Some(value) = storage_changes.get(&key) {
+						req.inject_value(
+							value.clone().map(|x| (iter::once(x), TrieEntryVersion::V1)),
```
tsenovilla: Same as below, `value` should be enough without needing of clones

crates/pop-fork/src/local.rs
```diff
@@ -486,6 +542,23 @@ impl LocalStorageLayer {
...
+	///
+	/// # Behavior
+	/// - The child shares the same `modifications` and `deleted_prefixes` via `Arc`
+	/// - Changes in the child affect the parent and vice versa
+	/// - Useful for creating temporary scopes that can be discarded
+	///
+	/// # Note
+	/// This is currently a simple clone. In the future, this may be updated to create
+	/// true isolated child layers with proper parent-child relationships.
+	pub fn child(&self) -> LocalStorageLayer {
```
tsenovilla: Why did you readded this function?? I don't see where it's being used in the executor context

al3mart: That was unintended. Seems I left it there while doing the latest rebase. Missing `next_key` tests gone for the same reason :S

crates/pop-fork/src/executor.rs
```diff
@@ -0,0 +1,1021 @@
...
+
+/// Check if a signature is a magic test signature.
+///
+/// Magic signatures start with `0xdeadbeef` and are padded with `0xcd`.
+fn is_magic_signature(signature: &[u8]) -> bool {
+	signature.starts_with(magic_signature::PREFIX) &&
+		signature[magic_signature::PREFIX.len()..]
+			.iter()
+			.all(|&b| b == magic_signature::PADDING)
+}
```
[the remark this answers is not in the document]
tsenovilla: Oops, In my mind ECDSA was 64 bytes too😅 Then I said nothing, it's ok as it's, just suggested it in case we could have it simpler but no worries

tsenovilla (approved):

PR #875, opened 2025-12-22
fix: find build artifacts

Daanvdplas:
This PR fixes contract artifact discovery for workspace members. It ensures the lookup checks the workspace target/ink when the contract lives inside a workspace, so commands correctly find the built .contract file instead of only checking the member’s local target/ink.

### DIFF
--- a/crates/pop-cli/src/common/contracts.rs
+++ b/crates/pop-cli/src/common/contracts.rs
@@ -586,7 +586,8 @@ mod tests {
 	#[test]
 	fn build_contract_artifacts_reports_error() -> anyhow::Result<()> {
 		let temp_dir = tempfile::tempdir()?;
-		let err = build_contract_artifacts(temp_dir.path(), true, Verbosity::Quiet, None)
+		let mut cli = MockCli::new();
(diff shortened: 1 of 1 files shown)


crates/pop-common/src/helpers.rs
```diff
@@ -58,6 +58,35 @@ pub fn get_relative_or_absolute_path(base: &Path, full: &Path) -> PathBuf {
...
+		let manifest_path = dir.join("Cargo.toml");
+		if manifest_path.is_file() &&
+			let Ok(manifest) = from_path(dir) &&
+			manifest.workspace.is_some()
+		{
+			return Some(dir.to_path_buf());
+		}
+		current = dir.parent();
+	}
+	None
```
moliholy: Nit: this walks up and returns the first ancestor with a `[workspace]` section, but it doesn't check that the starting project is actually listed in that workspace's `members`. An unrelated workspace higher up the tree could match.

The risk is low since `find_contract_artifact_path` checks artifact existence, but it could still return a wrong path if an unrelated workspace happens to have an artifact with the same package name.

PR #877, opened 2025-12-29
fix: default to `pop call chain`

moliholy:
This PR adapt the `pop call` command so that it defaults to `pop call chain` if not in a contract directory. Current behaviour is failing in case the user is not in a contract or chain directory.

### DIFF
--- a/crates/pop-cli/src/commands/call/mod.rs
+++ b/crates/pop-cli/src/commands/call/mod.rs
@@ -38,20 +38,20 @@ impl CallArgs {
 		}
 
 		// Auto-detect project type based on current directory
-		let current_dir = std::env::current_dir()?;
-
 		#[cfg(feature = "contract")]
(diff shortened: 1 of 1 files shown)


crates/pop-cli/src/commands/call/mod.rs
```diff
@@ -146,19 +146,7 @@ ink = "5.1.1"
...
-			let cargo_toml = r#"[package]
-name = "other-project"
-version = "0.1.0"
-
-[dependencies]
-regex = "1.10"
-"#;
-			fs::write(temp_dir.path().join("Cargo.toml"), cargo_toml)?;
-			assert!(CallArgs { command: None }.resolve_command().is_err());
+			assert!(matches!(CallArgs { command: None }.resolve_command()?, Command::Chain(_)));
```
Daanvdplas: Why remove the test of checking a project with a Cargo.toml file but without any relevant dependencies?

Issue #789, opened 2025-11-20
`pop up` defaults to `--use-wallet` when passing `-y`


al3mart:
When passing `pop up` the `-y` flag, it might be a better UX to default to a dev account like Alice instead of the `--use-wallet` flow.

Counter argument to this being that if pop is used in automations or other scenarios where the user is not necessarily a human, one could assume all necessary flags and options are being passed to the CLI. Making `--use-wallet` a better fit when `-y` is being used.

Daanvdplas:
It should error and ask for the signer, provided with --use-wallet or --suri

PR #884, opened 2025-12-31
feat: require signer for pop up when skipping confirmation

moliholy:
Closes https://github.com/r0gue-io/pop-cli/issues/789

This PR enforces users to specifying a signer when using `pop call` and `pop up`.


Daanvdplas (approved):
LGTM, left two small comments.

Issue #828, opened 2025-12-08
3.3 fork: Mock Inherent Providers


al3mart:
# 3.3 Mock Inherent Providers

## Summary

Implement mock inherent providers that generate necessary inherent extrinsics for block production.

## Branch

_note that my user al3mart is used in the branch names for the shake of having a full story. Please do check if the branches below are the correct ones_

Base: `al3mart/pop-fork-block`

## Location

`crates/pop-fork/src/inherents.rs`

## Description

Create inherent providers that generate the mandatory inherent extrinsics needed for valid blocks. For a forked chain with mock consensus, we need at minimum:
- **Timestamp** - Advances time for each block
- **Parachain validation data** (for parachains) - Mock validation from relay chain

### API to Implement

```rust
use async_trait::async_trait;

#[async_trait]
pub trait InherentProvider: Send + Sync {
    /// Identifier for this inherent
    fn identifier(&self) -> &'static str;

    /// Generate inherent extrinsic for a new block
    /// Returns None if no inherent is needed
    async fn provide(&self, parent: &Block, metadata: &ChainMetadata) -> Result<Option<Vec<u8>>>;
}

/// Timestamp inherent - advances time each block
pub struct TimestampInherent {
    slot_duration_ms: u64,
}

impl TimestampInherent {
    pub fn new(slot_duration_ms: u64) -> Self;

    /// Create with default 6-second slots
    pub fn default_relay() -> Self {
        Self::new(6000)
    }

    /// Create with default 12-second slots (parachains)
    pub fn default_para() -> Self {
        Self::new(12000)
    }
}

#[async_trait]
impl InherentProvider for TimestampInherent {
    fn identifier(&self) -> &'static str {
        "Timestamp"
    }

    async fn provide(&self, parent: &Block, metadata: &ChainMetadata) -> Result<Option<Vec<u8>>> {
        // 1. Read current timestamp from parent storage
        let timestamp_key = metadata.storage_key("Timestamp", "Now")?;
        let current = parent.storage().get(&timestamp_key).await?
            .map(|v| u64::decode(&mut &v[..]))
            .transpose()?
            .unwrap_or(0);

        // 2. Calculate new timestamp
        let new_timestamp = current + self.slot_duration_ms;

        // 3. Encode as Timestamp.set() call
        let call = metadata.encode_call("Timestamp", "set", &json!({
            "now": new_timestamp
        }))?;

        // 4. Wrap in unsigned extrinsic
[...]

PR #898, opened 2026-01-18
feat(pop-fork): inherents

al3mart:
Solves #828

### DIFF
--- a/crates/pop-fork/src/block.rs
+++ b/crates/pop-fork/src/block.rs
@@ -45,6 +45,7 @@
 
 use crate::{BlockError, ForkRpcClient, LocalStorageLayer, RemoteStorageLayer, StorageCache};
 use scale::Decode;
+use std::sync::Arc;
 use subxt::{Metadata, config::substrate::H256, ext::codec::Encode};
 use url::Url;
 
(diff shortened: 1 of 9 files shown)


crates/pop-fork/src/inherent/timestamp.rs
```diff
@@ -0,0 +1,502 @@
...
+		///
+		/// This test is ignored by default since it requires network access.
+		/// Run manually with:
+		/// ```bash
+		/// cargo nextest run -p pop-fork --lib "get_slot_duration_from_live_aura_chain" --run-ignored all --nocapture
+		/// ```
+		#[tokio::test(flavor = "multi_thread")]
+		#[ignore = "requires network access to live chain"]
+		async fn get_slot_duration_from_live_aura_chain() {
+			const ASSET_HUB_PASEO_ENDPOINT: &str = "wss://sys.ibp.network/asset-hub-paseo";
```
tsenovilla: As we have said in the past, we should avoid connecting to live chains in pop. However I don't see a better way to test this thing (and not personally against using live chains at all).. Maybe checking that it gets the right value from the ink node is enough for now?

tsenovilla: yeah definitely, let's keep it unchanged for the moment👍

crates/pop-fork/src/block.rs
```diff
@@ -82,6 +83,12 @@ pub struct Block {
 	/// The parent block. Keeping blocks in memory is cheap as the `LocalStorageLayer` is shared
 	/// between all fork-produced blocks.
 	pub parent: Option<Box<Block>>,
+	/// The runtime metadata for this block.
+	///
+	/// Fetched via RPC at fork time and inherited by child blocks.
+	/// This enables dynamic lookup of pallet and call indices for inherent
+	/// providers instead of relying on hardcoded values.
+	metadata: Metadata,
```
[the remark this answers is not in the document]
al3mart: Actually #900 should make this quite easy to implement

PR #901, opened 2026-01-21
chore(pop-fork): improve metadata handling

al3mart:
These changes remove `Metadata` from the `Block` struct and benefits from the already shared `LocalLayer` to maintain a record of what metadata is applicable to which blocks.

It also includes means to check if a code upgrade has happened.

As per the current implementation whether the metadata registry needs to reflect a change of metadata is handled on every block finalization. But maybe this could be improved later on if we see there are benefits from moving this logic somewhere else.

### DIFF
--- a/crates/pop-fork/src/builder.rs
+++ b/crates/pop-fork/src/builder.rs
@@ -58,8 +58,8 @@ use crate::{
 	Block, BlockBuilderError, RuntimeCallResult, RuntimeExecutor, inherent::InherentProvider,
 	strings::builder::runtime_api,
 };
-use scale::Encode;
-use subxt::config::substrate::H256;
+use scale::{Decode, Encode};
+use subxt::{Metadata, config::substrate::H256};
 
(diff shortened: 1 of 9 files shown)


crates/pop-fork/src/builder.rs
```diff
@@ -396,6 +411,39 @@ impl BlockBuilder {
...
 	}
 
+	/// Register new metadata after a runtime upgrade.
+	///
+	/// This is called when the `:code` storage key was modified in the parent block,
+	/// indicating a runtime upgrade occurred. Since the BlockBuilder's executor is
+	/// already running the new runtime (it was initialized with code from current state),
+	/// we simply call `Metadata_metadata` using the existing executor and register
+	/// the metadata for the current block (the first block using the new runtime).
+	async fn register_new_metadata(&self) -> Result<(), BlockBuilderError> {
```
tsenovilla: super nice :)

Issue #831, opened 2025-12-08
4.4 fork: Transaction Pool


al3mart:
# 4.4 Transaction Pool (Optional/Minimal)

## Summary

Implement a minimal transaction queue for collecting submitted extrinsics.

## Branch

_note that my user al3mart is used in the branch names for the shake of having a full story. Please do check if the branches below are the correct ones_

Base: `al3mart/pop-fork-blockchain`

## Location

`crates/pop-fork/src/tx_pool.rs`

### Do We Need This?

**Short answer**: A simple queue is useful but a full transaction pool is NOT needed.

**How chopsticks does it** (from `packages/core/src/blockchain/txpool.ts`):
```typescript
// Literally just an array
readonly #pool: { extrinsic: HexString; signer: string }[] = []
```

That's it. No validation, no nonce management, no fee ordering. Just a queue that gets drained when building blocks.

### Why NOT use Substrate's `sc-transaction-pool`?

Substrate's transaction pool (`substrate/client/transaction-pool`) is designed for production nodes:
- Heavy dependencies on Substrate client infrastructure (`sc-client-api`, `sp-blockchain`, etc.)
- Complex ready/future pools with nonce tracking
- Fee-based prioritization
- Transaction re-validation on new blocks
- Would bring significant dependency bloat

For a fork/simulation tool, this is massive overkill.

### Recommended Approach

**Option A: Inline in RPC Server (Simplest)**
No separate module. The RPC server maintains a simple `Vec<Vec<u8>>`:
```rust
// In RpcContext
pending_extrinsics: RwLock<Vec<Vec<u8>>>,
```

**Option B: Minimal TxPool (Like Chopsticks)**
A tiny module (~50-100 lines) if we want cleaner separation:

```rust
use tokio::sync::RwLock;

/// Minimal transaction queue - NOT a production transaction pool.
/// No validation, no ordering, just a FIFO queue.
pub struct TxPool {
    pending: RwLock<Vec<Vec<u8>>>,
}

impl TxPool {
    pub fn new() -> Self {
        Self { pending: RwLock::new(Vec::new()) }
    }

    /// Add extrinsic to queue, returns hash
    pub async fn submit(&self, extrinsic: Vec<u8>) -> H256 {
        let hash = sp_crypto_hashing::blake2_256(&extrinsic).into();
        self.pending.write().await.push(extrinsic);
        hash
    }

    /// Get all pending extrinsics
[...]

PR #902, opened 2026-01-22
feat: pop fork txpool

tsenovilla:
Solves #831


al3mart (approved):
Seems perfectly fine with the context we have at this time. Maybe we might need changes once we include a user & consumer of the pool.

PR #910, opened 2026-01-30
feat: pacman

moliholy:
This PR adds two CI steps to:

1) Generate the package for pacman and upload it to the releases.
2) Upload a new version to the general arch linux repository.

For reference: we already have the arch linux pop-cli's registration: https://aur.archlinux.org/packages/pop-cli


al3mart (commented):
AUR related changes look alright. Will love to test things before approving in an arch vm.

It seems there are a couple clippy complains that might need attention :)

PR #913, opened 2026-02-03
refactor(aur): PKGBUILD

al3mart:
- Includes `'lld'` and `'rust-wasm'` to the list of dependencies. Was having issues installing the package without them.
- Uses production profile.
- And removed the `check` step. We were executing all test immediately after the user finished installing the package.


moliholy (approved):
LGTM!

PR #916, opened 2026-02-04
chore(pop-fork): extrinsic presubmission validation

al3mart:


crates/pop-fork/src/blockchain.rs
```diff
@@ -51,15 +51,150 @@
...
 
 pub type BlockBody = Vec<Vec<u8>>;
+
+// Transaction validity types for decoding TaggedTransactionQueue_validate_transaction results.
+
+/// Result of transaction validation.
+///
+/// Mirrors `sp_runtime::transaction_validity::TransactionValidity`.
+#[derive(Debug, Clone, Decode)]
+pub enum TransactionValidity {
```
tsenovilla: Why do we need to redefine all these types? `sp_runtime` is already in the workspace and contains them

PR #921, opened 2026-02-05
refactor(pop-fork): rpc server

al3mart:


crates/pop-fork/src/inherent/timestamp.rs
```diff
@@ -275,12 +275,23 @@ impl InherentProvider for TimestampInherent {
 		// Calculate new timestamp
 		let new_timestamp = current_timestamp.saturating_add(slot_duration);
 
+		eprintln!(
```
tsenovilla: can we use the logging system instead of eprints? Like we do in the target branch. So these eprintln! should be `log::info!` or something like that

PR #926, opened 2026-02-06
chore(release): prepare for 0.13.0

Daanvdplas:
## Summary
- Bump workspace version to 0.13.0
- Regenerate CHANGELOG for v0.13.0

## Testing
- Not run (release prep only)


AlexD10S (commented):
Did you use RAI? When I tested the SKILL, it was also picking up deprecated messages and removing the unnecessary code, e.g: https://github.com/r0gue-io/pop-cli/blob/106addb1f1914adb7b85ec1f54f3dbce8ca618ac/crates/pop-cli/src/commands/call/contract.rs#L767

PR #929, opened 2026-02-09
feat(pop up): fork integration with `pop up`

al3mart:
## Summary

An alternative implementation without modifications to `pop up` can be found at: https://github.com/r0gue-io/pop-cli/pull/930

Adds `--fork [<BLOCK>]` flag to `pop up paseo|kusama|polkadot|westend`, enabling users to fork a live relay chain directly from the `pop up` interface instead of manually resolving RPC endpoints with `pop fork -e`.

Also extends `pop fork` with `--at <BLOCK>` for forking at a specific block number, and improves detach mode to display fork progress and results in the terminal before backgrounding.

## Overview

```mermaid
flowchart LR
    A["pop up paseo --fork"] --> B["Resolve RPC endpoints"]
    B --> C["Try endpoint"]
    C -->|Success| D["Fork running"]
    C -->|Failure| E["Warn & try next"]
    E --> C
    A2["pop up paseo --fork 1234"] --> B
```

### Usage

```bash
# Fork at latest finalized block
pop up paseo --fork

# Fork at specific block
pop up paseo --fork 1234

# Fork in background
pop up polkadot --fork --detach

# Full control (unchanged)
pop fork -e wss://custom-rpc.example.com
```

### Detach mode output (before → after)

**Before:** returns immediately with PID only, fork info buried in log file.

**After:** waits for the fork to be ready and displays results inline:

```
◇ Forking chain(s) (detached mode)
│ Forking wss://rpc.ibp.network/paseo...
◇ Forked paseo at block #10306170 -> ws://127.0.0.1:9945
◇ Fork started with PID 12345
│ Log file: /tmp/...
└ Run `kill -9 12345` or `pop clean node -p 12345` to stop.
```

### RPC endpoint fallback

When an endpoint fails, the next one from the pool is tried automatically:

```
◇ Forking chain(s)
│ Forking wss://bad-rpc.example.com...
⚠ wss://bad-rpc.example.com did not respond, trying next endpoint...
◇ Forking chain(s)
│ Forking wss://good-rpc.example.com...
◇ Forked paseo at block #10306170 -> ws://127.0.0.1:9945
```

## How to test

1. **Basic fork:** `cargo run -- up paseo --fork` → should fork Paseo at latest finalized block
2. **Fork at block:** `cargo run -- up paseo --fork 1234` → should fork at block 1234
3. **Detach mode:** `cargo run -- up paseo --fork --detach` → should show fork info before backgrounding
4. **RPC fallback:** temporarily break the first RPC URL in the pool → should fall back to next endpoint
5. **Help output:** `cargo run -- up paseo --help` → should show `--fork [<BLOCK>]` flag
6. **Direct fork unchanged:** `cargo run -- fork -e wss://rpc.ibp.network/paseo` → existing behavior preserved

### DIFF
--- a/crates/pop-cli/src/commands/fork.rs
+++ b/crates/pop-cli/src/commands/fork.rs
@@ -93,7 +93,7 @@ pub(crate) struct ForkArgs {
 	pub serve: bool,
 
 	/// Internal flag: path to write readiness info to (used by detach mode).
-	#[arg(long, hide = true)]
+	#[arg(long, hide = true, requires = "serve")]
 	#[serde(skip)]
 	pub ready_file: Option<PathBuf>,
 }


crates/pop-cli/src/commands/fork.rs
```diff
@@ -76,10 +83,19 @@ pub(crate) struct ForkArgs {
...
+	#[arg(long)]
+	pub at: Option<u32>,
+
 	/// Internal flag: run as background server (used by detach mode).
 	#[arg(long, hide = true)]
 	#[serde(skip)]
 	pub serve: bool,
+
+	/// Internal flag: path to write readiness info to (used by detach mode).
+	#[arg(long, hide = true)]
```
tsenovilla: Maybe add a `requires` to ensure it's only used with detach?

moliholy:
@al3mart `pop up` has many functionalities already:

- Deploy contracts.
- Deploy chains with PDP.
- Deploy a zombienet-based network.

If now add yet another variant it's definitely becoming cumbersome and hard to use. Not saying it's not doable, but at this stage I think we should think carefully the UX to make it happen so that users don't get lost. I personally think it's perfectly to have a `fork` command completely separated from `up`. Furthermore: I'd remove all PDP-related code into a separate command (deploy?) to simplify it.

tsenovilla:
I agree, the PR purpose in general feels weird. What I wouldn't do is to merge both commands, they serve to different purposes and work correctly on their own. I approved this because the proposal included it, so theoretically we should include it. But I wouldn't be against not merging it and justifying it using @moliholy's comment: https://github.com/r0gue-io/pop-cli/pull/929#issuecomment-3872329180

PR #948, opened 2026-02-13
fix: return non-zero exit codes for call command failures

### LINKED ISSUE #943: call contract/chain: hard failure output can still return exit code 0
## Summary
`pop call contract` can return exit code `0` while printing clear runtime failures.

## Environment
- pop-cli: `0.12.1`
- OS: macOS (Apple Silicon)

## Reproduction
1. Run this call against an unavailable endpoint:
```bash
pop call contract --path ./flipper --contract 0x0000000000000000000000000000000000000000 --message get --url ws://127.0.0.1:9944
```
2. Also reproducible with a chain-call encoding failure:
```bash
pop call chain --url ws://localhost:19944 --pallet Balances --function transfer_keep_alive --args 5FHneW46xGXgs5mUiveU4sbTyGBzmst2mK8z6kT5HTf3V1fJ 123456789 --suri //Alice --skip-confirm
```

## Actual
- Output shows explicit failures such as:
  - `Anyhow error: ... Connection refused`
  - `Failed to encode call data ...`
- Command can still exit with status `0`.

## Expected
- Commands that print hard execution failures should return non-zero status consistently.

moliholy:
## Summary

The CLI now returns proper exit codes when chain or contract calls fail. Previously, failures (RPC errors, transaction rejections, user cancellations) would exit with code 0, making it impossible for scripts and CI pipelines to detect failures. Now, only successful operations return 0, while errors return 1.

Additionally, the integration test for parachain lifecycle was optimized by moving an unnecessary 12-second wait from before read queries (which don't require waiting) to before the second signed extrinsic (where it's needed to avoid nonce conflicts).

Closes #943.

### DIFF
--- a/crates/pop-cli/tests/chain.rs
+++ b/crates/pop-cli/tests/chain.rs
@@ -24,7 +24,7 @@ use std::{
 };
 use strum::VariantArray;
 use tempfile::tempdir;
-use tokio::{process::Child, time::sleep};
+use tokio::process::Child;
 
 /// Utility child process wrapper to kill the child process on drop.
 ///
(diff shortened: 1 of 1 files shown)


Daanvdplas:
I am going to push a PR to this branch to avoid CI slowing us down.

PR #955, opened 2026-02-16
fix(call): prevent "not connected" error after successful queries

### LINKED ISSUE #942: call chain: successful decoded output can end with 'Error: not connected' and inconsistent command status
## Summary
`pop call chain` can print a successful decoded result and still end with `Error: not connected`, and command status behavior is inconsistent across runs.

## Environment
- pop-cli: `0.12.1`
- OS: macOS (Apple Silicon)

## Reproduction
1. Start a local ink node:
```bash
pop up ink-node -d -y -i 19944 -e 18545
```
2. Query a constant:
```bash
pop call chain --url ws://localhost:19944 --pallet Balances --function ExistentialDeposit
```
3. Query storage:
```bash
pop call chain --url ws://localhost:19944 --pallet System --function Account --args 0xd43593c715fdd31c61141abd04a99fd6822c8558854ccde39a5684e7a56da27d
```
4. Query events:
```bash
pop call chain --url ws://localhost:19944 --pallet System --function Events
```

## Actual
- Command output can include valid decoded values (`◆ ...`) and then append:
  - `Error: not connected`
- Status code behavior is inconsistent (observed both non-zero and zero in different runs while showing the same trailing error text).

## Expected
- Successful read/query calls should complete without contradictory trailing connection errors.
- Exit status should be deterministic and match the true result.

moliholy:
## Summary

Fixes spurious "Error: not connected" messages appearing after successful `pop call chain` operations for storage, constant, and event queries.

The issue was a race condition where subxt's background connection handler would attempt operations during cleanup, sometimes logging errors even though the query succeeded. This also caused inconsistent exit codes.

The fix ensures background tasks complete cleanly before the function returns.

Resolves #942

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9367,6 +9367,7 @@ dependencies = [
  "dirs",
  "duct 0.13.7",
  "env_logger",
+ "futures",
  "git2",
  "hex",
  "jsonrpsee",
--- a/crates/pop-cli/Cargo.toml
+++ b/crates/pop-cli/Cargo.toml
@@ -56,6 +56,8 @@ regex.workspace = true
 tracing-subscriber = { workspace = true, optional = true }
 scale-value = { workspace = true, optional = true }
 scale-info = { workspace = true }
+jsonrpsee = { workspace = true, features = ["ws-client"] }
+futures.workspace = true
 
 # telemetry
 pop-telemetry = { path = "../pop-telemetry", version = "0.13.0", optional = true }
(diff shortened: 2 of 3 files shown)


Daanvdplas (commented):
Nice catch on the race. One concern: `drop(chain)` + `tokio::task::yield_now().await` is still heuristic. A single scheduler yield does not guarantee subxt background tasks have completed, so this can still be flaky under load/scheduling variance.

Can we make this deterministic (or at least prove it) with a regression test that runs the constant/storage/events paths repeatedly and asserts both output and exit status never show `not connected`? If subxt exposes a shutdown/join path we should prefer that over `yield_now()`.

PR #959, opened 2026-02-16
fix(ci): add binary caching and HTTP retry with backoff

### LINKED ISSUE #846: Define strategy for tests including third party resources
Motivation: https://github.com/r0gue-io/pop-cli/pull/812#discussion_r2620178054

Right now we have several tests in the codebase that need to fetch third party resources (ink-node, omni-node, or other binaries). I've found tests failing because the server was down in different branches during the last days randomly, I think it's related to what's discussed here: https://github.com/orgs/community/discussions/181699.

Talking with @AlexD10S, we moved away from using live endpoints because sometimes tests failed because they were down, and started using GH tokens to fetch resources instead as this is more stable. But what last weeks prove is that this may be not enough, given that GH may respond with a 400 or a 504 from time to time.

We should define a clear strategy of what we should do in those cases. There's a few (draft) approaches coming to my mind, but none of them looks like an optimal solution.

1. Don't do anything at all, accept it's what it's and re-run tests later -> Lazy but maybe acceptable, at the end of the day, in real live resources may be down, so also for tests.
2. Accept "false negatives" in those tests, this is, if the resource cannot be fetched, we ignore the tests as the failure isn't related to something in our codebase. The pro of this approach is that we get a fully deterministic test env, the con is that we might think something is just a false negative when it's not.

Anyway, we should agree on what's the best approach here (for sure none of the above). Additionally I'd like to take advantage of this issue to suggest using live endpoints again once we decide what to do. Why? because querying them is probably faster than spawning a TestNode each time we need to test something against a running node, and if we have a strategy on what to do if a third party resource isn't available, it applies to any third party resource, both fetching something from GH or querying a live node

moliholy:
## Summary

- Cache downloaded binaries (polkadot, ink-node, etc.) between CI runs using `actions/cache@v4`, keyed by source files that declare binary versions for automatic invalidation on upgrades
- Add `get_with_retry()` helper with exponential backoff (2s, 4s, 8s) for HTTP downloads, retrying on 5xx, 408, 429, and connection errors while failing immediately on 404
- Update ink-node download in `pop-fork-integration-tests` to use the shared pop cache directory so the cache step is effective

Closes #846

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -109,7 +109,7 @@ jobs:
         with:
           path: |
             ${{ matrix.os == 'ubuntu-latest' && '~/.cache/pop' || '~/Library/Caches/pop' }}
-          key: pop-binaries-${{ matrix.os }}-${{ runner.arch }}-${{ hashFiles('crates/pop-chains/src/up/mod.rs', 'crates/pop-common/src/test_env.rs') }}
+          key: pop-binaries-${{ matrix.os }}-${{ runner.arch }}-${{ hashFiles('.github/workflows/integration-tests.yml', 'crates/pop-chains/src/up/mod.rs', 'crates/pop-common/src/test_env.rs') }}
           restore-keys: |
             pop-binaries-${{ matrix.os }}-${{ runner.arch }}-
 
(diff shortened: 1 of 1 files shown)


crates/pop-common/src/sourcing/mod.rs
```diff
@@ -503,6 +503,76 @@ impl From<&str> for TagPattern {
...
+/// Performs an HTTP GET request with exponential backoff retry logic.
+///
+/// Retries on transient errors: 5xx server errors, 408 (Request Timeout), 429 (Too Many
+/// Requests), and connection-level failures. Does not retry on other 4xx client errors (e.g. 404
+/// means the resource genuinely does not exist).
+///
+/// # Arguments
+/// * `url` - The URL to fetch.
+/// * `status` - Used to log retry attempts.
+async fn get_with_retry(url: &str, status: &impl Status) -> Result<reqwest::Response, Error> {
```
cmichi: There is https://crates.io/crates/reqwest-middleware, which has an exponential back-off mechanism. You'd pull in another dep though and I'm not sure if it has the same defintion of "transient errors".

moliholy: Created an issue as an improvement. Thanks!

PR #960, opened 2026-02-16
refactor(pop-cli): remove --provider from pop new chain

### LINKED ISSUE #905: refactor: pop new chain
Do we really need to provide `--provider` and `--template`, in stead of just `--template` (same what we introduced here for contracts (#700)?

moliholy:
## Summary

Removes the redundant `--provider` positional argument from `pop new chain`. Each `ChainTemplate` variant already encodes its provider via the strum `Provider` property, so the separate arg was unnecessary and created a validation step (`is_template_supported`) that only existed because the two args could conflict.

This aligns the chain scaffolding flow with the contract flow, which uses `--template` alone.

Closes #905

### DIFF
--- a/crates/pop-cli/tests/chain.rs
+++ b/crates/pop-cli/tests/chain.rs
@@ -14,7 +14,6 @@ use pop_common::{
 	polkadot_sdk::sort_by_latest_semantic_version,
 	pop, resolve_port,
 	sourcing::{ArchiveFileSpec, GitHub::ReleaseArchive},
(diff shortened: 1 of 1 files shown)


cmichi (approved):
I'm not sure if the changelog in this repo is to be updated with each PR, but as you remove a user-facing cli flag no harm in updating it with your PR: https://github.com/r0gue-io/pop-cli/blob/main/CHANGELOG.md.

PR #966, opened 2026-02-16
refactor(pop-common): replace hand-rolled retry with reqwest-middleware

### LINKED ISSUE #965: manage retries with reqwest
There is https://crates.io/crates/reqwest-middleware, which has an exponential back-off mechanism. You'd pull in another dep though and I'm not sure if it has the same defintion of "transient errors".

_Originally posted by @cmichi in https://github.com/r0gue-io/pop-cli/pull/959#discussion_r2813829115_

moliholy:
Closes #965

This PR replaces the custom `get_with_retry` exponential backoff loop with `reqwest-middleware` + `reqwest-retry` for a battle-tested, maintainable retry solution. Also upgrades `reqwest` from 0.12 to 0.13 for compatibility with the latest middleware crates.


cmichi (approved):
Nice!

PR #982, opened 2026-02-17
fix(fork): include detached output links for polkadot.js and PAPI

Daanvdplas:
## Summary
- split fork detached-output fix out of #981 into its own PR
- include polkadot.js and PAPI links in fork detached output
- include rustfmt/lint follow-up for the same file

## Validation
- cargo +nightly fmt --all -- --check


moliholy (approved):
Clean extraction — just one nit below.

al3mart (approved):

PR #992, opened 2026-02-19
feat(pop-cli): --json support for test command

### LINKED ISSUE #974: feat(pop-cli): --json support for test command (including runtime test subcommands)
Part of #850

## Context

The `test` command runs cargo tests. Its subcommands (`on-runtime-upgrade`, `execute-block`, `create-snapshot`, `fast-forward`) use `try-runtime-cli` for runtime operations. All spawn subprocesses.

### `test` (main)

```rust
struct TestOutput {
    command: String,        // e.g., "cargo test" or "cargo nextest run"
    success: bool,
}
```

### Runtime test subcommands

```rust
struct RuntimeTestOutput {
    subcommand: String,     // "on-runtime-upgrade" | "execute-block" | etc.
    success: bool,
    output_path: Option<String>, // snapshot path for create-snapshot
}
```

## Challenges

- **Subprocesses**: All test execution happens via child processes (`cargo test`/`cargo nextest`, `try-runtime-cli`). Their output must not pollute stdout in `--json` mode. Depends on #970.
- **Interactive prompts** (subcommands): State source selection, profile selection, block URI input. Must be flags in `--json` mode.
- **Exit codes**: Test failures should produce a structured error envelope with the test output in `details`, not just a non-zero exit.

## Approach

1. Thread `OutputMode` into `test::Command::execute()` and all subcommand execute functions.
2. Capture subprocess output; on failure, include relevant output in `CliError.details`.
3. Use `JsonCli` for subcommands that take a `cli` parameter.
4. Return structured result with success/failure and any relevant paths.

## Acceptance criteria

- `pop --json test` returns JSON envelope for test results.
- `pop --json test on-runtime-upgrade --runtime ...` returns JSON envelope.
- Test failures produce `BUILD_ERROR` envelope with test output in `details`.
- Human mode unchanged.

Daanvdplas:
Closes #974

Adds JSON envelope support for `pop test` and runtime test subcommands.

### DIFF
--- a/crates/pop-cli/src/output.rs
+++ b/crates/pop-cli/src/output.rs
@@ -138,22 +138,10 @@ pub(crate) fn invalid_input_error(message: impl Into<String>) -> anyhow::Error {
 /// Message used when `--json` mode cannot satisfy an interactive prompt.
 pub(crate) const JSON_PROMPT_ERR: &str = "interactive prompt required but --json mode is active";
 
-/// Error returned when a command in `--json` mode requires an interactive prompt.
-#[derive(Debug)]
-pub(crate) struct PromptRequiredError;
-
-impl std::fmt::Display for PromptRequiredError {
-	fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-		write!(f, "{}", JSON_PROMPT_ERR)
-	}
-}
-
-impl std::error::Error for PromptRequiredError {}
-
 /// Builds an I/O error carrying a typed prompt-required cause.
 #[allow(dead_code)]
 pub(crate) fn prompt_required_io_error() -> std::io::Error {
-	std::io::Error::other(PromptRequiredError)
+	std::io::Error::other(PromptRequiredError(JSON_PROMPT_ERR.to_string()))
 }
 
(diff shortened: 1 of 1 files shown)


crates/pop-cli/src/main.rs
```diff
@@ -23,6 +26,7 @@ pub(crate) mod output;
 mod style;
 #[cfg(feature = "wallet-integration")]
 mod wallet_integration;
+const JSON_PROMPT_ERR: &str = "interactive prompt required but --json mode is active";
```
moliholy: `JSON_PROMPT_ERR` is defined both here and in `cli.rs:469`. `json_error_response` relies on string-matching the error message to classify it as `INVALID_INPUT`. If someone changes the wording in `cli.rs` without updating this constant, the match silently fails and falls through to `INTERNAL`, wrong code, no compiler warning.

Two options:
- Export one constant from a shared location and import in both files
- Have `JsonCli` return `InvalidInputError` directly instead of `io::Error::other(...)`, which eliminates string matching entirely

moliholy (approved):
Clean implementation. One P1 note below about the error-classification approach.

PR #1005, opened 2026-02-23
chore: remove passethub

al3mart:
## Summary
- Removes PassetHub support: struct, registry entry, chain spec artifact, and temporary workaround in network spawning
- Replaces `passet-hub-spec.json` test fixture with `coretime-paseo-local.json` from [paseo-network/runtimes v2.0.6](https://github.com/paseo-network/runtimes/releases/tag/v2.0.6) The smallest fixture from the Paseo chains.
- Restores a previously commented-out assertion in `up/mod.rs` tests

PassetHub was a temporary chain added for a hackathon. It is no longer needed.

## Test plan
- [x] `cargo check --locked` — compiles
- [x] `cargo nextest run --lib --bins -p pop-chains` — 222 tests pass


moliholy (commented):
Mostly clean, just one thing: there's a leftover `"Passet Hub"` reference in `crates/pop-cli/src/commands/new/frontend.rs:144`. Since PassetHub is being removed, this should probably be updated or removed too, otherwise the Typink template generation would reference a network that no longer exists.

